PR #1, opened 2019-03-20
added cmd flags for http headers

davidcv5:
@hbagdi thanks a lot for building deck! The logic for http headers was already implemented by you, though the cmd wasn't setup to accept them. Please review. Thank you!

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -65,6 +65,13 @@ func init() {
 	viper.BindPFlag("kong-addr",
 		rootCmd.PersistentFlags().Lookup("kong-addr"))
 
+	rootCmd.PersistentFlags().StringSlice("headers", []string{},
+		"HTTP Headers to call Kong's Admin API.\n"+
+			"This value can also be set using DECK_HEADERS"+
+			" environment variable.")
+	viper.BindPFlag("headers",
+		rootCmd.PersistentFlags().Lookup("headers"))
+
 	rootCmd.PersistentFlags().Bool("tls-skip-verify", false,
 		"Disable verification of Kong's Admin TLS certificate.\n"+
 			"This value can also be set using DECK_TLS_SKIP_VERIFY "+
@@ -125,6 +132,7 @@ func initConfig() {
 	config.TLSServerName = viper.GetString("tls-server-name")
 	config.TLSSkipVerify = viper.GetBool("tls-skip-verify")
 	config.TLSCACert = viper.GetString("ca-cert")
+	config.Headers = viper.GetStringSlice("headers")
 	verbose = viper.GetInt("verbose")
 	noColor = viper.GetBool("no-color")
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -66,9 +66,8 @@ func init() {
 		rootCmd.PersistentFlags().Lookup("kong-addr"))
 
 	rootCmd.PersistentFlags().StringSlice("headers", []string{},
-		"HTTP Headers to call Kong's Admin API.\n"+
-			"This value can also be set using DECK_HEADERS"+
-			" environment variable.")
+		"HTTP Headers(key:value) to inject in all requests to Kong's Admin API.\n"+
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -65,6 +65,13 @@ func init() {
 	viper.BindPFlag("kong-addr",
 		rootCmd.PersistentFlags().Lookup("kong-addr"))
 
+	rootCmd.PersistentFlags().StringSlice("headers", []string{},
+		"HTTP Headers to call Kong's Admin API.\n"+
+			"This value can also be set using DECK_HEADERS"+
```
hbagdi: How will parsing multiple headers work in case of multiple headers?
Will it split based on a comma?

hbagdi: Viper splits this based on whitespace.

cmd/root.go
```diff
@@ -65,6 +65,13 @@ func init() {
 	viper.BindPFlag("kong-addr",
 		rootCmd.PersistentFlags().Lookup("kong-addr"))
 
+	rootCmd.PersistentFlags().StringSlice("headers", []string{},
+		"HTTP Headers to call Kong's Admin API.\n"+
```
hbagdi: Let's change this to `HTTP Headers(key:value) to inject in all requests to Kong's Admin API.\n. This flag can be specified multiple times to inject multiple headers.`

hbagdi:
@davidcv5 Thank you!

cmd/root.go
davidcv5: Definitely a better message 👍

cmd/root.go
davidcv5: @hbagdi I've been trying to find a way to use `Authorization` header as an env var, but because `viper` splits based on whitespace, the following won't work:
- `Authorization: Basic xyztoken` splits into `Authorization:`, `Basic`, `xyztoken`
- `Authorization:Basic xyz token` splits into `Authorization:Basic`, `xyztoken`
So your `RoundTrip` will get the wrong values.
For now I removed it from the argument description and changed the wording to what you suggested.
Please let me know what you think.. thank you!

davidcv5: I have successfully tested it as a cli argument and given those are passed separately by repeating `--headers`. It works perfect. Will be adding `deck` to our CI pipelines, thanks a lot for building it.. will send any feedback, recommendations, if I find any, once we have it fully integrated.

hbagdi: @davidcv5 Glad that it's working well for you. Please feel free to open up issues for feature requests or bugs.

Regarding the problem of `viper` splitting based on the whitespace, we can change to split it based on comma (PR welcome), but then a header value cannot have a comma (which is better than the current space restriction).

davidcv5: @hbagdi I agree.. will look into it. Also, might add some makefile or whatever you prefer to dockerize it and maybe create a package of some kind, brew, npm, etc.. I'm creating a docker image myself right now, but others might find it helpful, since not everyone is familiar with `go`.
I pushed an alpine based image here https://hub.docker.com/r/davidcv5/deck.. but it would make more sense to be under your name, since it's yours :)

hbagdi: > Also, might add some makefile or whatever you prefer to dockerize it and maybe create a package of some kind, brew, npm, etc.. I'm creating a docker image myself right now, but others might find it helpful, since not everyone is familiar with go.

We have a debian and RPM package that we publish with the [release](https://github.com/hbagdi/deck/releases). Happy to host a docker image as well if that works well for you and other users. 

Do you think you could PR the Dockerfile you've used?
Feel free to create issues to track these as well (so that they don't get lost with time), thanks!


With Go's single binary builds, usually a single binary download is super convenient but I totally get that something like a `brew install ...` is helpful to get started. Again, welcome these changes!

Charnnarong:
Error: parsing headers: splitting header key-value '123'
when --headers "hkey:hvalue,123". In this case what if value contains comma like `hvalue,123` ? How can I use comma in value?

PR #5, opened 2019-04-03
Dockerfile and build.sh script

zoobab:
Dockerfile and build.sh script.


hbagdi:
@zoobab, Thank you for the PR!
I had a local Dockerfile that I pushed up which uses multi-stage docker builds.
The v0.3.0 version will be released on Docker hub.

PR #6, opened 2019-04-04
using only service.id when updating a plugin

davidcv5:
@hbagdi Updating a plugin works when a plugin is set for a `route`, but when set for a `service` it fails, basically because of the following:

**Bad Request**: `PUT /plugins/<id> {..., service: {id: <service-id>, <everything else from service>}} // full service struct included`

```json
{
  "message": "schema violation (protocols: must match the protocols of at least one route pointing to this Plugin's service)",
  "name": "schema violation",
  "fields": {
    "protocols": "must match the protocols of at least one route pointing to this Plugin's service"
  },
  "code": 2
}
```

**This works**: `PUT /plugins/<id> {..., service: {id: <service-id>}} // only service.id included`

Creating a new `service` with only the ID and assigning it to the `plugin` before calling `s.client.Plugins.Create(nil, &plugin.Plugin)` solves the issue.

Again, thank you!!

### DIFF
--- a/solver/kong/plugin.go
+++ b/solver/kong/plugin.go
@@ -69,6 +69,9 @@ func (s *PluginCRUD) Update(arg ...crud.Arg) (crud.Arg, error) {
 	event := eventFromArg(arg[0])
 	plugin := pluginFromStuct(event)
 
+	if plugin.Service != nil {
+		plugin.Service = &kong.Service{ID: plugin.Service.ID}
+	}
 	updatedPlugin, err := s.client.Plugins.Create(nil, &plugin.Plugin)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


solver/kong/plugin.go
```diff
@@ -69,6 +69,9 @@ func (s *PluginCRUD) Update(arg ...crud.Arg) (crud.Arg, error) {
 	event := eventFromArg(arg[0])
 	plugin := pluginFromStuct(event)
 
+	if plugin.Service != nil {
+		plugin.Service = &kong.Service{ID: plugin.Service.ID}
+	}
```
hbagdi: Could you please add this for the Route and Consumer case? Thanks!

davidcv5: @hbagdi I just added it for `Route`, but `Consumer` doesn't seem to have a `Service`.

hbagdi: Oh, I'm really sorry, I meant, we should add this for `plugin.Route` and `plugin.Consumer` as well.

davidcv5: @hbagdi please review when you get a chance.. thanks!

hbagdi:
@davidcv5 Thanks for the report and fix!

PR #10, opened 2019-04-09
Feature: write to stdout by indicating "-" as the output file name - resolves #7

matthewbednarski:
Feature: write to stdout by indicating "-" as the output file name

Resolves issue #7

PR #11, opened 2019-04-09
Feature: diff/sync commands - read from stdin - resolves #8

### LINKED ISSUE #8: Feature Request: sync/diff command - use of '-' indicates to read from stdin
Hi,

It would be nice if the `sync` and `diff` commands would read try to read from `stdin` if a '`-`' (_dash_) is indicated on the command line.

matthewbednarski:
Feature: diff/sync commands - read from stdin - resolves #8


cmd/diff.go
```diff
@@ -56,8 +71,11 @@ that will be created or updated or deleted.
...
 	rootCmd.AddCommand(diffCmd)
 	diffCmd.Flags().StringVarP(&diffCmdKongStateFile,
-		"state", "s", "kong.yaml", "file containing Kong's configuration.")
+		"state", "s", "kong.yaml", "file containing Kong's configuration")
 	diffCmd.Flags().BoolVar(&dumpConfig.SkipConsumers, "skip-consumers",
 		false, "do not diff consumers or "+
 			"any plugins associated with consumers")
+	diffCmd.Flags().BoolVarP(&diffCmdKongStateFileFromStdin, "read-state-from-stdin",
+		"-", false, "read Kong's configuration state from stdin (overrides "+
+			"--state flag)")
```
hbagdi: Question: have you considered using `-` as an input to the -s/--state flag to read input from os.StdIn?
It would be nice if we can avoid this flag.

matthewbednarski: The way it is now, if you pass `deck diff -s -`  it will read from `stdin` like in issue #10 . 

I added the other `BoolVal` flag just so you could call it with `deck diff --` or `deck diff --read-state-from-stdin` and still explicitly read from stdin.

hbagdi: I feel that will confuse users and is unnecessary.
Could you please update the PR to remove this flag?
Also, please don't remove the `kong.yaml` file present in the repository.

file/reader.go
```diff
@@ -204,6 +207,9 @@ func GetStateFromFile(filename string) (*state.KongState, error) {
 
 func readFile(kongStateFile string) (*fileStructure, error) {
 
+	if kongStateFile == "-" {
+		return readStdin()
```
hbagdi: This elaborate method call can be simply replaced by:
`data, err = ioutil.ReadAll(os.Stdin)` and then the logic below will take care of unmarshaling it into a YAML.

Let's not duplicate the YAML Unmarshaling logic as well.

cmd/sync.go
```diff
@@ -57,4 +72,7 @@ func init() {
 	syncCmd.Flags().BoolVar(&dumpConfig.SkipConsumers, "skip-consumers",
 		false, "do not diff consumers or "+
 			"any plugins associated with consumers")
+	syncCmd.Flags().BoolVarP(&syncCmdKongStateFileFromStdin, "read-state-from-stdin",
+		"-", false, "read Kong's configuration state from stdin (overrides "+
+			"--state flag)")
```
hbagdi: Please remove this flag. `deck sync -s -` should be good end user experience for reading the content from stdin or writing it to stdout.

kong.yaml
```diff
@@ -1,229 +0,0 @@
-services:
```
hbagdi: Please do not delete this file.

cmd/sync.go
```diff
@@ -42,6 +45,18 @@ to get Kong's state in sync with the input state.`,
...
+		} else {
+			// cobra doesn't seem to allow managing the "--" or "-"
+			//   flag directly so it has to be parsed out of os.Args
+			for i, arg := range os.Args {
+				if i > 1 && arg == "--" {
+					syncCmdKongStateFile = "-"
+					break
+				}
+			}
+		}
```
hbagdi: This is also unnecessary and merely treating `-` as a special value to the `--state` arg should be for this feature.

kong.yaml
matthewbednarski: That was an accident.

hbagdi:
@matthewbednarski Thank you for the contribution!
I've manually merged this in c08a684.

PR #33, opened 2019-07-15
Add kong db-less compatibility to README.md

davidcv5:
@hbagdi here is a short PR, as suggested here https://discuss.konghq.com/t/migrate-from-db-to-db-less-declarative-config/3723.
As always, thanks for the great work!


davidcv5:
That sounds great @hbagdi, is simple enough for others to add it manually for now, I had no idea decK's output matched the db-less config until I found your post, you rock man.. thanks for all the great work!

PR #34, opened 2019-07-18
Dockerfile - add decK to path

davidcv5:
Current Dockerfile builds under `/deck` directory the `deck` binary, then adds the full directory with all source code to the image, then calls deck inside that directory.

The proposed change does the following:
- upgrade to latest `alpine` image, currently 3.10
- copies only the `deck` binary into the final image at `/usr/local/bin` so it can be in the `PATH`
- uses just `deck` as the entrypoint since it's in the `PATH`


davidcv5:
Awesome.. thank you!

PR #58, opened 2019-08-19
fix(state): use upstreamIDTarget index instead of target

zeeshen:
Related issue https://github.com/hbagdi/deck/issues/57

Remove the unique `target` index in `TargetsCollection`, replace it with a union index of (`upstreamID`, `target`). Allow different upstreams to share same targets.

I know this PR may break calls of `TargetsCollection.Get` outside this repo. But I checked the `kubernetes-ingress-controller` repo and there's no direct usage of `TargetsCollection`. So I think it might ok.

As a fan of Kong, we're considering using the office kubernetes-ingress-controller to replace our version. So truly appreciated if you can take a look at this PR. Thanks.

### DIFF
--- a/diff/diffTarget.go
+++ b/diff/diffTarget.go
@@ -36,7 +36,7 @@ func (sc *Syncer) deleteTarget(target *state.Target) (*Event, error) {
 			target)
 	}
 	// lookup by Name
-	_, err := sc.targetState.Targets.Get(*target.Target.Target)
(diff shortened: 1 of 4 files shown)


zeeshen:
Sorry, found that the `diffTarget` will delete newly added targets as the `upstreamID` in `targetState.Targets` is just a placeholder.

Issue #57, opened 2019-08-19
bug(feature?): should allow same `target.Target` for different upstreams


zeeshen:
Currently, [GetStateFromContent](https://github.com/hbagdi/deck/blob/master/file/reader.go#L45) returns a `duplicate target definitions found for: ...` error when different upstreams have same target. 

It's not rare that different services sharing the same group of pods in Kubernetes. Such restrictions causes the kubernetes-ingress-controller stop working. Maybe we can use `${upstream.id}/${target.Target}` here as the `id` in `TargetsCollection`?

hbagdi:
Fixed in https://github.com/hbagdi/deck/commit/b6265c3a6207ead319d94cfdceba9760ffe4adaf

PR #60, opened 2019-08-20
fix(state): use upstreamNameTarget index instead of target

zeeshen:
Related issue #57

Remove the unique target index in TargetsCollection, replace it with a union index of (upstreamName, target). Allow different upstreams to share same targets.

I know this PR may break calls of TargetsCollection outside this repo, and requiring upstream name in target is kind of tricky. But I checked the kubernetes-ingress-controller repo and there's no direct usage of TargetsCollection. So I think it might be ok.

As a fan of Kong, we're considering using the office kubernetes-ingress-controller to replace our version. So truly appreciated if you can take a look at this PR. Thanks.

### DIFF
--- a/diff/postProcess.go
+++ b/diff/postProcess.go
@@ -175,11 +175,18 @@ func (crud *targetPostAction) Create(arg ...crud.Arg) (crud.Arg, error) {
 	if !ok {
 		panic("whoops")
 	}
-	svc, ok := arg[1].(*state.Target)
+	target, ok := arg[1].(*state.Target)
 	if !ok {
 		panic("whoops")
 	}
-	s.Targets.Add(*svc)
+
+	u, err := s.Upstreams.Get(*target.Upstream.ID)
+	if err != nil {
+		return nil, err
+	}
+	target.Upstream = u.DeepCopy()
+
+	s.Targets.Add(*target)
 	return nil, nil
 }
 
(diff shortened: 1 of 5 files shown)


diff/postProcess.go
```diff
@@ -175,11 +175,18 @@ func (crud *targetPostAction) Create(arg ...crud.Arg) (crud.Arg, error) {
...
 	}
-	s.Targets.Add(*svc)
+
+	u, err := s.Upstreams.Get(*target.Upstream.ID)
+	if err != nil {
+		return nil, err
+	}
+	target.Upstream = u.DeepCopy()
+
+	s.Targets.Add(*target)
```
hbagdi: Could you explain the reason for doing these changes?

Other entities which have a foreign relation (like routes and services), don't actually search for a service and associate it with the route before adding/updating.

These shouldn't be required from what I can see. I might be missing something here.

hbagdi:
We will include this in Kong Ingress Controller 0.6.0.

hbagdi:
@zeeshen Thank you for the fix.
I've incorporated the above comments and manually merged in b6265c3a6207ead319d94cfdceba9760ffe4adaf.

diff/postProcess.go
zeeshen: If we didn't set the upstream field here, the `Upstream` field of `target` here will have only a `ID` field. As we require `target.Upstream.Name` in `TargetsCollection `, this `s.Targets.Add(*target)` will return an error here.
And as the error is ignored here, this may cause problem for later usage of `s.Targets`.

PR #80, opened 2019-09-26
feat(docker) optionally base of Kong image

Tieske:
When building the container, the argument `BASE_IMAGE` can be set
to an image name of a Kong image (must be Alpine). In that case
when executing, it will start Kong before executing decK, and stop
it afterwards.

This allows the container to be started with the same environment
variables as any other Kong container, and then decK can access
Kong locally (within the container).

example usecase is a Kong cluster in data-plane mode without an
admin api exposed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,5 @@
+ARG IMAGE_BASE=alpine:3.10
+
 FROM golang:1.13.0 AS build
 WORKDIR /deck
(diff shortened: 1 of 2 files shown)


Tieske:
closing this, as discussed. Probably better to use the orchestration tools to spin up a separate Kong and do the proper checks and balances, instead of adding all in one image.

Issue #82, opened 2019-10-06
dockerfile: use a non-root user to run decK


hbagdi:

digikin:
let me try this

rampreeth:
@digikin , are you working on this?

digikin:
Attempting to

PR #83, opened 2019-10-06
update Dockerfile to run as non-root

guusvw:
solves #82 
Signed-off-by: Guus van Weelden <guus.vanweelden@moia.io>

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,3 @@
 .git
+.travis.yml
 vendor
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,2 @@
 .git
-.travis.yml
 vendor
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.13 AS build
+FROM golang:1.13.0 AS build
 WORKDIR /deck
 COPY . . 
 RUN go mod download
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,10 +11,10 @@ RUN apk update \
     && apk upgrade \
     && apk --no-cache add ca-certificates
 
-ENV USER=deck
-ENV GROUP=deck
-ENV UID=9999
-ENV GID=9999
+ARG USER=deck
+ARG GROUP=deck
+ARG UID=9999
+ARG GID=9999
 
 RUN addgroup --gid "$GID" "$GROUP" \
     && adduser \
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@ RUN apk update \
     && apk upgrade \
     && apk --no-cache add ca-certificates
 
-ENV USER=appuser
+ENV USER=deck
 ENV GROUP=deck
 ENV UID=9999
 ENV GID=9999
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,4 +30,3 @@ COPY --from=build /deck/deck /usr/local/bin
 USER $USER
 
 ENTRYPOINT ["deck"]
-CMD ["help"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,31 @@
-FROM golang:1.13.0 AS build
+FROM golang:1.13 AS build
 WORKDIR /deck
-COPY go.mod ./
-COPY go.sum ./
+COPY . . 
 RUN go mod download
-ADD . .
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o deck
 
 FROM alpine:3.10
-RUN apk --no-cache add ca-certificates
-WORKDIR /root/
+RUN apk update \
+    && apk upgrade \
+    && apk --no-cache add ca-certificates
+
+ENV USER=appuser
+ENV GROUP=appusers
+ENV UID=9999
+ENV GID=9999
+
+RUN addgroup --gid "$GID" "$GROUP" \
+    && adduser \
+    --disabled-password \
+    --gecos "" \
+    --ingroup "$GROUP" \
+    --no-create-home \
+    --uid "$UID" \
+    "$USER"
+
 COPY --from=build /deck/deck /usr/local/bin
+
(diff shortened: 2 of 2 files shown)


.dockerignore
```diff
@@ -1,2 +1,3 @@
 .git
+.travis.yml
```
hutchic: I think we still want this

guusvw:
```suggestion
```

guusvw: removed it. it was part of the `.dockerignore`, but fine for me ;)

Dockerfile
```diff
@@ -1,13 +1,31 @@
-FROM golang:1.13.0 AS build
+FROM golang:1.13 AS build
```
hbagdi: We would like to have control over the patch version of Go as well.
Not pinning the patch version can result in using an older version of Go due to Docker's cache.

Please revert this change.

Dockerfile
```diff
@@ -1,13 +1,31 @@
...
 WORKDIR /deck
-COPY go.mod ./
-COPY go.sum ./
+COPY . .
```
hbagdi: Please revert this change.
It is intentional that we only first copy over `go.mod` and `go.sum`. 

This gives Docker image caching, meaning that every time we execute a `docker build`, it won't result in a new hash and re-download all the modules in the next step.

Dockerfile
```diff
@@ -1,13 +1,31 @@
...
 
 FROM alpine:3.10
-RUN apk --no-cache add ca-certificates
-WORKDIR /root/
+RUN apk update \
+    && apk upgrade \
+    && apk --no-cache add ca-certificates
+
+ENV USER=appuser
+ENV GROUP=appusers
```
hbagdi: Let's name the user and group as `deck`.

Dockerfile
```diff
@@ -1,13 +1,31 @@
...
+    --no-create-home \
+    --uid "$UID" \
+    "$USER"
+
 COPY --from=build /deck/deck /usr/local/bin
+
+USER $USER
+
 ENTRYPOINT ["deck"]
+CMD ["help"]
```
hbagdi: Why is this changed?

Dockerfile
guusvw:
```suggestion
FROM golang:1.13.0 AS build
```

Dockerfile
```diff
@@ -1,13 +1,31 @@
 FROM golang:1.13.0 AS build
 WORKDIR /deck
-COPY go.mod ./
-COPY go.sum ./
+COPY . .
```
guusvw:
```suggestion
COPY go.mod ./
COPY go.sum ./
```

Dockerfile
```diff
@@ -3,11 +3,30 @@ WORKDIR /deck
 COPY go.mod ./
 COPY go.sum ./
 RUN go mod download
```
guusvw:
```suggestion
RUN go mod download
ADD . .
```

Dockerfile
```diff
@@ -7,7 +7,27 @@ ADD . .
...
 
 FROM alpine:3.10
-RUN apk --no-cache add ca-certificates
-WORKDIR /root/
+RUN apk update \
+    && apk upgrade \
+    && apk --no-cache add ca-certificates
+
+ENV USER=appuser
+ENV GROUP=appusers
```
guusvw:
```suggestion
ENV GROUP=deck
```

Dockerfile
```diff
@@ -7,7 +7,27 @@ ADD . .
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o deck
 
 FROM alpine:3.10
-RUN apk --no-cache add ca-certificates
-WORKDIR /root/
+RUN apk update \
+    && apk upgrade \
+    && apk --no-cache add ca-certificates
+
+ENV USER=appuser
```
guusvw:
```suggestion
ENV USER=deck
```

Dockerfile
guusvw: @hbagdi I (personally) like to have a default command as long as I'm not passing anything to the container and the `help` command is a nice default target,  but I could also remove it

Dockerfile
```diff
@@ -7,7 +7,27 @@ ADD . .
...
+    --no-create-home \
+    --uid "$UID" \
+    "$USER"
+
 COPY --from=build /deck/deck /usr/local/bin
+
+USER $USER
+
 ENTRYPOINT ["deck"]
+CMD ["help"]
```
guusvw:
```suggestion
```

digikin:
There is no reason to add all that group and env into the file. You are only copying a binary into the folder.  Have you looked or ran my dockerfile?

Dockerfile
```diff
@@ -7,7 +7,26 @@ ADD . .
...
-RUN apk --no-cache add ca-certificates
-WORKDIR /root/
+RUN apk update \
+    && apk upgrade \
+    && apk --no-cache add ca-certificates
+
+ARG USER=deck
+ARG GROUP=deck
+ARG UID=9999
+ARG GID=9999
```
guusvw: This allows you to set those variables from outside during the build time of the container @digikin

Dockerfile
```diff
@@ -7,7 +7,26 @@ ADD . .
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o deck
 
 FROM alpine:3.10
-RUN apk --no-cache add ca-certificates
-WORKDIR /root/
+RUN apk update \
+    && apk upgrade \
```
hbagdi: Why is an upgrade necessary here?

Dockerfile
hbagdi: I don't think we really had to template these in the first place. This is a self-contained and small Dockerfile.

No strong feelings here so not changes necessary.

hbagdi:
@guusvw Friendly ping.

hbagdi:
#84 implements this.
Thank you for your PR!

PR #84, opened 2019-10-06
dockerfile: use a non-root user to run decK #82

digikin:


hbagdi:
Since #83 was opened up first, I'll accept that one.

hbagdi:
@digikin Thank you!

Issue #114, opened 2019-12-16
bug: select_tags also tags non-taggable resources


raittes:
Creating some consumer with an jwt_secret using `select_tags`  
results in the `tags` being set for the `jwt_secrets`,  that does not support tags.
Version: `decK v0.7.0 (69bdff4)`

```
# consumer.yaml
format_version: "1.1"
_info:
  select_tags:
  - sometag
consumers:
- username: someuser
  jwt_secrets:
  - algorithm: HS256
    key: somekey
    secret: somesecret
```

Results in:
```
deck sync -s consumer.yaml
creating consumer someuser
creating jwt-auth somekey for consumer 4a79914d-669f-4155-9d04-631396989bf1
Error: 1 errors occurred:
	while processing event: {Create} failed: 400 Bad Request
{
  "message":"schema violation (tags: unknown field)",
  "name":"schema violation",
  "fields":{"tags":"unknown field"},
  "code":2
}
```

Same error for resources: KeyAuths, BasicAuths, HMACAuths, JWTAuths, Oauth2Creds, ACLGroups

hbagdi:
This is not a bug.
Kong 1.4 ships with tagging available on all the resources mentioned above.

However, decK should be backwards compatible to work with older installations of Kong.

raittes:
Sorry, I've just checked docs and there where no tags there.
I'm working with latest kubernetes-ingress-controller that ships Kong 1.3, and I forgot to check newer Kong releases.

PR #115, opened 2019-12-16
fix(file/builder): dont tag non-taggable consumer resources

raittes:
Remove `MustMergeTags` from non-taggable consumer resources
Fix for: https://github.com/hbagdi/deck/issues/114

Resources: KeyAuths, BasicAuths, HMACAuths, JWTAuths, Oauth2Creds, ACLGroups

As this resources don't support tags, Merging Tags doesn't affect them (nil),  
except when using `select_tags`, that enforces the tags.

### DIFF
--- a/file/builder.go
+++ b/file/builder.go
@@ -222,7 +222,6 @@ func (b *stateBuilder) ingestKeyAuths(creds []kong.KeyAuth) error {
 				cred.ID = kong.String(*existingCred.ID)
 			}
 		}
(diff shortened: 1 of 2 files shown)


hbagdi:
As mentioned in #114, these resources can be tagged with Kong 1.4 and above.

This patch in it's current form can't be merged. We can disable the tags for these entities if Kong version is below 1.4.

PR #120, opened 2020-01-14
Exclude consumer plugins when --skip-consumers is on

owlwalks:
Linked issue: https://github.com/hbagdi/deck/issues/119

### DIFF
--- a/dump/dump.go
+++ b/dump/dump.go
@@ -81,6 +81,10 @@ func Get(client *kong.Client, config Config) (*utils.KongRawState, error) {
 	}
 	state.Targets = targets
 
+	if config.SkipConsumers {
+		state.Plugins = ExcludeConsumersPlugins(state.Plugins)
+	}
+
 	if !config.SkipConsumers {
 		consumers, err := GetAllConsumers(client, config.SelectorTags)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


dump/dump.go
```diff
@@ -81,6 +81,10 @@ func Get(client *kong.Client, config Config) (*utils.KongRawState, error) {
 	}
 	state.Targets = targets
 
+	if config.SkipConsumers {
```
hbagdi: nit: Please move this below line 52 to keep it near where the plugins are loaded.

dump/dump.go
```diff
@@ -444,3 +448,15 @@ func GetAllACLGroups(client *kong.Client, tags []string) ([]*kong.ACLGroup, erro
 	}
 	return aclGroups, nil
 }
+
+// ExcludeConsumersPlugins filter out consumer plugins
+func ExcludeConsumersPlugins(plugins []*kong.Plugin) []*kong.Plugin {
```
hbagdi: Please don't export this method. Methods beginning with a capital letter are available outside the package. It makes sense to keep this method internal to the package only.

dump/dump.go
```diff
@@ -444,3 +448,15 @@ func GetAllACLGroups(client *kong.Client, tags []string) ([]*kong.ACLGroup, erro
...
+	var filtered []*kong.Plugin
+	for _, p := range plugins {
+		if p.Consumer != nil && p.Consumer.ID != nil {
```
hbagdi: Please use `utils.Empty()` method here instead:
https://godoc.org/github.com/hbagdi/deck/utils#Empty

hbagdi:
For the record, know one thing that this gets messy if there are plugins that are present on a combination of a consumer and route or consumer and service. There are all sorts of corner cases which decK doesn't handle today.

This change looks good though.

owlwalks:
Yea, would be nicer if this is supported by kong admin api, I just pushed some changes regarding your comments.

PR #137, opened 2020-03-31
feat(state) support multiple files in args

raittes:
### Objective: 
I have different state-directories for each environment (ex: production, staging)
And I want to have common configuration for both environments, 
so I could create a symlink for a **common** directory inside each environment directory.
(ex: production/common -> ../common)

### Problem:
Currently, deck-cli state flag (--state) supports only one argument that can be a file or directory.
As [filepath.Walk](https://github.com/hbagdi/deck/blob/0da5af91305f118ea1ab9aae6955628a8b4f7427/file/readfile.go#L102) does [not follow symlinks](https://golang.org/src/path/filepath/path.go?s=12220:12265#L400), deck will ignore symlinks found in the state directory.

### Solution:
Supporting multiple arguments in state flag, we could load multiple state from different locations, as:
```
deck diff -s common -s production
or 
deck diff -s common,production
```

### DIFF
--- a/cmd/sync.go
+++ b/cmd/sync.go
@@ -32,7 +32,8 @@ to get Kong's state in sync with the input state.`,
 func init() {
 	rootCmd.AddCommand(syncCmd)
 	syncCmd.Flags().StringSliceVarP(&syncCmdKongStateFile,
-		"state", "s", []string{"kong.yaml"}, "file containing Kong's configuration. "+
+		"state", "s", []string{"kong.yaml"}, "file(s) containing Kong's configuration.\n"+
+			"This flag can be specified multiple times for multiple files.\n"+
 			"Use '-' to read from stdin.")
 	syncCmd.Flags().BoolVar(&dumpConfig.SkipConsumers, "skip-consumers",
 		false, "do not diff consumers or "+
--- a/cmd/sync.go
+++ b/cmd/sync.go
@@ -6,7 +6,7 @@ import (
 )
 
 var (
-	syncCmdKongStateFile string
+	syncCmdKongStateFile []string
 	syncCmdParallelism   int
 )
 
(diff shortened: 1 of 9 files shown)


hbagdi:
Can you please add a test case in the `file` package?

raittes:
done @hbagdi

cmd/sync.go
```diff
@@ -31,8 +31,8 @@ to get Kong's state in sync with the input state.`,
 
 func init() {
 	rootCmd.AddCommand(syncCmd)
-	syncCmd.Flags().StringVarP(&syncCmdKongStateFile,
-		"state", "s", "kong.yaml", "file containing Kong's configuration. "+
+	syncCmd.Flags().StringSliceVarP(&syncCmdKongStateFile,
+		"state", "s", []string{"kong.yaml"}, "file containing Kong's configuration. "+
```
hbagdi, line 35: Let's change the help test to instead:
```
file(s) containing Kong's configuration.
This flag can be specified multiple times for multiple files.
Use '-' to read from stdin.
```

Please note the line jumps. We want that in the `--help` output. 
Please change this in all other commands as well.

file/reader_test.go
```diff
@@ -40,9 +40,9 @@ func Test_ensureJSON(t *testing.T) {
 }
 
 func TestReadKongStateFromStdinFailsToParseText(t *testing.T) {
-	var filename = "-"
+	var filenames = []string{"-"}
 	assert := assert.New(t)
-	assert.Equal("-", filename)
+	assert.Equal("-", filenames[0])
```
hbagdi, line 45: There is an interesting case where some configuration can be specified via Stdin and other via flag.
I was considering to restrict `-` as standalone, meaning you either specify only file(s) or only Stdin.
I think it is okay to allow it.
Just leaving a thought. No changes required.

raittes:
Thank you @hbagdi!
When it will be released?

hbagdi:
Expect a release in a couple of weeks.
There are a few other features in progress.

Issue #113, opened 2019-12-12
feature: error out on invalid positional args


hbagdi:
```
deck sync foo.yaml
```

The above is an invalid command and should error out. This is not intuitive to all users.

malhotraamit236:
@hbagdi Raised a PR for this. Let me know your thoughts. https://github.com/hbagdi/deck/pull/146

PR #146, opened 2020-04-30
feature: #113 invalid postional args

malhotraamit236:
@hbagdi let me know if this is what you meant to achieve for this feature.
I considered an alternative: `Args: cobra.NoArgs` but it gives a generic error. That's why I used custom validator approach to customize error message

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -4,6 +4,8 @@ import (
 	"net/http"
 	"os"
 
+	"github.com/spf13/cobra"
+
 	"github.com/blang/semver"
 	"github.com/fatih/color"
 	"github.com/hbagdi/deck/diff"
@@ -141,4 +143,13 @@ func kongVersion(config utils.KongClientConfig) (semver.Version, error) {
 
 	v, err := utils.CleanKongVersion(root["version"].(string))
 	return semver.ParseTolerant(v)
+}
+
+func validateNoArgs(cmd *cobra.Command, args []string) error {
+	if len(args) > 0 {
+		return errors.New("This command cannot take any positional arguments. " +
+			"Try using a flag instead.\n" +
+			"For usage information: deck <command> -h")
+	}
+	return nil
 }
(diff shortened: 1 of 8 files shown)


cmd/diff.go
```diff
@@ -21,6 +21,14 @@ It will load entities form Kong and then perform a diff on those with
 the entities present in files locally. This allows you to see the entities
 that will be created or updated or deleted.
 `,
+	Args: func(cmd *cobra.Command, args []string) error {
+		if len(args) > 0 {
+			return errors.New("diff command cannot take any positional arguments. " +
+				"Try using a flag instead.\n" +
+				"For usage information: deck diff -h")
+		}
```
hbagdi, line 29: Let's not repeat here and provide the same error message for all commands.
The Go way to do this would be to create a function definition inside the cmd package and then reference it in all places.

malhotraamit236: @hbagdi thanks for the feedback. I have refactored args validation into a `validateNoArgs` func in `cmd/common.go` and now referencing that func in all commands.
The common error message is: `Error: This command cannot take any positional arguments. Try using a flag instead.
For usage information: deck <command> -h`

cmd/common.go
```diff
@@ -142,3 +144,12 @@ func kongVersion(config utils.KongClientConfig) (semver.Version, error) {
 	v, err := utils.CleanKongVersion(root["version"].(string))
 	return semver.ParseTolerant(v)
 }
+
+func validateNoArgs(cmd *cobra.Command, args []string) error {
+	if len(args) > 0 {
+		return errors.New("This command cannot take any positional arguments. " +
+			"Try using a flag instead.\n" +
+			"For usage information: deck <command> -h")
```
hbagdi, line 152: This is very nitpicky but is important for a uniform developer experience throughout the tool.
Let's change the error to:
> positional arguments are not valid for this command, please use flags instead
Run 'deck --help' for usage.

Case-sensitivity also matters here.

Other than this, this is ready for merge.

malhotraamit236: Will make the change and push today.

malhotraamit236: Done

PR #157, opened 2020-05-26
feat(sync/diff) Add workspace arg to sync and diff commands

cwurm:
Allow specifying a workspace with `-w / --workspace` for the `sync` and `diff` commands - similar to how this is possible today for `dump` and `reset`.

### Use case(s)
1. Enables rolling out the same configuration across workspaces, e.g. the same logging plugin.
2. Easier sharing of deck configurations between Kong deployments (e.g. between different organizations, or different parts of the same organization), where the eventual workspace names cannot be known beforehand.

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -11,6 +11,7 @@ import (
 	"github.com/hbagdi/deck/diff"
 	"github.com/hbagdi/deck/dump"
 	"github.com/hbagdi/deck/file"
+	"github.com/hbagdi/deck/print"
 	"github.com/hbagdi/deck/solver"
 	"github.com/hbagdi/deck/state"
 	"github.com/hbagdi/deck/utils"
@@ -81,7 +82,9 @@ func syncMain(filenames []string, dry bool, parallelism, delay int, workspace st
 		return err
 	}
 	// prepare to read the current state from Kong
-	if len(workspace) > 0 {
+	if workspace != "" {
+		print.DeletePrintf("Warning: Specified workspace '%v' is different from "+
+			"workspace '%v' found in state file(s)\n", workspace, targetContent.Workspace)
 		config.Workspace = workspace
 	} else {
 		config.Workspace = targetContent.Workspace
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -82,9 +82,9 @@ func syncMain(filenames []string, dry bool, parallelism, delay int, workspace st
 		return err
 	}
 	// prepare to read the current state from Kong
-	if workspace != "" {
-		print.DeletePrintf("Warning: Specified workspace '%v' is different from "+
-			"workspace '%v' found in state file(s)\n", workspace, targetContent.Workspace)
+	if workspace != targetContent.Workspace {
+		print.DeletePrintf("Warning: Workspace '%v' specified via --workspace flag is "+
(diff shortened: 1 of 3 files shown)


hbagdi:
Discussing (IRL) an alternate way to handle this with @cwurm so holding on this right now.
I'm also thinking about risks associated with this.
This addition makes sense but on the other hand I had intentionally not implemented this before to ensure that the entire state of the world is reflected in the state file.

hbagdi:
@cwurm ping. Do we still need this feature?

cmd/common.go
```diff
@@ -80,7 +80,11 @@ func syncMain(filenames []string, dry bool, parallelism int) error {
 		return err
 	}
 	// prepare to read the current state from Kong
-	config.Workspace = targetContent.Workspace
+	if len(workspace) > 0 {
```
hbagdi, line 83: The more idiomatic way to do this `workspace != ""`.

hbagdi: Please also add a warning (using print.DeletePrintln from print package in this repo) when the workspace supplied via cli and that supplied via state file do not match up.

cmd/sync.go
```diff
@@ -36,6 +37,10 @@ func init() {
 		"state", "s", []string{"kong.yaml"}, "file(s) containing Kong's configuration.\n"+
 			"This flag can be specified multiple times for multiple files.\n"+
 			"Use '-' to read from stdin.")
+	syncCmd.Flags().StringVarP(&syncWorkspace, "workspace", "w",
```
hbagdi, line 40: Let's remove the `-w` shorthand flag. This flag will not be used often and `-w` will hamper readability of commands later down the road.

cmd/common.go
cwurm: Done.

cmd/sync.go
cwurm: Done.

cwurm:
@hbagdi Thanks, I've incorporated the review comments and rebased to solve conflicts that were introduced since opening this PR.

cmd/common.go
```diff
@@ -81,7 +82,13 @@ func syncMain(filenames []string, dry bool, parallelism, delay int) error {
 		return err
 	}
 	// prepare to read the current state from Kong
-	config.Workspace = targetContent.Workspace
+	if workspace != "" {
+		print.DeletePrintf("Warning: Specified workspace '%v' is different from "+
+			"workspace '%v' found in state file(s)\n", workspace, targetContent.Workspace)
```
hbagdi, line 87: bug: Even if the `--workspace` provides the same workspace as the one in the state file(s), this warning will still be printed.
The right thing to do here is to first check if targetContent.Workspace and workspace are not same and then override config.Workspace as necessary.

hbagdi: > Specified workspace

That's generic. May be "Workspace via --workspace flag '%v' ..."

cwurm: Thanks, fixed. The only edge case now is the default workspace which can be expressed as `default` and as an empty string. Let me know if you want to handle that as well and I'll add logic for it.

hbagdi: > The only edge case now is the default workspace which can be expressed as default and as an empty string
 
It is not an edge case as long as user doesn't mix the two. If they do mix, they are doing it wrong and we can correct it.

PR #166, opened 2020-06-03
fix(file) populate port according to protocol in url

hbagdi:
Populate port according to the protocol scheme in the `url` sugar
attribute. Default port of `80` is now populated to match Kong's
behavior in the same case.


hbagdi:
cc @rainest

PR #177, opened 2020-06-26
fix(cmd) report a descriptive error when workspace is not present in kong

### LINKED ISSUE #169: bug: checkWorkspace doesn't display the appropriate error
The first err check results in true and subsequent err checks never happen:

https://github.com/hbagdi/deck/blob/master/cmd/common.go#L54-L65

hbagdi:
Fix #169

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -46,21 +46,22 @@ func checkWorkspace(config utils.KongClientConfig) error {
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


hbagdi:
Merged in https://github.com/hbagdi/deck/commit/90722b61e6f32222b7230240f96dd9302bc95376

PR #182, opened 2020-07-01
Remove required jwt-auth fields from schema

### LINKED ISSUE #181: deck enforces a secret for RS256 JWT auth details
Trying to create a consumer with JWT credentials using the RS256 algorithm. For the RS256 Algorithm, the secret field is non-mandatory. But, deck sync command mandates to provide the field, as given below

`# deck sync --headers kong-admin-token:my-token
	consumers.0.jwt_secrets.0: secret is required`

The admin API does allow a null secret to be used with the RS256 algorithm. I tried editing the kong.yaml file and setting the secret to nothing, null and an empty string and all gave the same error i.e. secret is required

rainest:
Per https://docs.konghq.com/hub/kong-inc/jwt/#create-a-jwt-credential
jwt-auth does not truly have required fields. It has pseudo-required
fields based on the value of other fields and/or will auto-populate
fields with random values.

Both of these concepts are too complex to express in decK's schema, so
removing all validation logic from it.

Fix #181

### DIFF
--- a/file/codegen/main.go
+++ b/file/codegen/main.go
@@ -91,8 +91,7 @@ func main() {
 	schema.Definitions["ACLGroup"].Required = []string{"group"}
 	schema.Definitions["BasicAuth"].Required = []string{"username", "password"}
 	schema.Definitions["HMACAuth"].Required = []string{"username", "secret"}
-	schema.Definitions["JWTAuth"].Required = []string{"algorithm", "key",
-		"secret"}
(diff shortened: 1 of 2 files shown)


hbagdi:
As discussed offline, we don't allow users to auto-gen any fields as we can't have deterministic declarative behavior in that case.

rainest:
Won't merge, see similar discussion in https://github.com/Kong/kubernetes-ingress-controller/pull/754#issuecomment-653141411

PR #183, opened 2020-07-04
fix: Deduplicate tags before sending to Kong

cwurm:
When getting entities from Kong, decK will restrict the query to the tags in the state files. However, when using multiple state files, it will AND all tags it finds together - regardless of whether they are unique tags or just the same tags.

When the total number of tags is above 5, Kong will reject the query outright.

This PR deduplicates tags before they are sent to Kong. This allows having the same tag(s) in every file (to make sure it is never synced tag-less) - as long as the total unique number is not above 5.

### DIFF
--- a/dump/dump.go
+++ b/dump/dump.go
@@ -20,10 +20,24 @@ type Config struct {
 	SelectorTags []string
 }
 
+func deduplicate(stringSlice []string) []string {
+	existing := map[string]struct{}{}
+	result := []string{}
+
+	for _, s := range stringSlice {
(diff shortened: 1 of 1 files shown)


hbagdi:
Although decK merges the tags from all files, that was not what I had intended. I was aware of the behavior but I was too lazy to fix it, which has now come back to bite me. Another lesson that limit the room for interpretation to the minimum in a user facing API. I've learned this lesson so many times but it just doesn't stick in my head. (:rage:, Harry) 

Thanks for the fix here @cwurm!

Issue #185, opened 2020-07-10
Segregation problem using decK and RBAC users


nutellinoit:
When using an RBAC user with the default role `workspace-super-admin`  `deck sync` fails. `deck dump` is working against the workspace. (using kong 1.5.0.3-enterprise-edition)

My idea is that the RBAC users (also with custom roles attached) are not able to get the root path of the admin api `/` and the `workspaces/`

Error during sync: 

```bash
deck sync --config deck.yml --verbose 10                                                  
GET / HTTP/1.1                                                                                                                                                                                          
Host: 192.168.1.182:8001                                                                                                                                                                                
User-Agent: Go-http-client/1.1
Accept-Encoding: gzip


HTTP/1.1 403 Forbidden
Content-Length: 77
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: http://192.168.1.182:8002
Connection: keep-alive
Content-Type: application/json; charset=utf-8
Date: Fri, 10 Jul 2020 07:28:57 GMT
Server: kong/1.5.0.3-enterprise-edition
Vary: Origin
X-Kong-Admin-Latency: 36
X-Kong-Admin-Request-Id: nW2lXw5aKi43kQE8Q9uRcU4eVEXO0eC9

{"message":"samuele-admin, you do not have permissions to read this resource"}
Error: reading Kong version: 403 Forbidden {"message":"samuele-admin, you do not have permissions to read this resource"}
```

I'm creating a PR with some modification as a starting point to discuss this issue.

PR #186, opened 2020-07-10
Added get version from /:workspace/kong and added flag skip-workspace-crud to skip workspace existence

nutellinoit:
Following #185 

Added some parameters to declare kong version and skip-check (workspace existence for now)

Example `deck.yaml`:

```yaml
kong-addr: http://192.168.1.182:8001
version: 1.5.0.3-enterprise-edition
skip-check: true
headers:
  - "Kong-Admin-Token:000000"
```

This way an RBAC user with workspace-superadmin default role is able to do a `deck sync`


hbagdi:
Hey @nutellinoi,
Thanks for the issue and the PR and explaining the rationale behind this.

I agree with the overall approach here. Some comments:
- We should not ask users to provide the version of Kong. This is error prone and asks users to do more than needed. Do you know that there exists a `/workspaces/:workspace/kong` endpoint which returns back the same responses as `/` on the Admin API? We can use that to figure out the version of Kong. decK is written to be user friendly and I think keeping it so will help.
- `--skip-check` is a very generic name. Can we rename this to `--skip-workspace-crud`? This should apply to `diff` and `sync` commands, and not the root command.
- Please make sure `deck ping` also works with the reduced set of permissions.


Please be aware that #187 is also being worked on that is at odds with your use case. We are going to make sure that both are merged in and are compatible with each other.

nutellinoit:
Thank you @hbagdi i'll do the suggested changes 😊

nutellinoit:
@hbagdi , is `/workspaces/:workspace/kong` available in the 1.5.x.x version of Kong enterprise?

nutellinoit:
Hi @hbagdi , the admin path needed is `/:workspace/kong` 😊  

I was able to make it work with this path if `/` is not giving any valid response.

I also changed the deck.yml flag/key to `skip-workspace-crud` as suggested:

```yaml
kong-addr: http://192.168.1.182:8001
skip-workspace-crud: true
headers:
  - "Kong-Admin-Token:000000"
```

I also did some modification to "ping" command to make it work with RBAC users

cmd/ping.go
```diff
@@ -2,12 +2,16 @@ package cmd
...
 	"github.com/spf13/cobra"
 )
 
+var (
+	pingCmdKongStateFile []string
+)
+
 // pingCmd represents the ping command
 var pingCmd = &cobra.Command{
 	Use:   "ping",
```
hbagdi, line 17: The entire purpose of `ping` command is to verify connectivity with Kong. It shouldn't take state file as input. What we can do is instead add in a `--workspace` flag as proposed in #157

cmd/common.go
```diff
@@ -138,21 +139,40 @@ func syncMain(filenames []string, dry bool, parallelism, delay int) error {
...
+		}
+		return semver.ParseTolerant(v)
 	}
 
 	v, err := utils.CleanKongVersion(root["version"].(string))
 	if err != nil {
 		return semver.Version{}, err
 	}
 	return semver.ParseTolerant(v)
+
```
hbagdi, line 175: Please remove this line jump

cmd/common.go
```diff
@@ -1,11 +1,10 @@
 package cmd
 
 import (
+	"github.com/spf13/cobra"
```
hbagdi, line 4: Although gofmt doesn't complain, this should be placed with the other github.com/... imports below.

cmd/ping.go
nutellinoit: Ok i should do like the other PR, thank you @hbagdi

nutellinoit:
@hbagdi done! ✅

cmd/common.go
```diff
@@ -138,14 +139,32 @@ func syncMain(filenames []string, dry bool, parallelism, delay int) error {
...
+		_, err = client.Do(nil, req, &resp)
+		if err != nil {
+			return semver.Version{}, err
+		}
+
+		v, err := utils.CleanKongVersion(resp["version"].(string))
+		if err != nil {
+			return semver.Version{}, err
+		}
+		return semver.ParseTolerant(v)
```
hbagdi, line 167: This piece of code is being repeated again at the bottom. Can you store the version from this call or the previous call on line 148 in a common variable and then use the code below? It would be better to avoid repetition here.

cmd/ping.go
```diff
@@ -36,4 +46,8 @@ can connect to Kong's Admin API or not.`,
 
 func init() {
 	rootCmd.AddCommand(pingCmd)
+	pingCmd.Flags().StringVarP(&pingWorkspace, "workspace", "w",
+		"", "Ping configuration with a specific workspace "+
+			"(Kong Enterprise only).\n"+
+			"Use this with an RBAC user.")
```
hbagdi, line 52: `"Useful when RBAC permissions are scoped to a workspace"` might be more fitting.

cmd/root.go
```diff
@@ -121,6 +121,11 @@ func init() {
 		"disable colorized output")
 	viper.BindPFlag("no-color",
 		rootCmd.PersistentFlags().Lookup("no-color"))
+
+	rootCmd.PersistentFlags().Bool("skip-workspace-crud", false,
+		"Skip Workspace CRUD checks")
+	viper.BindPFlag("no-color",
```
hbagdi, line 127: copy-pasta `no-color`.

cmd/root.go
```diff
@@ -121,6 +121,11 @@ func init() {
 		"disable colorized output")
 	viper.BindPFlag("no-color",
 		rootCmd.PersistentFlags().Lookup("no-color"))
+
+	rootCmd.PersistentFlags().Bool("skip-workspace-crud", false,
+		"Skip Workspace CRUD checks")
```
hbagdi, line 126: Please change this to `Skip API calls related to Workspaces (Kong Enterprise only)`.

cmd/ping.go
```diff
@@ -22,7 +25,14 @@ can connect to Kong's Admin API or not.`,
...
 		if err != nil {
-			return errors.Wrap(err, "connecting to kong")
+			config.Workspace = pingWorkspace
+			version, err := kongVersion(config)
+			if err != nil {
+				return errors.Wrap(err, "reading Kong version")
+			}
+			fmt.Println("Successfully connected to Kong!")
+			fmt.Println("Kong version: ", version)
+			return nil
```
hbagdi, line 35: The call on line 26 is not needed in this case since client.Root() is already called in `kongVersion()`.
Please simplify this.

nutellinoit: Hello @hbagdi , ping command simplified!

cmd/common.go
nutellinoit: Code optimized!

cmd/common.go
```diff
@@ -138,17 +139,32 @@ func syncMain(filenames []string, dry bool, parallelism, delay int) error {
...
 		return semver.Version{}, err
 	}
-
 	root, err := client.Root(nil)
 	if err != nil {
-		return semver.Version{}, err
-	}
-
-	v, err := utils.CleanKongVersion(root["version"].(string))
+		// try with workspace path
```
hbagdi, line 150: Please guard this block with another if:
```go
if config.Workspace != ""
```

If workspace is not provided, then return back an error.

hbagdi:
Also since #187 was merged in first, you will have to rework quite a bit of logic. I was hoping to merge this in and then ask @cwurm to do the extra work but it turned out the other way.

Please be careful with the merge, the logic is going to get elaborate with a lot of conditional statements.

nutellinoit:
Hi @hbagdi , i merged everything from main branch, i changed the logic accordingly to #187

hbagdi:
Merged manually.

PR #187, opened 2020-07-12
feat(sync/diff) Create workspace if needed

cwurm:
Adds the workspace as another entity to decK.

Some notes on the implementation:
1. A workspace is only ever created, not updated or deleted. Updating is not really necessary since the deck syntax at the moment only supports the workspace name in the `_workspace` property. Deleting a workspace is quite tricky esp. when the Dev Portal is enabled.
2. Supports both `sync` and `diff` - other workspace-related behavior e.g. in `reset` and `dump` remains unchanged (though some code has changed).
3. Updates to the current `master` of `go-kong` for workspace support.
4. Since `go-kong` does not offer complete support for workspaces a separate `rootClient` is needed to create a workspace. Fortunately, the code changes for that were not too bad.
5. The changes in `file/schema.go` are not from this PR - somewhere something changed and it wasn't updated.

Example output:
```ShellSession
$ deck sync -s workspace-demo.yaml -s openid-connect-okta.yaml
creating workspace Demo
creating service mockbin-okta
creating route okta
creating plugin openid-connect for route 8a7f17df-2998-4c71-9da8-0b75ff72373b
Summary:
  Created: 4
  Updated: 0
  Deleted: 0
```

Resolves https://github.com/hbagdi/deck/issues/158.


cmd/common.go
```diff
@@ -33,38 +32,38 @@ func SetStopCh(stopCh chan struct{}) {
...
-	}
-	resp, err := client.Do(nil, req, nil)
-	if resp != nil && resp.StatusCode == http.StatusNotFound {
-		return errors.Errorf("workspace '%v' does not exist in Kong, "+
-			"please create it before running decK.", workspace)
-	}
+	// Using ListAll() to get all workspaces instead of Get() to get just
+	// the one to make sure there is no underlying error - Get() does not
+	// expose the response.
+	workspaces, err := rootClient.Workspaces.ListAll(nil)
```
hbagdi, line 54: > the one to make sure there is no underlying error

Yes, this is a limitation and go-kong needs to improve to fix this problem but I still think we should use Get() instead. You can use [IsNotFoundErr()](https://godoc.org/github.com/hbagdi/go-kong/kong#IsNotFoundErr) to check if it was a 404 or not.

If there is an underlying error, return it.

hbagdi:
This is great work and provides a very good end-user experience but I feel this is more complicated than it should be.

- We can get away with simulating a workspace sync rather than add it to solver, diff and setup the whole machinery around it. This is possible because we are only ever going to create a workspace and never going to update or delete it.
- Another reason to avoid adding this in is because we are not at a point where we would like to manage workspaces using decK. If this PR is merged as is, it will open gates for adding declarative configuration of workspaces entity in Kong itself, which is plagued with problems and I think it will be best for decK to not get anywhere near that dragon.
- We could do the following, which will result in a smaller and more maintainable patch:
  - Diff
    - check if workspace exists, if not then run a diff against an empty state (the current patch in the PR already does this)
    - if workspace exists, then dump and then do the diff
  - Sync
    - if workspace doesn't exist, create it and then start the dump and diff logic
    - if does exist, then nothing needs to be done

Also, please also take a look at #186 which will need to work with this.

hbagdi:
> Sure, so that we be something close to the first commit (which already worked for sync just needs to be extended to diff): ebad319 (#187)

Pretty much. The only thing to note is that `deck diff` should not create a workspace in kong.

> The reason I expanded to the full solver/syncer is that without it the workspace doesn't appear in the entity counter at the end. But if that's not a concern I agree there is a much simpler way to do this and I'll just go back to the first commit and expand it. Let me know.

That's could be considered a little odd but if we are taking the stance that workspace management is not something deck is responsible for, that's okay in my opinion.

> But - this PR alone does not commit us in any way to go in this direction - there never has to be a further change to this code.

This PR doesn't intend to do that but in the current for, it has the potential to give new contributors ideas around what could be done with workspaces and decK. OSS users can and do sometimes get very creative, which is not always good from a Software Engineering standpoint. 

Ping me once you have reworked this one. Thanks!

cmd/common.go
cwurm: Makes sense, and done.

cwurm:
@hbagdi Alright, I've simplified the logic.

The output is now:
```ShellSession
$ deck sync -s workspace-demo.yaml -s openid-connect-okta.yaml
creating workspace Demo
creating service mockbin-okta
creating route okta
creating plugin openid-connect for route ff9dcb66-0aae-45ef-abad-c627ebd2201c
Summary:
  Created: 3
  Updated: 0
  Deleted: 0
```

The created workspace no longer appears as a created entity in the summary - but the code is a lot simpler.

PR #200, opened 2020-08-11
fix: will panic on sendAnalytics faild

marlonfan:
there will panic nil point err


FonzMP:
By chance, would we have an anticipated release date with these changes?

PR #198, opened 2020-08-10
fix: Break infinite wait when encountering many errors

cwurm:
I encountered a condition where decK would hang up indefinitely - after processing the first few entities it would not continue and not even a Ctrl+C would terminate it.

I started looking into it and a `kill -ABRT` showed it was in an infinite loop in `diff/diff.go`:
```ShellSession
$ deck sync -s .
creating consumer acl22
creating consumer gold-partner
creating consumer jwt-user
creating consumer acl11
creating consumer myConsumer
creating consumer oauthApp
creating consumer external
creating consumer acl12
creating consumer partner
creating consumer silver-partner
creating consumer oauthUser
^Creceived interrupt , terminating...
SIGABRT: abort
PC=0x7fff69344882 m=0 sigcode=0

goroutine 0 [idle]:
runtime.pthread_cond_wait(0x1b044a8, 0x1b04468, 0x7ffe00000000)
	/usr/local/Cellar/go/1.14.6/libexec/src/runtime/sys_darwin.go:390 +0x39
runtime.semasleep(0xffffffffffffffff, 0x7ffeefbfeda0)
	/usr/local/Cellar/go/1.14.6/libexec/src/runtime/os_darwin.go:63 +0x85
runtime.notesleep(0x1b04268)
	/usr/local/Cellar/go/1.14.6/libexec/src/runtime/lock_sema.go:173 +0xe0
runtime.stoplockedm()
	/usr/local/Cellar/go/1.14.6/libexec/src/runtime/proc.go:1977 +0x88
runtime.schedule()
	/usr/local/Cellar/go/1.14.6/libexec/src/runtime/proc.go:2460 +0x4a6
runtime.park_m(0xc000001c80)
	/usr/local/Cellar/go/1.14.6/libexec/src/runtime/proc.go:2696 +0x9d
runtime.mcall(0x1061e16)
	/usr/local/Cellar/go/1.14.6/libexec/src/runtime/asm_amd64.s:318 +0x5b

goroutine 1 [semacquire]:
sync.runtime_Semacquire(0xc000210008)
	/usr/local/Cellar/go/1.14.6/libexec/src/runtime/sema.go:56 +0x42
sync.(*WaitGroup).Wait(0xc000210000)
	/usr/local/Cellar/go/1.14.6/libexec/src/sync/waitgroup.go:130 +0x64
github.com/kong/deck/cmd.Execute()
	/Users/cwurm/git/hbagdi/deck/cmd/root.go:64 +0xc8
main.main()
[...]

PR #202, opened 2020-08-14
fix(sync) resolve concurrenct wait bug

hbagdi:
Background
----------

Syncer works with a single producer, multiple consumers with a queue in
between them which has a constant buffer size of 10:
`const eventBuffer = 10`
Each time the producer ATTEMPTS to add an item into queue, inflightOps
is incremented atomically unconditionally inside queueEvent().
A consumer decrements the counter once it processes an event, no matter
the event was processed successfully or failed.

A consumer's thread (goroutine) exits once the event channel is closed.
This channel is closed by the producer when the producer is done.

Problem
-------

There are two problems:

1. If a producer's queue is full and `stopChan` is closed,
`queueEvent()` incorrectly increments inflightOps and returns an error,
which causes the producer to exit and eventually all consumers to exit.
This causes the inFlightOps to be incremented incorrectly but doesn't
explain the deadlock.
2. If SIGKILL is sent by the user and producer is stuck at `wait()`
because kong is taking time to respond to requests, `wait()` blocks
indefintely and the process appears to be stuck. Given enough time, this
should resolve and the process should eventually exit once all the
timeouts are reached. If this is not the case, we are dealing with yet
another unknown bug.

Solution
--------

Fix problem (1) by decrementing the counter when the counter increment
was not successful. The counter has to be incremented before sending to
the channel otherwise the producer can wrongly conclude that all
outstanding events are complete in `wait()`.

Fix problem (2) by returning from `wait()` if stopChan is closed.

There could be more such deadlocks lurking in the code and we have
simply gotten lucky that they never show up.
In the long-run, we need to refactor the diff.go file for readability
and maintainance using packages such as `github.com/hbagdi/gang`
which simplify goroutine management.

See #198

Co-authored-by: Christoph Wurm <christoph@konghq.com>

### DIFF
--- a/diff/diff.go
+++ b/diff/diff.go
@@ -270,6 +270,7 @@ func (sc *Syncer) queueEvent(e Event) error {
 	case sc.eventChan <- e:
 		return nil
 	case <-sc.stopChan:
+		atomic.AddInt32(&sc.inFlightOps, -1)
 		return errEnqueueFailed
 	}
 }
@@ -281,8 +282,12 @@ func (sc *Syncer) eventCompleted() {
 func (sc *Syncer) wait() {
 	time.Sleep(time.Duration(sc.StageDelaySec) * time.Second)
 	for atomic.LoadInt32(&sc.inFlightOps) != 0 {
-		// TODO hack?
-		time.Sleep(1 * time.Millisecond)
+		select {
+		case <-sc.stopChan:
+			return
+		default:
+			time.Sleep(1 * time.Millisecond)
+		}
 	}
 }
 


cwurm (approved):
Makes sense to decrement the counter as well. It does not mean `inFlightOps` will always be at 0 at the end - since all consumers might have exited after each encountering an error. In my reproduction of the bug even with this fix `inFlightOps` is still `3` at the end of `Run()`.

diff/diff.go
```diff
@@ -281,8 +282,12 @@ func (sc *Syncer) eventCompleted() {
 func (sc *Syncer) wait() {
 	time.Sleep(time.Duration(sc.StageDelaySec) * time.Second)
 	for atomic.LoadInt32(&sc.inFlightOps) != 0 {
-		// TODO hack?
-		time.Sleep(1 * time.Millisecond)
+		select {
+		case <-sc.stopChan:
+			return
```
cwurm, line 287: Thinking about this - could it ever happen that `stopChan` is closed while we are still waiting for legitimate operations, e.g. if we are syncing up 100s of entities and it is taking a while? The `close(sc.stopChan)` statement seems pretty standalone.

hbagdi: This should still be okay because returning from here will make the producer stop. The consumers will still continue to run to completion (either successfully or error out).

hbagdi:
> Makes sense to decrement the counter as well. It does not mean inFlightOps will always be at 0 at the end - since all consumers might have exited after each encountering an error. In my reproduction of the bug even with this fix inFlightOps is still 3 at the end of Run().

inFlightOps being non-zero is harmless but there is another bug that we need to fix it.


The behavior we want to have is: if there is an error with any operation against kong's api, error out and stop the sync process. The current behavior is buggy where it only stops the single consumer which encountered the error. Replacing the spaghetti code in diff.go with a simpler version using [gang](https://github.com/hbagdi/gang) for managing goroutine life-cycle.

PR #207, opened 2020-09-01
feat: Read multi-document YAML

cwurm:
This change allows decK to read multi-document YAML files correctly, i.e. all contents/documents. Currently, only the first document is read.

While decK will not write out multi-document YAML itself, this makes it easier to work with third-party YAML e.g. generated by tools such as `ytt`. My concrete use case was generating a very large Kong configuration using `ytt` and I could find no straightforward way around it generating a multi-document YAML.

Some notes on the implementation:
1. While the go-yaml README [says](https://github.com/go-yaml/yaml/blob/v2/README.md) it does not support multi-document unmarshaling it actually does. There is an open [issue](https://github.com/go-yaml/yaml/issues/361) and [PR](https://github.com/go-yaml/yaml/pull/425) to clarify this.
2. Rather than add a new testfile I modified an existing testfile to be multi-document. We could add a new testfile as well, but this was simpler.
3. The `Content.String()` function and modification to `readfile_test.go` is just to make the test failure output much easier to consume - formatted YAML instead of a bunch of (mostly `<nil>`) pointers.

### DIFF
--- a/file/reader.go
+++ b/file/reader.go
@@ -84,6 +84,21 @@ func yamlToJSON(m map[interface{}]interface{}) map[string]interface{} {
 		switch v2 := v.(type) {
 		case map[interface{}]interface{}:
 			res[fmt.Sprint(k)] = yamlToJSON(v2)
+		case []interface{}:
+			var array []interface{}
+			for _, element := range v2 {
+				switch el := element.(type) {
+				case map[interface{}]interface{}:
+					array = append(array, yamlToJSON(el))
+				default:
+					array = append(array, el)
+				}
+			}
+			if array != nil {
+				res[fmt.Sprint(k)] = array
+			} else {
+				res[fmt.Sprint(k)] = v
+			}
 		default:
 			res[fmt.Sprint(k)] = v
 		}
--- a/file/reader_test.go
+++ b/file/reader_test.go
@@ -99,12 +99,13 @@ func TestReadKongStateFromStdin(t *testing.T) {
 	os.Stdin = tmpfile
 
 	c, err := GetContentFromFiles(filenames)
-	assert.NotNil(c)
 	assert.Nil(err)
 
-	assert.Equal(kong.Service{
-		Name: kong.String("test service"),
-		Host: kong.String("test.com"),
-	},
-		c.Services[0].Service)
+	if assert.NotNil(c) && assert.NotEmpty(c.Services) {
+		assert.Equal(kong.Service{
+			Name: kong.String("test service"),
+			Host: kong.String("test.com"),
+		},
+			c.Services[0].Service)
+	}
 }
--- a/file/readfile.go
+++ b/file/readfile.go
@@ -3,7 +3,6 @@ package file
 import (
 	"bufio"
 	"io"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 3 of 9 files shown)


cwurm:
This currently disables JSON schema validation. I'll think about how to keep it.

cwurm:
Finally had time to continue on this. Neither of the two libraries involved in schema validation (`github.com/ghodss/yaml` and `github.com/xeipuuv/gojsonschema`) support multi-document YAML, unfortunately.

I could keep `gojsonschema` and just run it in a loop over the documents in a file. I didn't see a way to keep `ghodss/yaml` but did not immediately see why we need it (e.g. we are not using the JSON struct tags and custom marshalling methods it enables).

@hbagdi can you have a look?

hbagdi:
`ghodss/yaml` is used because it marshals correctly from YAML to JSON.

file/reader.go
```diff
@@ -84,6 +84,21 @@ func yamlToJSON(m map[interface{}]interface{}) map[string]interface{} {
...
+					array = append(array, yamlToJSON(el))
+				default:
+					array = append(array, el)
+				}
+			}
+			if array != nil {
+				res[fmt.Sprint(k)] = array
+			} else {
+				res[fmt.Sprint(k)] = v
+			}
```
hbagdi, line 101: When would this else execute?

file/reader.go
```diff
@@ -84,6 +84,21 @@ func yamlToJSON(m map[interface{}]interface{}) map[string]interface{} {
 		switch v2 := v.(type) {
 		case map[interface{}]interface{}:
 			res[fmt.Sprint(k)] = yamlToJSON(v2)
+		case []interface{}:
```
hbagdi, line 87: Why was this necessary?

(I've a vague memory that this is what ghodss/yaml took care of for us).

file/reader_test.go
```diff
@@ -99,12 +99,13 @@ func TestReadKongStateFromStdin(t *testing.T) {
...
-		Name: kong.String("test service"),
-		Host: kong.String("test.com"),
-	},
-		c.Services[0].Service)
+	if assert.NotNil(c) && assert.NotEmpty(c.Services) {
+		assert.Equal(kong.Service{
+			Name: kong.String("test service"),
+			Host: kong.String("test.com"),
+		},
+			c.Services[0].Service)
```
hbagdi, line 109: Why this diff? I'm sure there is a good reason for this but I can't see why this is a conditional assert. Shouldn't these tests be deterministic?

cwurm:
> `ghodss/yaml` is used because it marshals correctly from YAML to JSON.
> It seems there are no tests which catch that but I'm reasonably sure that this is the case and will create a problem if we remove it.

If we can't remove `ghodss/yaml` then this PR might get too complicated for its benefits - `ghodss/yaml` does not support multi-document YAML so I imagine we'd have to vendor and patch it or something.

Before I answer the review comments then: Since this is at the moment just a convenience PR not necessary for normal decK operations, maybe we just shelve it? Wdyt @hbagdi?

hbagdi:
> Since this is at the moment just a convenience PR not necessary for normal decK operations, maybe we just shelve it? Wdyt @hbagdi?

We can shelve it if this is not a priority but please open a Github issue with this feature request.
I'd love to see this feature being added to decK. I expect users to use this feature to manage large configuration.

Alternatively, you could also contribute multi-document support to ghodss/yaml. I'm sure there would be some interest in the Go community for it.

hbagdi:
Tracking this in #241 and closing this out as we don't have a path forward.

PR #210, opened 2020-09-08
fix set default upstream

chenjinxuan:
Kong's Upstream.Healthchecks.Active The default value of. HTTPSVerifyCertificate is true. Therefore, when the test parameter is not set, kubernetes inress controller must be different when comparing current upstream and target upstream. Because HTTPSVerifyCertificate of current upstream is always true
image

### DIFF
--- a/utils/constants.go
+++ b/utils/constants.go
@@ -31,7 +31,7 @@ var (
 		Slots: kong.Int(defaultSlots),
 		Healthchecks: &kong.Healthcheck{
 			Active: &kong.ActiveHealthcheck{
-				Concurrency: kong.Int(defaultConcurrency),
+				Concurrency:            kong.Int(defaultConcurrency),
 				HTTPSVerifyCertificate: kong.Bool(true),
 				Healthy: &kong.Healthy{
 					HTTPStatuses: []int{200, 302},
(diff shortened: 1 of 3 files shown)


utils/constants.go
```diff
@@ -31,7 +31,8 @@ var (
 		Slots: kong.Int(defaultSlots),
 		Healthchecks: &kong.Healthcheck{
 			Active: &kong.ActiveHealthcheck{
-				Concurrency: kong.Int(defaultConcurrency),
+				Concurrency:            kong.Int(defaultConcurrency),
+				HTTPSVerifyCertificate: kong.Bool(true),
```
hbagdi, line 35: The problem here is that if you set the default to  `true`, there is absolutely no way to set it to `false` even if someone want's to do it.

chenjinxuan: If you do not set the parameter httpsverifycertificate, Kong will default to true?
According to my observation, Kong sets httpsverifycertificate = true by default

hbagdi: Yes, the default is true in Kong: https://docs.konghq.com/2.2.x/admin-api/#upstream-object

hbagdi:
Closing as this has gone stale. Please re-open if needed.

Issue #218, opened 2020-10-16
deck requires permissions to /workspaces beyond workspace permissions


mflendrich:
Per @hbagdi,
> decK performs a check to make sure that the workspace exists, if tries to create it.
> The check and creation both require more privileges than the workspace super-admin grant.

As a workspace super-admin, I want to be able to operate on that workspace using deck. However, the current privilege checks disallow that. The scope of this issue is to ensure that workspace rights are enough to preform reset/sync.

```console
$ deck reset --workspace Customers --headers "Kong-Admin-Token:kong" --verbose 10
This will delete all configuration from Kong's database.
> Are you sure? y
GET /workspaces/Customers HTTP/1.1
Host: localhost:8001
User-Agent: Go-http-client/1.1
Accept-Encoding: gzip
HTTP/1.1 403 Forbidden
Content-Length: 116
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: http://localhost:8002
Connection: keep-alive
Content-Type: application/json; charset=utf-8
Date: Fri, 18 Sep 2020 18:40:03 GMT
Server: kong/1.5.0.5-enterprise-edition
Vary: Origin
X-Kong-Admin-Latency: 43
X-Kong-Admin-Request-Id: 54FehCuOSfkVJxJ0TxqgNVVdGLzy4Ljv
{"message":"fel@konghq.com-4071bc34-bea5-48a9-88bd-eddcb726294a, you do not have permissions to read this resource"}
Error: error when getting workspace: 403 Forbidden {"message":"fel@konghq.com-4071bc34-bea5-48a9-88bd-eddcb726294a, you do not have permissions to read this resource"}
$ 
```

hbagdi:
@mflendrich Do you have a solution in mind on how to solve this problem?

mflendrich:
@hbagdi
`deck reset --workspace=X`, before attempting to delete items in the workspace, performs a `GET` on `/workspaces/X`. We can solve the problem by removing this GET and proceeding with deletion straight away. If a workspace does not exist, the first of the delete operations will result in a 404 with a `message` along the lines of `workspace 'X' does not exist`.

The fix would be to:
- patch go-kong to include the message in returned errors,
- remove the GET `/workspaces/X` in the `reset` command.

mflendrich:
Fixes mentioned above sent for review: #219, Kong/go-kong#8.

mflendrich:
[...]

PR #219, opened 2020-10-19
feat(reset): do not Get() workspace when resetting

mflendrich:
🚧 DRAFT because this needs to include Kong/go-kong#8 as a Go module. Otherwise ready for review 🚧

Part of #218 (handles `reset`).

Before this commit, `deck reset --workspace=X` performs a GET
on :8001/workspaces/<name> in order to tell whether the workspace
exists.

This prevents a user with full privileges _within_ the workspace
from resetting it, because a typical "workspace admin" role
(that is, having endpoint=* actions=* within the workspace) does
not permit GETting /workspaces/<name>.

After this commit, `deck reset` no longer tries to `GET` a workspace
before resetting it. Author's intention is that this change does not
modify workspace resetting semantics.

- When the calling user has endpoints=* actions=* on the workspace: no
change in behavior

- When the calling user has partial permissions to endpoints in the
workspace: workspace reset will fail with a partially reset state.
Removing the Get check does not change this behavior.

- When the calling user has no write permissions to any endpoints in the
workspace: no resetting takes place and the workspace is left intact. No
change in behavior here either.

There is one visible change, though:
- When the calling user has broad permissions but the workspace does not
exist: entity listing operations fail with 404 and a message that workspace
was not found)

Per #218.

### DIFF
--- a/cmd/reset.go
+++ b/cmd/reset.go
@@ -69,16 +69,6 @@ By default, this command will ask for a confirmation prompt.`,
 			}
 		}
 		if resetWorkspace != "" {
-			config.Workspace = resetWorkspace
-
-			exists, err := workspaceExists(config)
-			if err != nil {
-				return err
-			}
-			if !exists {
-				return errors.Errorf("workspace '%v' does not exist in Kong", resetWorkspace)
-			}
-
 			workspaces = append(workspaces, resetWorkspace)
 		}
 


mflendrich:
Closing this PR - will modify `workspaceExists()` to use in-workspace privileges instead.

PR #223, opened 2020-10-19
fix(builder) reject creds without required fields

rainest:
If attempting to search for a credential with no ID, reject the credential entirely if its search field is not present.

This addresses the fatal panic in https://github.com/Kong/kubernetes-ingress-controller/issues/532. These resources remain invalid, but now throw the controller into configuration sync failure loop instead of panicking, e.g.

```
time="2020-10-19T19:26:34Z" level=info msg="syncing configuration" component=controller
time="2020-10-19T19:26:34Z" level=error msg="failed to update kong configuration: building state: key-auth credential is invalid: no key" component=controller
time="2020-10-19T19:26:34Z" level=error msg="failed to sync: building state: key-auth credential is invalid: no key" component=sync-queue
```

Ideally, this should print either the Secret or KongConsumer name to help users find the problem resource, but we cannot handle that at the deck level, where we only know the contents of the proposed credential, and cannot really print anything meaningful about it (you only trigger this error if you provide neither the ID nor search field, which are de facto the only identifiers deck has).

The sync failure loop isn't ideal either, but I'm not sure there's any way we can avoid it, as the admission controller does not reject configuration that does reject invalid credentials on creation: the controller only attempts to load them after when processing the KongConsumer, and the linked invalid credential does not make the consumer itself invalid. Adding consumer or credential context to the log is the next best option.

No tests here as this is essentially a KIC-only issue: after accepting this we should add a new KIC card to add log context and tests to confirm the error appears as expected. I've tested this manually using a temporary KIC build that incorporates the deck module at cbc05ec.

### DIFF
--- a/file/builder.go
+++ b/file/builder.go
@@ -268,6 +268,9 @@ func (b *stateBuilder) ingestKeyAuths(creds []kong.KeyAuth) error {
 	for _, cred := range creds {
 		cred := cred
 		if utils.Empty(cred.ID) {
+			if utils.Empty(cred.Key) {
+				return errors.Errorf("key-auth credential is invalid: no key")
+			}
 			existingCred, err := b.currentState.KeyAuths.Get(*cred.Key)
 			if err == state.ErrNotFound {
 				cred.ID = uuid()
@@ -289,6 +292,9 @@ func (b *stateBuilder) ingestBasicAuths(creds []kong.BasicAuth) error {
 	for _, cred := range creds {
 		cred := cred
 		if utils.Empty(cred.ID) {
+			if utils.Empty(cred.Username) {
+				return errors.Errorf("basic-auth credential is invalid: no username")
+			}
 			existingCred, err := b.currentState.BasicAuths.Get(*cred.Username)
 			if err == state.ErrNotFound {
 				cred.ID = uuid()
@@ -310,6 +316,9 @@ func (b *stateBuilder) ingestHMACAuths(creds []kong.HMACAuth) error {
 	for _, cred := range creds {
 		cred := cred
 		if utils.Empty(cred.ID) {
+			if utils.Empty(cred.Username) {
+				return errors.Errorf("hmac-auth credential is invalid: no username")
+			}
 			existingCred, err := b.currentState.HMACAuths.Get(*cred.Username)
 			if err == state.ErrNotFound {
 				cred.ID = uuid()
@@ -331,6 +340,9 @@ func (b *stateBuilder) ingestJWTAuths(creds []kong.JWTAuth) error {
 	for _, cred := range creds {
 		cred := cred
 		if utils.Empty(cred.ID) {
+			if utils.Empty(cred.Key) {
+				return errors.Errorf("jwt-auth credential is invalid: no key")
+			}
 			existingCred, err := b.currentState.JWTAuths.Get(*cred.Key)
 			if err == state.ErrNotFound {
 				cred.ID = uuid()
@@ -352,6 +364,9 @@ func (b *stateBuilder) ingestOauth2Creds(creds []kong.Oauth2Credential) error {
 	for _, cred := range creds {
 		cred := cred
 		if utils.Empty(cred.ID) {
+			if utils.Empty(cred.ClientID) {
+				return errors.Errorf("oauth2 credential is invalid: no client_id")
+			}
 			existingCred, err := b.currentState.Oauth2Creds.Get(*cred.ClientID)
 			if err == state.ErrNotFound {
 				cred.ID = uuid()
@@ -373,6 +388,9 @@ func (b *stateBuilder) ingestACLGroups(creds []kong.ACLGroup) error {
 	for _, cred := range creds {
 		cred := cred
 		if utils.Empty(cred.ID) {
+			if utils.Empty(cred.Group) {
+				return errors.Errorf("acl credential is invalid: no group")
+			}
 			existingCred, err := b.currentState.ACLGroups.Get(
 				*cred.Consumer.ID,
 				*cred.Group)


hbagdi:
This is not the right solution to the problem at hand.

One of the most important design considerations for KIC is to **not** block the reconciliation loops at all cost. A blocked reconciliation loop is extremely problematic for users and maintainers alike. With that in mind, the correct alternate here is to skip ingesting such bad credentials, log them and then continue to sync configuration to Kong.

Also, for the standalone decK setup, we have code in decK which rejects a bad [file-based input](https://github.com/Kong/deck/blob/main/file/codegen/main.go#L91-L99). This is not executed in the code-path that KIC employs and that is intentional.

rainest:
> A blocked reconciliation loop is extremely problematic for users and maintainers alike.

No disagreement there. My rationale for this PR was that while blocked reconciliation is bad, it's maybe better than the current "panic and die, and then panic and die again when K8S respawns the container until the Pod enters CrashLoopBackoff" approach.

IMO the long-term strategic approach to that is what I'd wanted for the old internal Jira report in CLD-603: if the KIC parser encounters encounters a resource that it cannot apply (because the Kong admin API won't accept it), that it should then retry, and if it encounters the same error repeatedly, it marks the resource bad and stops attempting to apply it, and proceeds onward to other resources.

While we attempted to address that in part with the admission controller, it still has issues in practice, namely:
- The admission controller isn't enabled by default, and unless instructed to enable it, users don't until they wind up in a reconciliation failure loop.
- The admission controller cannot catch all issues. I'm unsure of the full scope of those limitations, but it is a demonstrated problem, e.g. as reported in FTI-1896

Addressing that is larger and more complex task however, hence my attempt here to create a smaller fix tailored to this specific problem.

> the correct alternate here is to skip ingesting such bad credentials

Do you mean reject it in the controller? I'd considered that, but was wary of it for the reasons below.

Overall, the controller's [existing credential validation](https://github.com/Kong/kubernetes-ingress-controller/blob/1.0.0/internal/ingress/controller/parser/kongstate/consumer.go#L27-L85) is pretty sparse. For most credentials, it only performs a cursory "is this valid JSON?" check before passing it onward.

The only current exception is [jwt-auth](https://github.com/Kong/kubernetes-ingress-controller/blob/1.0.0/internal/ingress/controller/parser/kongstate/consumer.go#L58-L73), whose validation addresses a similar problem to what we're facing here: the credential is valid according to Kong because it omits a value that Kong populates by default, but it's bad for deck because deck can't rely on Kong's defaults: it isn't aware of them, and can only diff its static input against the exact values it observes from the Kong admin API. That JWT case is a bit different, as it _does_ have a single default (a jwt-auth credential with no algorithm defaults to `algorithm=HS256` and `algorithm=HS256` only), whereas the here we have effectively infinite default value (the default key for key-auth credentials with no key is a random string).

That presents a dilemma: there are some Kong proxy behaviors that are essentially incompatible with declarative configuration, and the only way for our declarative tools to work around those is to enforce restrictions on them that are not present in Kong itself. IMO there are a few approaches to address that:

1. Kong provides some means to inform declarative tools about these fields, i.e. the `/schema` endpoints would add a mechanism to indicate the defaults or whether the value is irrelevant. In the case of an explicit default like the jwt-auth algorithm, Kong would report that the default is `HS256`, and the declarative tools could then use that in their diff logic: a declared jwt-auth with no algorithm diffed against a jwt-auth with `algorithm=HS256` doesn't require an update. In the case of random value defaults, the declarative tool would just ignore the value. From a user perspective, relying on this is probably a bad idea (you leave yourself at the mercy of possible default changes, or possible data loss if you didn't back up an existing random value), but it removes the maintenance burden from the declarative tool.
1. Some part, and only one part, of the declarative toolchain serves as the source of truth for either knowing what the defaults are or requiring that users specify explicit values for them. While having a single part of the toolchain responsible for this reduces the maintenance burden for keeping track of schema changes, I don't think the controller is a good candidate for being that single part: we allow use of deck and go-kong independent of the controller, and the controller leverages them both, suggesting that deck or go-kong would be better candidates for this. As we've seen here, however, there are obstacles to that: in this case, deck doesn't have a mechanism to report back _which_ piece of the declaration has the issue, only that some issue exists.
1. We continue with the current state affairs, adding validation logic piecemeal to individual parts of the toolchain as we need them. This is flexible, but requires more work to maintain both initially (e.g. if there's some problem that affects both the controller and deck, as is the case here, we have to add different mechanisms to each separately) and over time (if Kong's behavior changes, we have to update both validation implementations).

I like option 2 most, but we're currently effectively on 3. I do think the maintenance burden for either would be reduced by having some means of generating such validations automatically, somehow parsing Kong schemas to infer which fields need extra validation for declarative config (e.g. for [key-auth](https://github.com/Kong/kong/blob/2.2.0/kong/plugins/key-auth/daos.lua#L17) we have `auto = true`) and for [jwt-auth](https://github.com/Kong/kong/blob/2.2.0/kong/plugins/jwt/daos.lua#L31) we have `default = "HS256"`). The wrench in the works there is custom plugins: such an approach relies on our knowing what schemas matter in advance, and that's not something we can rely on in practice. The API-based approach in option 1 is one way to address that; we could alternately create some sort of declarative metadata and rely on plugin authors to fill it out.

hbagdi:
So it seems like DRYing out code and maintenance burden are the primary concerns.
I do agree that validation is in two places here and it should be only in one place. It is unfortunate timing of development where refactoring and DRYing out was not an option. This can be done at an upfront cost that I would recommend against at this point. DRY in this area hasn't been important because these pieces of code are touched less frequently, sometimes once a year or so. The low ROI on this is what makes me hesitant on pursuing this further.

Option 1 of dynamic defaults is something I've considered time and again but it has never been a priority. We do use schemas endpoint to perform defaulting logic for plugins: https://github.com/Kong/kubernetes-ingress-controller/blob/main/internal/ingress/controller/plugin_schema_helper.go so it is doable but doesn't provide enough ROI to get to it.

`go-kong` seems like a good place to implement this and it can certainly be done in a way that decK and KIC both re-use the same implementations. The way they reflect back errors to the end user can be different. This requires less refactoring in KIC but a lot more in decK.

While putting more band-aid on something that is already bleeding after putting on a band-aid 2 times already sounds like a bad idea, I'd still do it because the band-aid is quick and easy. We do want to fix it properly but now is not the time.

hbagdi:
@rainest Is this still required or can it be closed out?

PR #225, opened 2020-10-20
feat: reimplement workspaceExists() without the /workspaces endpoint

### LINKED ISSUE #218: deck requires permissions to /workspaces beyond workspace permissions

mflendrich:
fixes #218 
replaces #219 

This PR:
- cleans up deck's usage of non-workspaced ("root") and workspaced clients and configs,
- makes workspaceExists() GET `/<workspaceName>/routes/` instead of `/workspaces/<workspaceName>` because users of deck are much more likely to have permissions to access the latter endpoint, rather than the former (e.g. in a case where the calling user is a workspace admin and has permissions only within the namespace)

The change seems justified because it appears impractical to use deck on a workspace without read permissions to the `/<workspaceName>/routes/` endpoint.

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -37,6 +37,11 @@ func SetStopCh(stopCh chan struct{}) {
 
 // workspaceExists checks if workspace exists in Kong.
 func workspaceExists(config utils.KongClientConfig) (bool, error) {
+	if config.Workspace == "" {
+		// default workspace always exists
+		return true, nil
+	}
+
 	if config.SkipWorkspaceCrud {
 		// if RBAC user, skip check
 		return true, nil
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -75,32 +75,33 @@ func syncMain(filenames []string, dry bool, parallelism, delay int, workspace st
 		return err
 	}
 
-	rootClient, err := utils.GetKongClient(config)
+	rootClient, err := utils.GetKongClient(rootConfig)
 	if err != nil {
 		return err
 	}
 
+	var wsConfig utils.KongClientConfig
 	// prepare to read the current state from Kong
 	if workspace != targetContent.Workspace && workspace != "" {
 		print.DeletePrintf("Warning: Workspace '%v' specified via --workspace flag is "+
 			"different from workspace '%v' found in state file(s).\n", workspace, targetContent.Workspace)
-		config.Workspace = workspace
+		wsConfig = rootConfig.ForWorkspace(workspace)
 	} else {
-		config.Workspace = targetContent.Workspace
+		wsConfig = rootConfig.ForWorkspace(targetContent.Workspace)
 	}
 
 	// load Kong version after workspace
-	kongVersion, err := kongVersion(config)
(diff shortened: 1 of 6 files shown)


cmd/common.go
```diff
@@ -36,35 +37,25 @@ func SetStopCh(stopCh chan struct{}) {
 
 // workspaceExists checks if workspace exists in Kong.
 func workspaceExists(config utils.KongClientConfig) (bool, error) {
-
-	workspace := config.Workspace
-	if workspace == "" {
-		// default workspace always exists
-		return true, nil
-	}
```
hbagdi, line 44: Please do not remove this optimization. This saves a round-trip to Kong for all Kong OSS installations.

utils/types.go
```diff
@@ -79,6 +79,13 @@ type KongClientConfig struct {
 	Headers []string
 }
 
+// ForWorkspace returns a copy of KongClientConfig that produces a KongClient for the workspace specified by argument.
+func (kc *KongClientConfig) ForWorkspace(name string) KongClientConfig {
```
hbagdi, line 83: Nice one. I like this.

hbagdi (approved):
LGTM apart from one comment. Please merge after making the change.

A general node: This part of this codebase has become complex and fragile over time and we should test it. This has never been prioritized as there haven't been many quality issues in this area (yet).

cmd/common.go
mflendrich: 942b42b

utils/types.go
mflendrich: Realized that I missed the `ping` operation. Fixed in e61a5b7.

PR #226, opened 2020-10-20
feat(sync) Retry sync operations

cwurm:
Admin API calls to Kong can fail for a number of reasons, some permanent (e.g. configuration syntax error). Others are transient, e.g.:
1. Kong - or rather the underlying configuration database - gets overloaded with too many requests and will start returning `500` errors. I found this almost certain to happen with very large configurations of thousands or tens of thousands of entities.
2. Plugins can have race conditions, e.g. on setup when several plugin instances are created at once by decK. I have seen this just today.

When these transient errors happen a decK run will abort since the current behavior is for each decK worker to stop when it encounters any error whatsoever. Having to restart decK for transient reasons either manually or automatically is not ideal - esp. when it's a large configuration it will have to re-query all configuration, and potentially re-sync a lot of content (e.g. when using a decK config without default values specified).

This PR adds a `--retry` and `--retry-delay` parameter to the `deck sync` command. Default values are `1` retry and no delay. This should solve the most obvious transient problems by default without adding a lot of execution time (adding a delay can, since then each worker might have to wait for that time before shutting down, which could be a long time if there are a lot of errors).

To test this, what I've seen work well is to run deck with a very high number of workers, e.g. `deck sync -s kong.yaml --parallelism 10000 --retries 0`. With any reasonably large Kong configuration, this will likely fail (it does for me with about 120 entities). With `deck sync -s kong.yaml --parallelism 10000 --retries 1 --retry-delay 2` (retry delay set to 2s here) it should work.

### DIFF
--- a/cmd/sync.go
+++ b/cmd/sync.go
@@ -9,8 +9,6 @@ var (
 	syncCmdKongStateFile []string
 	syncCmdParallelism   int
 	syncCmdDBUpdateDelay int
-	syncCmdRetries       int
-	syncCmdRetryDelay    int
 	syncWorkspace        string
 )
 
@@ -23,8 +21,7 @@ var syncCmd = &cobra.Command{
 to get Kong's state in sync with the input state.`,
 	Args: validateNoArgs,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return syncMain(syncCmdKongStateFile, false, syncCmdParallelism,
-			syncCmdDBUpdateDelay, syncCmdRetries, syncCmdRetryDelay, syncWorkspace)
+		return syncMain(syncCmdKongStateFile, false, syncCmdParallelism, syncCmdDBUpdateDelay, syncWorkspace)
 	},
 	PreRunE: func(cmd *cobra.Command, args []string) error {
 		if len(syncCmdKongStateFile) == 0 {
@@ -58,8 +55,4 @@ func init() {
 		0, "aritificial delay in seconds that is injected between insert operations \n"+
 			"for related entities (usually for cassandra deployments).\n"+
 			"See 'db_update_propagation' in kong.conf.")
-	syncCmd.Flags().IntVar(&syncCmdRetries, "retries",
-		1, "Number of times decK will retry an insert or update operation")
-	syncCmd.Flags().IntVar(&syncCmdRetryDelay, "retry-delay",
-		0, "Number of seconds decK will wait before retrying an insert or update operation")
 }
--- a/cmd/sync.go
+++ b/cmd/sync.go
@@ -9,6 +9,8 @@ var (
 	syncCmdKongStateFile []string
 	syncCmdParallelism   int
 	syncCmdDBUpdateDelay int
+	syncCmdRetries       int
+	syncCmdRetryDelay    int
 	syncWorkspace        string
 )
 
@@ -21,7 +23,8 @@ var syncCmd = &cobra.Command{
 to get Kong's state in sync with the input state.`,
 	Args: validateNoArgs,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return syncMain(syncCmdKongStateFile, false, syncCmdParallelism, syncCmdDBUpdateDelay, syncWorkspace)
+		return syncMain(syncCmdKongStateFile, false, syncCmdParallelism,
+			syncCmdDBUpdateDelay, syncCmdRetries, syncCmdRetryDelay, syncWorkspace)
 	},
 	PreRunE: func(cmd *cobra.Command, args []string) error {
 		if len(syncCmdKongStateFile) == 0 {
@@ -55,4 +58,8 @@ func init() {
 		0, "aritificial delay in seconds that is injected between insert operations \n"+
 			"for related entities (usually for cassandra deployments).\n"+
 			"See 'db_update_propagation' in kong.conf.")
+	syncCmd.Flags().IntVar(&syncCmdRetries, "retries",
+		1, "Number of times decK will retry an insert or update operation")
+	syncCmd.Flags().IntVar(&syncCmdRetryDelay, "retry-delay",
+		0, "Number of seconds decK will wait before retrying an insert or update operation")
 }
--- a/diff/diff.go
+++ b/diff/diff.go
@@ -5,6 +5,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/kong/deck/crud"
 	"github.com/kong/deck/state"
 	"github.com/pkg/errors"
@@ -292,9 +293,7 @@ func (sc *Syncer) wait() {
 }
 
 // Run starts a diff and invokes d for every diff.
-func (sc *Syncer) Run(done <-chan struct{}, parallelism int, retries int,
-	retryDelay int, d Do) []error {
-
+func (sc *Syncer) Run(done <-chan struct{}, parallelism int, d Do) []error {
 	if parallelism < 1 {
 		return append([]error{}, errors.New("parallelism can not be negative"))
 	}
@@ -311,7 +310,10 @@ func (sc *Syncer) Run(done <-chan struct{}, parallelism int, retries int,
 	wg.Add(parallelism)
 	for i := 0; i < parallelism; i++ {
 		go func() {
-			sc.eventLoop(retries, retryDelay, d)
+			err := sc.eventLoop(d)
+			if err != nil {
+				sc.errChan <- err
+			}
 			wg.Done()
 		}()
 	}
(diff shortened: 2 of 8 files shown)


cmd/sync.go
```diff
@@ -55,4 +58,8 @@ func init() {
 		0, "aritificial delay in seconds that is injected between insert operations \n"+
 			"for related entities (usually for cassandra deployments).\n"+
 			"See 'db_update_propagation' in kong.conf.")
+	syncCmd.Flags().IntVar(&syncCmdRetries, "retries",
+		1, "Number of times decK will retry an insert or update operation")
+	syncCmd.Flags().IntVar(&syncCmdRetryDelay, "retry-delay",
+		0, "Number of seconds decK will wait before retrying an insert or update operation")
```
hbagdi, line 64: Please do not make retry delay configurable (yet).
Why?
1. Keeping configuration parameters to minimum is important for a good user experience.
2. Configuring delay in such cases seldom adds value. The best practice in this area is to perform an exponential backoff delay which saves the day in a lot of such failure scenarios. Thundering herd is a real problem especially when default parallelism is 10.

Hardcode a delay scheme using a good exponential backoff algorithm library.

cmd/sync.go
```diff
@@ -55,4 +58,8 @@ func init() {
 		0, "aritificial delay in seconds that is injected between insert operations \n"+
 			"for related entities (usually for cassandra deployments).\n"+
 			"See 'db_update_propagation' in kong.conf.")
+	syncCmd.Flags().IntVar(&syncCmdRetries, "retries",
+		1, "Number of times decK will retry an insert or update operation")
```
hbagdi, line 62: I'd advise even against making retries configurable in the beginning. I do see value here but I'd like to see users asking for it before we add it. Having an internally hardcoded value of `5` retries should do it for most users.

hbagdi (changes requested):
Apologies for the delay in review here.

cwurm:
@hbagdi Thanks for the review, I finally had time to look at this again and rewrote the implementation to use the `cenkalti/backoff` exponential backoff library.

The tricky part is that to have a successful default retry policy the potential maximum backoff is very long. In my tests, to be able to load several thousand Kong entities with `--parallelism 100` (trying to approximate a very large, complex and frequently automatically changed Kong deployment) a backoff of up to 10 seconds was required.

This is annoying when the errors are not retryable, e.g. wrong plugin configuration, access not authorized, etc. If all classes of errors are retried, decK will run through all retries even when the error is permanent and take a long time (15-20 seconds) doing nothing before finally exiting with the error.

So I needed to add some logic for which errors are retryable. In my tests, Kong will usually return a `500 Internal Server Error` in those cases, e.g. when its database connection gets overloaded. Bad plugin configuration is a `400 Bad Request`, missing access rights is `401 Unauthorized`, etc.

Unfortunately, `go-kong` does not currently expose the status code in an easily consumable way. We could change it to do that, what do you think? For now I made it work by checking the error message returned from go-kong.

cmd/sync.go
cwurm: Makes sense, I changed to an exponential backoff.

cmd/sync.go
cwurm: Makes sense, I've experimented with some values for the updated implementation with exponential backoff and came up with 3 retries after around 1 second, 3 seconds, and 9 seconds (randomized to avoid the thundering herd). Let me know what you think.

diff/diff.go
```diff
@@ -363,15 +362,23 @@ func (sc *Syncer) Run(done <-chan struct{}, parallelism int, d Do) []error {
...
+		var err error
+		for retries >= 0 {
+			err = sc.handleEvent(d, event)
+			if err == nil {
+				sc.eventCompleted()
+				break
+			}
+			retries--
+			time.Sleep(time.Duration(retryDelay) * time.Second)
+		}
```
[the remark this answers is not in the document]
cwurm: Let me know what you think of the updated implementation. Still in `eventLoop()`, I guess could move it to `handleEvent()` though I think they'd look pretty much the same.

solver/solver.go
```diff
@@ -59,6 +62,12 @@ func Solve(doneCh chan struct{}, syncer *diff.Syncer,
 			// fire the request to Kong
 			result, err = r.Do(e.Kind, e.Op, e)
 			if err != nil {
+				// Make only 500 errors retryable - this is what Kong returns when e.g. the
+				// database is temporarily overloaded.
+				// TODO: Replace with better check when/if go-kong exposes status code
+				if !strings.HasPrefix(err.Error(), "500") {
```
hbagdi, line 68: Can we make this a little more robust by checking against a regex instead to account for 501/502/503 or rather 5xx?

I've also opened up https://github.com/Kong/go-kong/issues/10 to make this better in future.

diff/diff.go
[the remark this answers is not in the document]
hbagdi: I think it is best to move it to handleEvent as it keeps the code testable and keeps the responsibilities of functions separate.

diff/diff.go
```diff
@@ -365,7 +366,26 @@ type Do func(a Event) (crud.Arg, error)
...
+
+		// For various reasons, Kong can temporarily fail to process
+		// a valid request (e.g. when the database is under heavy load).
+		// We retry each request up to 3 times on failure, after around
+		// 1 second, 3 seconds, and 9 seconds (randomized exponential backoff).
+		exponentialBackoff := backoff.NewExponentialBackOff()
+		exponentialBackoff.InitialInterval = 1 * time.Second
+		exponentialBackoff.Multiplier = 3
+		backoffObject := backoff.WithMaxRetries(exponentialBackoff, 4)
```
hbagdi, line 383: Please declare backoffObject as a global variable and initialize it in an `init()`. That keeps this function (or rather handleEvent fn after refactoring) cleaner. Readability and maintainability in this part of the code is more important than usual because it is easy to trip over.

diff/diff.go
```diff
@@ -365,7 +366,26 @@ type Do func(a Event) (crud.Arg, error)
...
+
+		err := backoff.Retry(func() error {
+			return sc.handleEvent(d, event)
```
hbagdi, line 386: Can we please emit a log line when an event is being retried? That would help quite a bit with debugging.

hbagdi (changes requested):

solver/solver.go
cwurm: I'm not sure if anything other than 500 can be a transient error? The trouble with retrying a permanent error would be that the decK run will take a long time while it is retrying every request the maximum number of times. So far I've only seen 500 error be transient, e.g. when the database was overloaded with too many requests. So would it not make sense to be conservative here?

diff/diff.go
cwurm: I agree. However, I don't see any debug logging facility in decK? We have the print package with things like CreatePrintln/DeletePrintln, but those will always print. I imagine that ideally this kind of debug information could be turned on or off, e.g. with a verbose or debug flag.

solver/solver.go
hbagdi: In practice, a lot of times there is another hop or proxy between decK and the Admin endpoint of Kong. That intermediate layer can send a 502 or 503 intermittently. This is certainly a hypothetical scenario and we can defer the change.

diff/diff.go
hbagdi: There is a verbose flag in decK. You are right that the machinery around debug logging is lacking and there is not much that can be done easily. Tracking in #248 and #249.

diff/diff.go
cwurm: done

solver/solver.go
cwurm: Yeah, I see how there can be other transient errors. I like the idea of keeping the change small and restricted to just the specific case where we know decK does not do what a user would expect (load a large config with 1000s/10000s of Kong entities). We can always easily expand the reach if it comes up?

diff/diff.go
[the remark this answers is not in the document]
cwurm: Makes sense, I moved it.

hbagdi (approved):

hbagdi:
Holding this merge until #250 gets merged in and released. That keeps the release process simpler.

hbagdi:
@mflendrich Can I request you review this patch?

solver/solver.go
mflendrich, line 68: - Instead of parsing the error message, please update go-kong to `v0.14.0` and extract `Code()` from the [`APIError` type](https://github.com/Kong/go-kong/blob/f71c83b51f21bf129b63a60aa88447203d6bfb8b/kong/error.go#L8).
- Then, instead of literal `500`, use `http.StatusInternalServerError`

cwurm: Glad to see the addition to go-kong, done.

diff/diff.go
```diff
@@ -5,15 +5,28 @@ import (
...
 
+func init() {
+	// For various reasons, Kong can temporarily fail to process
+	// a valid request (e.g. when the database is under heavy load).
+	// We retry each request up to 3 times on failure, after around
+	// 1 second, 3 seconds, and 9 seconds (randomized exponential backoff).
+	exponentialBackoff := backoff.NewExponentialBackOff()
+	exponentialBackoff.InitialInterval = 1 * time.Second
+	exponentialBackoff.Multiplier = 3
+	backoffObject = backoff.WithMaxRetries(exponentialBackoff, 4)
```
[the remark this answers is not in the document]
cwurm: So this is just implementing the review from @hbagdi here: https://github.com/Kong/deck/pull/226#discussion_r535421432

Happy to do this any way that works, just let me know.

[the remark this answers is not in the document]
mflendrich: Oh, apologies about contradicting review comments :joy:. I'll get this sorted out.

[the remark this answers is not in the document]
mflendrich: Sorry about changing expectations on our side. I have sorted this out with @hbagdi and we'd like to go for a helper function.

I meant to save you a round-trip by pushing the change to your PR branch but it seems that GH doesn't allow me to push. Therefore, please cherry-pick 35bccc0a3bfe5d3583888db75133ea88f40c8a74 or make an equivalent change yourself.

Thanks!

diff/diff.go
```diff
@@ -375,16 +397,33 @@ func (sc *Syncer) eventLoop(d Do) error {
...
-	return nil
+	err := backoff.Retry(func() error {
+		res, err := d(event)
+		if err != nil {
+			err = errors.Wrapf(err, "while processing event")
+
+			var kongAPIError *kong.APIError
+			if errors.As(err, &kongAPIError) &&
+				kongAPIError.Code() == http.StatusInternalServerError {
+
```
mflendrich, line 408:
```suggestion

```

mflendrich (commented):
Final nit (+ the backoff helper function comment), otherwise LGTM. Thanks for your contribution!

cwurm:
Thanks @mflendrich, helper function implemented!

mflendrich (approved):

cal5barton:
Is there any reason why the flags are not listed in the general documentation?

rainest:
As best I can tell, per https://github.com/Kong/deck/pull/226#discussion_r513654575 we actually removed the flags from the final version, but forgot to omit them from the commit message. The actual functionality just always retries with a set count and delay configuration.

Asked around to see if someone else with more history on this feature knows more, but that's what looks to be the case based on what I can glean from the PR discussion.

cal5barton:
Right now I'm experiencing 502 errors ( #526 ) and was hoping that this retry functionality would solve something like that.

Issue #144, opened 2020-04-21
Plugin configuration containing arrays not syncing


mattburton:
We have a custom plugin where the configuration is an array - the JSON looks like this:
```json
{
  "headers": [
    {
      "name": "X-Foo",
      "rules": [
        {
          "value": "0",
          "upstream": "foo.com"
        },
        {
          "value": "1",
          "upstream": "bar.com"
        }
      ]
    }
  ]
}
```
cURL:
```shell
curl -i -X POST \
  --url http://localhost:8001/routes/some-service/plugins \
  --header 'content-type: application/json' \
  --data '{"name": "foo-plugin", "config": {"headers":[{"name":"X-Foo","rules":[{"value":"0","upstream":"foo.com"},{"value":"1","upstream":"bar.com"}]}]}}'
```
`deck` dumps the plugin entry like so:
```yaml
- name: foo-plugin
  config:
    headers:
    - name: X-Foo
      rules:
      - upstream: foo.com
        value: "0"
      - upstream: bar.com
        value: "1"
  enabled: true
  protocols:
  - grpc
  - grpcs
  - http
  - https
```
That appears to be a valid YAML array, but when performing a `deck sync` it seems that `deck` is omitting the `headers` array entirely when sending it:
```shell
while processing event: {Update} failed: 400 Bad Request {"message":"schema violation (config.headers: required field missing)","name":"schema violation","fields":{"config":{"headers":"required field missing"}},"code":2}
```
Other than that `deck` has been working wonderfully - thanks for such a great tool!

hbagdi:
That's unusual.
Can you do `deck sync --verbose 2` to see what is the HTTP reqeust that decK does?

mattburton:
@hbagdi sure thing - here's the relevant raw output - seems to be omitting the config parameter entirely:
```
updating plugin foo-plugin for service 6f8508f5-0874-409c-9505-5698e4cd7ae7  {
-  "config": {
-    "headers": [
-      {
-        "name": "X-Foo",
-        "rules": [
-          {
-            "upstream": "foo.com",
-            "value": "0"
-          },
-          {
-            "upstream": "bar.com",
-            "value": "1"
-          },
-          {
-            "upstream": "baz.com",
-            "value": "*"
-          }
-        ]
-      }
-    ]
-  },
-  "created_at": 1.587487361e+09,
   "enabled": true,
   "id": "331a75cf-5fad-47b2-9b62-6df041ee2a72",
   "name": "foo-plugin",
[...]

PR #229, opened 2020-11-09
fix(file) convert yaml to JSON

hbagdi:
`go-yaml` marshals into `map[interface{}]interface{}`.
This change ensures that `map[interface{}]interface{}` is converted into
`map[string]interface{}` RECURSIVELY.

~This patch needs tests before it can be merged.~ Added a test.

See #144

### DIFF
--- a/file/readfile.go
+++ b/file/readfile.go
@@ -8,9 +8,9 @@ import (
 	"path/filepath"
 	"regexp"
 
+	ghodssyaml "github.com/ghodss/yaml"
 	"github.com/imdario/mergo"
 	"github.com/pkg/errors"
-	yaml "gopkg.in/yaml.v2"
 )
 
 // getContent reads all the YAML and JSON files in the directory or the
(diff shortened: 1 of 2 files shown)


file/readfile.go
```diff
@@ -91,13 +91,19 @@ func readContent(reader io.Reader) (*Content, error) {
...
 	if err != nil {
 		return nil, err
 	}
 	return &content, nil
 }
 
+// yamlUnmarshal is a wrapper around yaml.Unmarshal to ensure that the right
+// yaml package is in use. The verification for this is done using a test.
+func yamlUnmarshal(bytes []byte, v interface{}) error {
+	return yaml.Unmarshal(bytes, v)
```
mflendrich, line 104: optional suggestion: import by a nonstardard name, e.g. `ghodssyaml`, to make it harder for goimports to tamper with this import

file/readfile.go
```diff
@@ -91,13 +91,19 @@ func readContent(reader io.Reader) (*Content, error) {
...
-	err = yaml.Unmarshal(bytes, &content)
+	err = yamlUnmarshal(bytes, &content)
 	if err != nil {
 		return nil, err
 	}
 	return &content, nil
 }
 
+// yamlUnmarshal is a wrapper around yaml.Unmarshal to ensure that the right
+// yaml package is in use. The verification for this is done using a test.
```
mflendrich, line 102: please explain (to the future reader) what behavior is desired (and what is undesired) from this function.

mflendrich (approved):
LGTM but see comments

rainest (approved):
Tested against my local environment and confirmed that the example provided by #144 now successfully restores/doesn't delete itself on subsequent sync.

Explanation covers the high level reason for the YAML parser not creating a map that the JSON parser can then unmarshal successfully [later](https://github.com/Kong/go-kong/blob/v0.13.0/kong/types.go#L180-L187).

PR #240, opened 2020-12-03
fix(types) ignore tag order for object comparisons

### LINKED ISSUE #239: Ordered tag comparisons result in spurious diff reports
Originally reported internally in Jira ticket FTI-2097. tl;dr deck thinks tag order matters and tries to act on it. Tag order doesn't actually matter, and deck shouldn't do that.

Assume you have a service whose tags are `[foo,bar]`. When dumping configuration using `--select-tag foo`, deck stores the resulting dump with something like the following:

```
_format_version: "1.1"
_info:
  select_tags:
  - foo
services:
- host: example.com
  name: example
  ...
  tags:
  - bar
```
Although the selected tag is present on the object, it isn't shown directly in the object itself, only in the `_info` block. When syncing, tag information is rebuilt, and tags are added to internal Go structures in some specific order. This order does not necessarily match the order reported by the admin API on Cassandra-backed nodes for an unknown reason (I did not investigate to see exactly how deck and Cassandra build the tag set to determine why this happens, but observed that it did happen with the provided customer data).

Because deck represents the tag set internally with [an ordered slice](https://github.com/Kong/go-kong/blob/v0.13.0/kong/types.go#L22), this order change can result in a reported difference when we [compare the target object (from the dump YAML) to the current object (from the admin API)](https://github.com/Kong/deck/blob/v1.2.3/state/types.go#L109). This difference in the Go structures does not correspond to an actual difference, however: the tags associated with a Kong object are sets, and their order is only an artifact of the YAML or JSON arrays we use to display them. 

Although the apparent difference has no logical impact on the object, deck perceives one, and will attempt to update or create it (depending on whether or not you ignore IDs, I think). This either results in unnecessary requests or a failure (attempting to create a new object with the "correct" tag order will usually result in a uniqueness conflict).

Although the example code here is for services, this can affect any taggable entity (i.e. most entities).

rainest:
Ignore tag order for object comparisons by sorting the string slices that contain them before checking object equality. This avoids issues with spurious diffs resulting in unnecessary changes when the tag order constructed by deck from dump YAML and the tag order reported by admin API JSON differ.

Because Kong object tags are logical string sets, their order does not matter: it is only a perceived difference because JSON and YAML lack set types and represent them with ordered arrays.

Fix #239 and a minor typo ("Copt" to "Copy" in one of the functions).

**Implementation notes/questions:**

We have `Equal()` and `EqualWithOpts()` functions for each object. The former is just `return reflect.DeepEqual(thing1, thing2)`, is not modified by this PR, and would still report the objects not equal as such. The latter copies the objects first and then nulls out their IDs and/or created and updated timestamps based on the options provided before performing a comparison. This PR does modify `EqualWithOpts()` to ignore tag order, though it doesn't introduce a new option: IDs and timestamp differences do actually matter for objects, but tag order never does, so it doesn't seem like it makes sense to add an option.

I'm curious about the history behind this, since as far as I know deck never cares about timestamps (there's no CLI options to toggle that option, and dumps don't include timestamp data, but this doesn't mean that deck repeatedly updates objects because the `nil` timestamps in dumps do not match the not-`nil` timestamps) and only sort of cares about IDs (`dump` has an option to include them or omit them, but `sync` doesn't have an option to ignore them).

- Do we use `Equal()` in practice, and if so, how/why? Do we need to modify it to also create copies and sort their tags?
- Ditto for `EqualWithOpts()`: do we ever use that without ignoring IDs and timestamps?

Unrelated to that, types_test.go does not currently include tags on any of its test objects. We can cover that pretty easily by adding differently-ordered Tags to objects, e.g. after https://github.com/Kong/deck/blob/v1.2.3/state/types_test.go#L51 and confirming that `EqualWithOpts()` is still true, but I wanted to get a pass on the proposed implementation here before writing those out for every object (Golang generics, you cannot arrive soon enough).

### DIFF
--- a/state/types.go
+++ b/state/types.go
@@ -2,6 +2,7 @@ package state
 
 import (
 	"reflect"
+	"sort"
 
 	"github.com/kong/go-kong/kong"
 )
@@ -95,6 +96,11 @@ func (s1 *Service) EqualWithOpts(s2 *Service,
 	s1Copy := s1.Service.DeepCopy()
 	s2Copy := s2.Service.DeepCopy()
 
+	// Cassandra can sometimes mess up tag order, but tag order doesn't actually matter: tags are sets
+	// even though we represent them with slices. Sort before comparison to avoid spurious diff detection.
+	sort.Slice(s1Copy.Tags, func(i, j int) bool { return *(s1Copy.Tags[i]) < *(s1Copy.Tags[j]) })
+	sort.Slice(s2Copy.Tags, func(i, j int) bool { return *(s2Copy.Tags[i]) < *(s2Copy.Tags[j]) })
+
 	if ignoreID {
 		s1Copy.ID = nil
 		s2Copy.ID = nil
@@ -143,6 +149,9 @@ func (r1 *Route) EqualWithOpts(r2 *Route, ignoreID,
 	ignoreTS, ignoreForeign bool) bool {
 	r1Copy := r1.Route.DeepCopy()
 	r2Copy := r2.Route.DeepCopy()
+
+	sort.Slice(r1Copy.Tags, func(i, j int) bool { return *(r1Copy.Tags[i]) < *(r1Copy.Tags[j]) })
+	sort.Slice(r2Copy.Tags, func(i, j int) bool { return *(r2Copy.Tags[i]) < *(r2Copy.Tags[j]) })
 
 	if ignoreID {
 		r1Copy.ID = nil
@@ -195,6 +204,9 @@ func (u1 *Upstream) EqualWithOpts(u2 *Upstream,
 	ignoreID bool, ignoreTS bool) bool {
 	u1Copy := u1.Upstream.DeepCopy()
 	u2Copy := u2.Upstream.DeepCopy()
+
+	sort.Slice(u1Copy.Tags, func(i, j int) bool { return *(u1Copy.Tags[i]) < *(u1Copy.Tags[j]) })
+	sort.Slice(u2Copy.Tags, func(i, j int) bool { return *(u2Copy.Tags[i]) < *(u2Copy.Tags[j]) })
 
 	if ignoreID {
 		u1Copy.ID = nil
@@ -251,6 +263,9 @@ func (t1 *Target) EqualWithOpts(t2 *Target, ignoreID,
 	t1Copy := t1.Target.DeepCopy()
 	t2Copy := t2.Target.DeepCopy()
 
+	sort.Slice(t1Copy.Tags, func(i, j int) bool { return *(t1Copy.Tags[i]) < *(t1Copy.Tags[j]) })
+	sort.Slice(t2Copy.Tags, func(i, j int) bool { return *(t2Copy.Tags[i]) < *(t2Copy.Tags[j]) })
+
 	if ignoreID {
 		t1Copy.ID = nil
 		t2Copy.ID = nil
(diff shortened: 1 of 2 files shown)


hbagdi:
> I'm curious about the history behind this, since as far as I know deck never cares about timestamps (there's no CLI options to toggle that option, and dumps don't include timestamp data, but this doesn't mean that deck repeatedly updates objects because the nil timestamps in dumps do not match the not-nil timestamps) and only sort of cares about IDs (dump has an option to include them or omit them, but sync doesn't have an option to ignore them).

There is development history behind it. In an early PoC of decK, the code looked a lot different than it is today and there was a DAG built to solve the diffing problem. It turned out to be over-engineered. The code simplification kept major parts of the codebase, including this. There are no parts of code that care about timestamps in diffing.
ID is a little different. Around decK 0.5.0 or 0.6.0, decK again underwent a major re-write to simplify code and ensure 100% correctness without an elaborate test suite. Before this re-write, some portions of code ignored ID for comparison and some didn't. I don't recall why that was the case, git spelunking might help here.

> Do we use Equal() in practice, and if so, how/why? Do we need to modify it to also create copies and sort their tags?

We don't but these functions are now diverging. Order of tags should be ignored no matter what. I'd recommend the following refactor as part of this change:

```go
fn Equal() {
  return EqualWithOpt(with, default, arguments)
}
```

> Ditto for EqualWithOpts(): do we ever use that without ignoring IDs and timestamps?

A quick grep says no.

hbagdi:
Consider the PR approved once you add in tests and refactor the Equal fns.

rainest:
Double-checking on the refactor, but I think we're good to go here. Grepping shows that we actually do have ID checks _enabled_ in current calls, though presumably most users aren't using `--with-id` when dumping and/or aren't changing object IDs between runs, so that doesn't really matter in practical usage.

The complete signatures follow a `func (c1 *Object) EqualWithOpts(c2 *Object, ignoreID bool, ignoreTS bool, ...) bool` pattern, e.g. [for aclGroup](https://github.com/Kong/deck/blob/v1.2.3/diff/aclGroup.go#L86), and across the codebase:

```
$ grep -Ir EqualWithOpts | egrep -ce "\(\w*, false" 
32
$ grep -Ir EqualWithOpts | egrep -ce "\(\w*, true"  
0
```
Have made that the default in the `Equal()` refactor as such. Would we want to flip that for any reason?

Not sure if there's some more idiomatic way to handle the tag test without creating the test `Tags` slice in each test function--doesn't appear to be any way to create constant `*string` slices, even if you make them fixed-sized arrays, so the generator function works around that.

state/types.go
```diff
@@ -84,7 +85,7 @@ func (s1 *Service) Console() string {
 
 // Equal returns true if s1 and s2 are equal.
 func (s1 *Service) Equal(s2 *Service) bool {
-	return reflect.DeepEqual(s1.Service, s2.Service)
+	return s1.EqualWithOpts(s2, false, true)
```
hbagdi, line 88: This should be `(s2, false, false)`. The contract for this function is that it compares the two objects without ignoring any fields. This comment also applies to all other `Equal()`.

rainest: Gotcha--misunderstood and thought we wanted `Equal()` to match our use of `EqualWithOpts()` in practice. Fixed now; forced since the commit message was inaccurate otherwise.

hbagdi (approved):
I had a thought that there might be a way to DRY this out a little but I ignored it because I expect this code to not be touched enough.

PR #242, opened 2020-12-06
Stop tracking your users

marcelometal:


hbagdi:
Hey @marcelometal,

Thanks for opening the PR. I completely understand your concern here.
We are not tracking users using our anonymous analytics. It is **strictly** meant to track rate of adoption of decK and it's various versions. We intentionally do not include any sort of identifying information in the data that is sent back.

If you would like, you can disable the anonymized reports by setting the environment variable `DECK_ANALYTICS=off`.

marcelometal:
Hi @hbagdi ,

Since you don't intend to remove this part of the code, could you please add a note in README.md to inform all users about deck analytics and how to disabled it?

hbagdi:
We call it out in our documentation: https://docs.konghq.com/deck/commands/#analytics.
README.md is meant to provide an introduction to the project and I'm a little hesitant to talk about this detail in there. Is that acceptable?

marcelometal:
Sorry, but IMHO no, that's not enough...

For me, transparency is one of the most compelling benefits of open source software. So, there are no surprises and no hidden functions...

For example, Debian Popularity Contest ([popcon](https://popcon.debian.org/
)). The default value is "no" and it's a personal choice.

[Features section](https://github.com/Kong/deck#features
) in README has practically the same content of [Commands Section](https://docs.konghq.com/deck/commands/#analytics) in Kong`s website. So, please add a small paragraph about deck analytics in README. It will increases credibility too.

image

hbagdi:
> So, there are no surprises and no hidden functions...

I understand where you are coming from here but please note that we have absolutely no intentions of hiding any feature. 

> For me, transparency is one of the most compelling benefits of open source software.

I completely agree. Transparency is one of the reasons I wake up every day to contribute to open-source software. In this specific case, we keep the behavior configurable and document it. If we didn't document this and forced you to share anonymized data, then anyone questioning our transparency could have had a reasonable case. This is not the case here and to our knowledge, we are doing our best to communicate this.

> So, please add a small paragraph about deck analytics in README. It will increases credibility too.

I'm sorry but I'm not yet convinced that the README is the right place for such a detail. Our purpose of having a README is to provide an introduction and pointers to get started using this project.

marcelometal:
> I understand where you are coming from here but please note that we have absolutely no intentions of hiding any feature.

Yes, you are hiding this "feature" and tracking your users without consent.

This kind of information should be close to the source/binary (README or --help command)

scorphus:
I totally agree with @marcelometal here. These kind of "features" should be opt-in instead of opt-out. No question about it.

marcelometal:
https://github.com/Kong/deck/pull/301

PR #243, opened 2020-12-11
Fix --verbose

rainest:
We originally set the verbosity config after loading it from the `--verbose flag`. #225's changes to fix workspace issues [grouped that config set with other config settings, before loading the `--verbose` flag value](https://github.com/Kong/deck/compare/v1.2.2...v1.2.3#diff-ab967ab1a2f3a1b769106eeb7bfe892ef0e81d1d27811fa15be08e6749feee1fL157-L166). As a result, deck always uses the default verbosity.

This change moves the `--verbose` flag load above those settings, restoring the ability to control verbosity.

Originally reported internally in FTI-2144. FWIW, no-color is still after everything in that block, but that doesn’t appear to matter/it’s still applied correctly in 1.2.3. Didn’t dig into how that works, but it doesn’t appear that needs a similar fix, so left it alone.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -155,6 +155,8 @@ func initConfig() {
 		}
 	}
 
+	verbose = viper.GetInt("verbose")
+
 	rootConfig.Address = viper.GetString("kong-addr")
 	rootConfig.TLSServerName = viper.GetString("tls-server-name")
 	rootConfig.TLSSkipVerify = viper.GetBool("tls-skip-verify")
@@ -163,6 +165,5 @@ func initConfig() {
 	rootConfig.SkipWorkspaceCrud = viper.GetBool("skip-workspace-crud")
 	rootConfig.Debug = verbose >= 1
 
-	verbose = viper.GetInt("verbose")
 	noColor = viper.GetBool("no-color")
 }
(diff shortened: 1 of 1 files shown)


mflendrich:
@rainest,

The issue with `verbose` was because #225 moved the assignment after the point where we use the value. This would've been prevented from happening if `var verbose` hadn't been declared earlier as a global variable.

`verbose` does not need to be a global variable, it makes sense to get rid of it.

Similarly, `noColor` is a global variable, and the flow of setting `color.noColor` is quite convoluted (value obtained in `cobra.OnIntialize` and then applied in `PersistentPreRunE`). I don't see a reason not to set `color.noColor` directly in the former place.

See https://github.com/Kong/deck/commit/666ce733083afa530bcb416c0a6dbe353fc94cb3, which builds on top of this PR and streamlines handling of these parameters, that is:
- removes an intermediate global `var verbose` (which was the cause of confusion leading to this bug),
- removes an intermediate global `var noColor` and moves the `color.NoColor` assignment from `PersistentPreRunE` to the place where `var noColor` was previously initialized.

I recommend cherry-picking that commit into this PR if that looks good to you.

mflendrich (approved):

PR #246, opened 2020-12-16
feat(state) handle consumers with custom_id only

rainest:
This draft adds `custom_id` as one of the schema-allowed fields for consumers along with new lookup and de-duplication code to support them. This allows syncing files like:

```
_format_version: "1.1"
consumers:
- custom_id: foo
```

More in-depth background:

deck dumps and KIC usage of deck's libraries can omit IDs, but still allow syncing/diffing based on unique fields. At present, these fields are predetermined, and all entities have a single unique field only.

Consumers are (possibly) unique in that both their username and custom_id fiends are unique, and only one or the other is required. deck wants to look up consumers via username, however, so we impose a de facto requirement that you always populate the username, normally enforced by schema validity checks. KIC bypasses this check, however, so it can pass consumers without usernames, resulting in [a nil reference crash](https://github.com/Kong/kubernetes-ingress-controller/issues/550) when [attempting to retrieve a consumer by a non-existent username](https://github.com/Kong/deck/blob/v1.2.3/file/builder.go#L165).

The [exposed Consumer getter](https://github.com/Kong/deck/blob/v1.2.3/state/consumer.go#L84-L93) accepts some arbitrary string, and then attempts a lookup [against either the `id` or `Username` indices](https://github.com/Kong/deck/blob/v1.2.3/state/consumer.go#L67-L68). We effectively rely on a soft assumption that nobody attempts to create a problematic consumer set like:

```
{{"id":"32d32db5-e7a0-4d3f-a15d-1734fa0af347","username":"foo"},
{"id":"126d965b-6c59-474a-8858-c51c811f8a1e","username":"32d32db5-e7a0-4d3f-a15d-1734fa0af347"}}
```

That could theoretically trip up the [current getter](https://github.com/Kong/deck/blob/v1.2.3/state/consumer.go#L65-L93), which tries to [search all fields indices](https://github.com/Kong/deck/blob/v1.2.3/state/utils.go#L36-L55) without regards for for their contents. We could expand this to search to include the new CustomID index, which would allow collisions between these consumers:

```
{{"id":"32d32db5-e7a0-4d3f-a15d-1734fa0af347","username":"foo"},
{"id":"126d965b-6c59-474a-8858-c51c811f8a1e","custom_id":"foo"}}
```
We could avoid this by refactoring across the board to search only a single index, i.e. here we'd lookup by ID, Username, and CustomID each separately, and for services by ID and name separately, and so on, or keep that general pattern with an additional CustomID-only search in the event that the initial Username+ID search fails. This takes the consumer-specific approach.

### DIFF
--- a/file/builder.go
+++ b/file/builder.go
@@ -162,7 +162,13 @@ func (b *stateBuilder) consumers() {
 	for _, c := range b.targetContent.Consumers {
 		c := c
 		if utils.Empty(c.ID) {
-			consumer, err := b.currentState.Consumers.Get(*c.Username)
+			var consumer *state.Consumer
+			var err error
+			if !utils.Empty(c.Username) {
+				consumer, err = b.currentState.Consumers.Get(*c.Username)
+			} else if !utils.Empty(c.CustomID) {
+				consumer, err = b.currentState.Consumers.GetByCustomID(*c.CustomID)
+			}
 			if err == state.ErrNotFound {
 				c.ID = uuid()
 			} else if err != nil {
--- a/state/consumer.go
+++ b/state/consumer.go
@@ -25,6 +25,12 @@ var consumerTableSchema = &memdb.TableSchema{
 			Indexer:      &memdb.StringFieldIndex{Field: "Username"},
 			AllowMissing: true,
 		},
+		"CustomID": {
+			Name:         "CustomID",
+			Unique:       true,
+			Indexer:      &memdb.StringFieldIndex{Field: "CustomID"},
+			AllowMissing: true,
+		},
 		all: allIndex,
 	},
 }
@@ -47,7 +53,16 @@ func (k *ConsumersCollection) Add(consumer Consumer) error {
 	if !utils.Empty(consumer.Username) {
 		searchBy = append(searchBy, *consumer.Username)
 	}
-	_, err := getConsumer(txn, searchBy...)
+	_, err := getConsumer(txn, []string{"Username", "id"}, searchBy...)
+	if err == nil {
+		return fmt.Errorf("inserting consumer %v: %w", consumer.Console(), ErrAlreadyExists)
+	} else if err != ErrNotFound {
+		return err
+	}
+
+	if !utils.Empty(consumer.CustomID) {
+		_, err = getConsumer(txn, []string{"CustomID"}, *consumer.CustomID)
+	}
 	if err == nil {
 		return fmt.Errorf("inserting consumer %v: %w", consumer.Console(), ErrAlreadyExists)
 	} else if err != ErrNotFound {
@@ -62,10 +77,9 @@ func (k *ConsumersCollection) Add(consumer Consumer) error {
 	return nil
 }
 
-func getConsumer(txn *memdb.Txn, IDs ...string) (*Consumer, error) {
+func getConsumer(txn *memdb.Txn, indices []string, IDs ...string) (*Consumer, error) {
 	for _, id := range IDs {
-		res, err := multiIndexLookupUsingTxn(txn, consumerTableName,
-			[]string{"Username", "id"}, id)
+		res, err := multiIndexLookupUsingTxn(txn, consumerTableName, indices, id)
 		if err == ErrNotFound {
 			continue
 		}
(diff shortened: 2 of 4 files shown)


state/consumer.go
```diff
@@ -47,7 +53,16 @@ func (k *ConsumersCollection) Add(consumer Consumer) error {
 	if !utils.Empty(consumer.Username) {
 		searchBy = append(searchBy, *consumer.Username)
 	}
-	_, err := getConsumer(txn, searchBy...)
+	_, err := getConsumer(txn, []string{"Username", "id"}, searchBy...)
+	if err == nil {
+		return fmt.Errorf("inserting consumer %v: %w", consumer.Console(), ErrAlreadyExists)
+	} else if err != ErrNotFound {
+		return err
+	}
```
mflendrich, line 61: The way I understand the algorithm implemented in this function is:
1. Try to find consumer by ID.
2. If the above yielded not result, if username not empty, try to find the consumer by username.
3. If the above yielded no result, try to find the consumer by custom ID.
4. If any of the above yielded anything other than ErrNotFound, fail insertion.

Is it right? If so, the implementation appears somewhat complicated for reasons I don't understand - it first performs a multi-index lookup searching for both username and ID in both indices (on the condition of non-empty username), and then takes a different approach by doing a single-index lookup on custom ID.

Would a more streamlined alternative (pseudocode) work as well? Like:

```go
_, err := getConsumer(txn, []string{"id"}, *consumer.ID)
// return error if err == nil or err != ErrNotFound
_, err := getConsumer(txn, []string{"Username"}, *consumer.Username)
// return error if err == nil or err != ErrNotFound
_, err := getConsumer(txn, []string{"CustomID"}, *consumer.CustomID)
// return error if err == nil or err != ErrNotFound
```

Furthermore, the "find a potential collision on either ID or Username or CustomID" should, given its complexity, imo be wrapped in a function - "checkCustomerAlreadyExists" or something like that.

mflendrich: Also, should we test this (preventing a duplicate CustomID from being inserted)?

mflendrich (commented):
Some comments (referring purely to the logic that I could read from the code) below. I believe that @hbagdi, given that he has way more context than I do, should review this before merging as well.

file/builder.go
```diff
@@ -162,7 +162,13 @@ func (b *stateBuilder) consumers() {
...
 		c := c
 		if utils.Empty(c.ID) {
-			consumer, err := b.currentState.Consumers.Get(*c.Username)
+			var consumer *state.Consumer
+			var err error
+			if !utils.Empty(c.Username) {
+				consumer, err = b.currentState.Consumers.Get(*c.Username)
+			} else if !utils.Empty(c.CustomID) {
+				consumer, err = b.currentState.Consumers.GetByCustomID(*c.CustomID)
+			}
```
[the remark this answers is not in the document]
rainest: We make some (valid) assumptions about what deck has already done and what can possibly exist in the Kong database here.

The current database state may have a consumer with a username or custom ID that matches our target consumer's, in which case one of these `Get()` calls will successfully retrieve it, and we can update our target consumer with whatever ID we retrieved along with it. If both fail/no consumer is found, we know that our current state contains no such consumer, and we need to generate a new UUID for the target consumer, which we'll inject into the database later.

Theoretically you could hand-write a state file that has a consumer with neither, but schema validation (which occurs before this) will reject it.

state/consumer.go
rainest: That's the "refactoring across the board to search only a single index" approach. It's fine, but we'd then reasonably do the same for other resources. I'm not aware of others that'd actually need it, so I chose to preserve the status quo.

file/builder.go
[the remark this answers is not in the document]
hbagdi: I agree with what the reasoning here.
Question: We should ensure that KIC rejects (doesn't include the consumer in the target state) a consumer which doesn't contain both the `username` and `custom_id`.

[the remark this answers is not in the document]
hbagdi: +1, the state package is critical and heavily tested as a bug here can be disastrous.  This is a blocker for merge.

state/consumer.go
```diff
@@ -47,7 +53,16 @@ func (k *ConsumersCollection) Add(consumer Consumer) error {
 	if !utils.Empty(consumer.Username) {
 		searchBy = append(searchBy, *consumer.Username)
 	}
-	_, err := getConsumer(txn, searchBy...)
+	_, err := getConsumer(txn, []string{"Username", "id"}, searchBy...)
```
hbagdi, line 56: I might be wrong here but why can't we inject `custom_id` into `searchBy` and then add `CustomID` to the indices param?

mflendrich:
As @rainest said, there is no immediate plan to merge this PR. It will serve as reference for a future implementation.

hbagdi:
> As @rainest said, there is no immediate plan to merge this PR. It will serve as reference for a future implementation.

In that case, I'll close this PR.

PR #252, opened 2021-01-06
Fix regression preventing workspace creation

### LINKED ISSUE #251: deck does not create a workspace if a new workspace is needed
When the spec file contains a new workspace (or the --workspace parameter is used), the non-existant workspace is not created in deck version from 1.2.2 onwards.

Example state file starts like this;

```
_format_version: "1.1"
_workspace: haproxy-hc
services:
- connect_timeout: 60000
  host: status-upstream
  name: Healthcheck
  path: /status
  port: 80
  protocol: http
  read_timeout: 60000
  retries: 5
  write_timeout: 60000
  tags:
  - ha-proxy_healthcheck
  routes:
<snip?
```

Running deck with this command'
`$ deck sync --state=haproxy-hc.yaml --verbose 2`

New workspace creation is working with deck <= v1.2.1

```
HTTP/1.1 404 Not Found
Content-Length: 23
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: https://api.kong.lan:8445
Connection: keep-alive
Content-Type: application/json; charset=utf-8
Date: Wed, 06 Jan 2021 15:06:43 GMT
Server: kong/2.2.1.0-enterprise-edition
Vary: Origin
X-Kong-Admin-Latency: 110
X-Kong-Admin-Request-Id: 74oRmIAV4foKlRp4gtMhCWskRKJMozbU

{"message":"Not found"}
creating workspace haproxy-hc
POST /workspaces HTTP/1.1
Host: api.kong.lan:8444
Content-Length: 21
Content-Type: application/json
Accept-Encoding: gzip

{"name":"haproxy-hc"}
HTTP/1.1 201 Created
Content-Length: 699
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: https://api.kong.lan:8445
Connection: keep-alive
Content-Type: application/json; charset=utf-8
Date: Wed, 06 Jan 2021 15:06:43 GMT
Server: kong/2.2.1.0-enterprise-edition
Vary: Origin
X-Kong-Admin-Latency: 94
X-Kong-Admin-Request-Id: R2bvb3jZyQbPTRLKt2lHRhWMB922EibU

{"comment":null,"created_at":1609945603,"config":{"portal_cors_origins":null,"portal_invite_email":null,"portal_reset_success_email":null,"portal_reset_email":null,"portal_is_legacy":null,"portal_auth":null,"portal_developer_meta_fields":"[{\"label\":\"Full Name\",\"title\":\"full_name\",\"validator\":{\"required\":true,\"type\":\"string\"}}]","portal_auto_approve":null,"portal_emails_from":null,"portal_token_exp":null,"portal_emails_reply_to":null,"portal_auth_conf":null,"portal_access_request_email":null,"meta":null,"portal_approved_email":null,"portal_session_conf":null,"portal":false},"id":"cf9b3f84-e5fd-4fcd-9ac8-0dcabe695aed","name":"haproxy-hc","meta":{"color":null,"thumbnail":null}}
```

New workspace creation is not working with deck >= v1.2.2 due to the empty name parameter;

```
HTTP/1.1 404 Not Found
Content-Length: 46
Connection: keep-alive
Content-Type: application/json; charset=utf-8
Date: Wed, 06 Jan 2021 14:52:47 GMT
Server: kong/2.2.1.0-enterprise-edition
X-Kong-Admin-Latency: 1
X-Kong-Admin-Request-Id: X42cfPp9DgxK1doUS9vPztVlZjjVETHr

{"message":"Workspace 'haproxy-hc' not found"}
creating workspace haproxy-hc
POST /workspaces HTTP/1.1
Host: api.kong.lan:8444
Content-Length: 11
Content-Type: application/json
Accept-Encoding: gzip

{"name":""}
HTTP/1.1 400 Bad Request
Content-Length: 145
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: https://api.kong.lan:8445
Connection: keep-alive
Content-Type: application/json; charset=utf-8
Date: Wed, 06 Jan 2021 14:52:47 GMT
Server: kong/2.2.1.0-enterprise-edition
Vary: Origin
X-Kong-Admin-Latency: 109
X-Kong-Admin-Request-Id: apSQ0mjcbleBsdi0xJo0yKfxPeY4SuoC

{"message":"schema violation (name: length must be at least 1)","name":"schema violation","fields":{"name":"length must be at least 1"},"code":2}
Error: 400 Bad Request {"message":"schema violation (name: length must be at least 1)","name":"schema violation","fields":{"name":"length must be at least 1"},"code":2}
```

rainest:
https://github.com/Kong/deck/pull/225/files#diff-2b0546a527b3560f4e7f8501ef7907999b2e1a7f2498a60719630310f7778956R74 attempts to create a workspace with no name, as we only populate it in https://github.com/Kong/deck/pull/225/files#diff-2b0546a527b3560f4e7f8501ef7907999b2e1a7f2498a60719630310f7778956R84-R86

Fix appears to be using `wsConfig` instead; that works when testing manually.

@mflendrich it looks like this was probably unintentional leftovers from an earlier draft before the commit was pushed. Do you recall if there's any reason we would indeed want to use `rootConfig` here? Doesn't look like it based on a brief review of the [wsConfig population code](https://github.com/Kong/deck/blob/main/utils/types.go#L82-L87) and [go-kong workspace `Create()`](https://github.com/Kong/go-kong/blob/v0.13.0/kong/workspace_service.go#L14-L39), but to echo Harry's previous comment:

> A general node: This part of this codebase has become complex and fragile over time and we should test it. This has never been prioritized as there haven't been many quality issues in this area (yet).

Figured it was worth double-checking as such.

Fix #251

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -129,7 +129,7 @@ func syncMain(filenames []string, dry bool, parallelism, delay int, workspace st
 		}
 
 		if !dry {
-			_, err = rootClient.Workspaces.Create(nil, &kong.Workspace{Name: &rootConfig.Workspace})
+			_, err = rootClient.Workspaces.Create(nil, &kong.Workspace{Name: &wsConfig.Workspace})
 			if err != nil {
 				return err
 			}


cmd/common.go
```diff
@@ -129,7 +129,7 @@ func syncMain(filenames []string, dry bool, parallelism, delay int, workspace st
 		}
```
hbagdi, line 130: @mflendrich You usually have good testing tricks up your sleeve. Is there a way to test this **without** spinning up a Kong instance and **without** significant mocking overhead? I might be asking for too much.

Thinking out loud, we should be able to refactor and unit test large parts of this component.
However, that would be a separate task and not part of this PR.
Leaving this comment to have a brief discussion while we are here.

mflendrich: > without spinning up a Kong instance and without significant mocking overhead?

Without a Kong instance I cannot think of a test here that wouldn't be a "change detector" test.

> we should be able to refactor and unit test large parts of this component

For logic (like, translating something to something else) I'm all in for unit tests, however for integrations (like, "action X results in effect Y") I'd opt for tests featuring a real but minimal Kong. What makes it complex is that such a Kong would need to be a part of the test harness, now that we don't have a way to spin up a clean fresh small instance of Kong programmatically.

mflendrich (approved):

PR #254, opened 2021-01-06
fix: send analytics on a best effort basis

### LINKED ISSUE #253: decK analytics noticeably slow down commands which should complete fast
I recently discovered that the reason decK was slow in one of the environments I am using it in is that it's trying to send analytics to some server.

I found two major problems with the current decK analytics implementation (in addition to the ones pointed to in issue #136):

1. The analytics process will block completion of the decK process until the analytics can be sent. This is a problem because if Internet access is firewalled, any decK command could hang until the analytics timeout (3 seconds) even if it otherwise completes instantly.

2. The analytics server seems slow in general, so it will add a noticeable lag even if the analytics can be successfully sent.

The workaround I found is to disable analytics.

Logs of case 1: An environment that is firewalled and has no internet access (requests to outside servers hang):

```sh
$ docker run --rm --entrypoint='' -t kong/deck:v1.2.4 /bin/sh -c 'echo | time deck validate -s -'
real	0m 3.00s
user	0m 0.00s
sys	0m 0.00s
$ docker run --rm --entrypoint='' -e DECK_ANALYTICS=off -t kong/deck:v1.2.4 /bin/sh -c 'echo | time deck validate -s -'
real	0m 0.00s
user	0m 0.00s
sys	0m 0.00s
```

Logs of case 2: On my local computer which has internet access

```sh
$ docker run --rm --entrypoint='' -t kong/deck:v1.2.4 /bin/sh -c 'echo | time deck validate -s -'
real	0m 0.35s
user	0m 0.00s
sys	0m 0.00s
$ docker run --rm --entrypoint='' -e DECK_ANALYTICS=off -t kong/deck:v1.2.4 /bin/sh -c 'echo | time deck validate -s -'
real	0m 0.00s
user	0m 0.00s
sys	0m 0.00s
```

Part of the problem is that the analytics server is slower than it should:

```sh
$ time curl http://d.yolo42.com/polluting_your_statistics_hahaha -o /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

real	0m0,352s
user	0m0,007s
sys	0m0,000s
$ time curl http://www.google.com -o /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 13851    0 13851    0     0  98935      0 --:--:-- --:--:-- --:--:-- 98935

real	0m0,146s
user	0m0,007s
sys	0m0,000s
```

hbagdi:
With this patch, analytics is sent only on a best-effort basis.
Failure to send analytics doesn't result in any slowdown for the end
user.

Fix #253

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"context"
 	"fmt"
 	"net/url"
 	"os"
@@ -46,13 +47,16 @@ func Execute() {
 	const threads = 2
 	wg.Add(threads)
 
+	ctx, cancel := context.WithCancel(context.Background())
+
 	go func() {
-		sendAnalytics()
+		sendAnalytics(ctx)
 		wg.Done()
 	}()
 
 	go func() {
 		err = rootCmd.Execute()
+		cancel()
 		wg.Done()
 	}()
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -50,14 +50,14 @@ func Execute() {
 	ctx, cancel := context.WithCancel(context.Background())
 
 	go func() {
+		defer wg.Done()
 		sendAnalytics(ctx)
-		wg.Done()
 	}()
 
 	go func() {
+		defer wg.Done()
+		defer cancel()
 		err = rootCmd.Execute()
-		cancel()
-		wg.Done()
 	}()
 
 	wg.Wait()
--- a/cmd/analytics.go
+++ b/cmd/analytics.go
@@ -4,13 +4,11 @@ import (
 	"context"
 	"net/http"
 	"os"
-	"time"
 )
 
-func sendAnalytics() {
+func sendAnalytics(ctx context.Context) {
 	const (
-		minOSArgs        = 2
-		analyticsTimeout = 3 * time.Second
+		minOSArgs = 2
 	)
 
 	if os.Getenv("DECK_ANALYTICS") == "off" {
@@ -31,9 +29,6 @@ func sendAnalytics() {
 	// HTTP to avoid latency due to handshake
 	URL := "http://d.yolo42.com/" + cmd
 
-	ctx, cancel := context.WithDeadline(context.Background(),
-		time.Now().Add(analyticsTimeout))
-	defer cancel()
 	req, _ := http.NewRequestWithContext(ctx, "GET", URL, nil)
 	req.Header["deck-version"] = []string{VERSION}
 


rainest (approved):

cmd/root.go
```diff
@@ -41,22 +40,9 @@ It can be used to export, import or sync entities to Kong.`,
...
-
-	go func() {
-		err = rootCmd.Execute()
-		wg.Done()
-	}()
-
-	wg.Wait()
+	go sendAnalytics()
+
+	err := rootCmd.Execute()
```
mflendrich, line 45: I think that we should do it by leaving the `WaitGroup` as it was, but passing a cancellable context to `sendAnalytics` and cancelling it after `rootCmd.Execute()` finishes.

Otherwise you cannot guarantee clean execution (for example, honoring `defer`red calls)

mflendrich (changes requested):

cmd/root.go
hbagdi: I've updated the code to use context.

Are there any advantages to that approach other than clean execution (which doesn't matter here)?

cmd/root.go
```diff
@@ -46,13 +47,16 @@ func Execute() {
...
 	go func() {
-		sendAnalytics()
+		sendAnalytics(ctx)
 		wg.Done()
 	}()
 
 	go func() {
 		err = rootCmd.Execute()
+		cancel()
 		wg.Done()
```
mflendrich, line 60:
```suggestion
		defer wg.Done()
		defer cancel()
		err = rootCmd.Execute()
```

Without this change, the call to `cancel()` is not guaranteed.

cmd/root.go
```diff
@@ -46,13 +47,16 @@ func Execute() {
 	const threads = 2
 	wg.Add(threads)
 
+	ctx, cancel := context.WithCancel(context.Background())
+
 	go func() {
-		sendAnalytics()
+		sendAnalytics(ctx)
 		wg.Done()
```
mflendrich, line 54:
```suggestion
		defer wg.Done()
		sendAnalytics(ctx)
```

for consistency with the suggestion in lines below. Please accept either both or none.
On a side note, this has a side effect of the `wg.Wait()` call being guaranteed to happen.

cmd/root.go
hbagdi: > not guaranteed.

🤔 In which scenario will that fail?

mflendrich: If the control flow of the call stack gets interrupted for example by an ongoing `panic`.

hbagdi: My bad. I had a wrong understanding of the relationship between `panic` and `defer`.

cmd/root.go
mflendrich: Discussed out of band. Recording the conclusions for future reference:
1. Running a goroutine that has a possibility of outliving `main()` results in an unpredictable, abrupt shutdown of that goroutine. This leads to surprising interruption of execution, and no guarantee of execution of `defer` calls.
2. If `main()` gets refactored in the future (and e.g. put in a library), then return from the function containing this snippet won't necessarily cause program shutdown, and thus the spawned goroutine may continue to run indefinitely.
3. `context` is the solution of choice for interruption of operations (especially HTTP calls).

mflendrich (approved):

PR #258, opened 2021-01-15
fix(file): configFilesInDir extension match (#255)

### LINKED ISSUE #255: configFilesInDir(): extension match not anchored at end of filename
Summary: `configFilesInDir(dir)`, when it traverses the directory tree for yaml/json files, does not reliably identify json|yaml files.

Current behavior: `aaa.yaml.old` and `this-is-not-yaml.pdf` match. It might be the case that `/some/path/yaml/aaa.pdf` matches too, but I haven't checked.
Desired behavior: `aaa.yaml.old`and `this-is-not-yaml.pdf` don't match

The culprit is this regexp https://github.com/Kong/deck/blob/8e68931d7c3c050437d3f30dcd685b6e01246e9d/file/readfile.go#L114

mflendrich:
fixes #255 

Before this PR, `configFilesInDir()` treats paths `emptyfiles/not-a-file.yaml/info.txt`, `emptyfiles/foo.notyaml` and `emptyfiles/foo.yaml.pdf` as YAML/JSON files.
After this PR, `configFilesInDir()` no longer returns the above paths as yaml/json files.

The updated test did not pass before the code change and passed after the code change (TDD).

This PR replaces the regexp match in `Walk` by `strings.ToLower(filepath.Ext(path))`.

### DIFF
--- a/file/readfile.go
+++ b/file/readfile.go
@@ -6,7 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"regexp"
+	"strings"
 
 	ghodssyaml "github.com/ghodss/yaml"
 	"github.com/imdario/mergo"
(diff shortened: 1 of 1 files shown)


hbagdi (approved):
One less regexp import, you get a gold star!

Issue #260, opened 2021-01-28
deck fails with new kong parameter


michaelhsk:
decK Version 1.3.0
Kong Version 2.2.1

It seems that deck don't know about the new kong (route) parameter
response_buffering and request_buffering.

Don't found that parameter in a deck dump, and if i try to add them to a yaml file validate fails with:

$ ./deck validate -s ./testservice.yaml
Error: reading file: validating file content: 2 errors occurred:
        services.0.routes.0: Additional property request_buffering is not allowed
        services.0.routes.0: Additional property response_buffering is not allowed

testservice.yaml:
```
_format_version: "1.1"
_info:
  select_tags:
  - deck-test
services:
- connect_timeout: 60000
  host: localhost
  name: xservice
  port: 443
  protocol: https
  read_timeout: 60000
  retries: 5
  write_timeout: 60000
  tags:
  - deck-test
  routes:
  - name: xroutes
    paths:
    - /xxx/
    path_handling: v0
    preserve_host: false
    protocols:
    - http
    - https
    regex_priority: 0
    response_buffering: false
    strip_path: false
    https_redirect_status_code: 426
    tags:
    - deck-test
```

shaneutt:
This feature is now enabled in [v1.4.0](https://github.com/Kong/deck/releases/tag/v1.4.0)

PR #261, opened 2021-01-29
feat: add request and response buffering options

shaneutt:
- chore: update go-kong to v0.15.0
- feat: add req/resp buffering options to schema
- chore: update tests for req/resp buffering

fixes https://github.com/Kong/deck/issues/260

### DIFF
--- a/file/builder_test.go
+++ b/file/builder_test.go
@@ -1767,6 +1767,27 @@ func Test_stateBuilder(t *testing.T) {
 								},
 							},
 						},
+						{
+							Service: kong.Service{
+								Name: kong.String("large-payload-service"),
+							},
+							Routes: []*FRoute{
+								{
+									Route: kong.Route{
+										Name:              kong.String("dont-buffer-these"),
+										RequestBuffering:  kong.Bool(false),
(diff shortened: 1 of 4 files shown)


mflendrich (approved):
LGTM but please get a stamp from @hbagdi as well

hbagdi (approved):
I was going to suggest to [add defaults](https://github.com/Kong/deck/blob/f9a1df4facc8fd79bfd07d7d06ef094213097875/utils/constants.go#L14-L20) for these two new properties but we shouldn't do that in decK.
Kong choose a true by default API which is very problematic for this code-base.
In Go, differentiating between the zero value and absence of a boolean is not possible. Even with pointers, we have failed to achieve it because of limitations in libraries we use. If we do introduce the default of `true`, then we the user can't set them to `false`.

This patch looks good in the current form.

shaneutt:
Indeed, I had been looking at the default settings bu then realized that it really wont matter if the user doesn't configure the options at all, and if they do the defaults wont really matter anyhow. I'm hoping that the documentation I put over in `go-kong` for the relevant `Route` attributes will be a decent breadcrumb for anyone trying to integrate :thinking:

Issue #270, opened 2021-02-15
feature: Manage RBAC Entities


tjrivera:
## Problem
My team currently leverages Kong's Enterprise Edition RBAC features to manage access to our control-plane. 

We currently use some custom tooling that depends on [go-kong](https://github.com/kong/go-kong) to create custom roles, but we'd prefer to manage RBAC roles (and their associated endpoints and permissions) in a declarative format. i.e.

```yaml
_format_version: "1.1"
services:
- connect_timeout: 60000
  host: localhost
  name: foo-service
  path: /
  port: 8001
  protocol: http
  read_timeout: 60000
  retries: 5
  write_timeout: 60000
  tags:
  - foo
  routes:
  - id: 6c6a94ff-4404-4e80-8c57-e6506eea4492
    paths:
    - /foo
    path_handling: v0
    preserve_host: false
    protocols:
    - http
    regex_priority: 0
    strip_path: true
    tags:
    - foo
    https_redirect_status_code: 426
plugins:
- name: rate-limiting
  config:
    day: null
    header_name: null
    hour: 10000
    limit_by: consumer
    minute: null
    month: null
    path: null
    policy: local
    second: null
    year: null
  enabled: true
  protocols:
  - grpc
  - grpcs
  - http
  - https
rbac_roles:
- name: admin
  comment: Full access to all endpoints, across all workspaces—except RBAC Admin API
  is_default: false
  endpoint_permissions:
  - workspace: '*'
    endpoint: '*'
    actions:
    - delete
    - create
    - update
    - read
    negative: false
  - workspace: '*'
    endpoint: /rbac/*
    actions:
    - delete
    - create
    - update
    - read
    negative: true
  - workspace: '*'
    endpoint: /rbac/*/*
    actions:
    - delete
    - create
    - update
    - read
    negative: true
  - workspace: '*'
    endpoint: /rbac/*/*/*
    actions:
    - delete
    - create
    - update
    - read
    negative: true
  - workspace: '*'
    endpoint: /rbac/*/*/*/*
    actions:
    - delete
    - create
    - update
    - read
    negative: true
  - workspace: '*'
    endpoint: /rbac/*/*/*/*/*
    actions:
    - delete
    - create
    - update
    - read
    negative: true
- name: super-admin
  comment: Full access to all endpoints, across all workspaces
  is_default: false
  endpoint_permissions:
  - workspace: '*'
    endpoint: '*'
    actions:
    - delete
    - create
    - update
    - read
    negative: false
```

[...]

PR #271, opened 2021-02-16
Feat/issue 270 manage RBAC Roles and Endpoint Permissions

tjrivera:
Hi folks, 

This addresses #270.

Apologies in advance for the size of this patch. I tried to provide a patch that was atomic in its functionality and still somewhat manageable. I left out `RBACEntitiyPermission` management here with the goal of incorporating that assuming that the approach here fits expectations. 

### Notes

* A key difference with `RBACEndpointPermission` is that it doesn't have an ID. It relies on a composite key which is implemented by establishing an `Identifier()` method on the struct. Ref: https://github.com/Kong/deck/compare/main...tjrivera:feat/issue-270-manage-rbac-entities?expand=1#diff-d749ed180037f91f4f85e9d21a8a16ff32c26408a1025b16b26681cfb24c2b22R1090-R1095

This identifier is established as the `ID` for the endpoint permission in all interactions with gomemdb. 

* Permissions are relevant here only as child entities of a role. Permissions cannot be declared independent of roles.
* It may be worth documenting somewhere the method by which a developer should generate schema (which is very handy!) in case additional entities are added in the future.


hbagdi:
Thanks for your hard work here. I'll take a look and get back.

hbagdi:
The patch is good but we need to consider one user experience problem:

With this path, decK would expect RBAC entities to be present alongwith their configuration. If this is not the case, it would error out.

The reality is that most users want to manage configuration and RBAC separately and that is essential to implement access control correctly.
We need to figure out a way to not manage RBAC at same as configuration. Could be a flag based solution.
We can't use tags here because RBAC entities don't have tags (yet). We also need to make sure that the code doesn't error out if decK doesn't have enough privileges to invoke `/rbac/*` endpoints.

hbagdi:
@tjrivera I'm working cleaning this patch up a little and also writing a patch for the problem I mentioned in my last comment. I'll let you know once I've something more concrete.

tjrivera:
@hbagdi good point, I could imagine RBAC management being an opt-in flag. Something like `--include-rbac`, with that flag also being available to set via environment variables. 

Looking forward to your work on this patch. I'll stay tuned. Thanks.

hbagdi:
Following problems arise when we add RBAC resources into decK:
- Open-source doesn't have RBAC resources so we need to handle it transparently
- With most enterprise setups, we highly encourage to manage RBAC resources 
  separately. There is a separation of concern and privilege when in it comes to
  these two classes of resources. We would like this to be the case by default.
- With most decK use-cases, the credentials being used will/should not have
  access to CRUD RBAC resources. While this might not be the case always, we
  would like to build workflows that promote such a setup.

Keeping these problems and motivations in mind, I've updated the proposed patch
to add `--rbac-resources-only` flag to `diff`, `sync`, `dump` and `validate`
sub-commands. RBAC resources are managed only when this flag is provided.
In addition to that, when this flag is set, proxy configuration cannot be part
of the input state to decK.

I anticipate certain users will want to manage RBAC and proxy configurations
using the same credential. This workflow is possible by running `decK` twice,
once with the `--rbac-resources-only` and once without it. This can look
sub-optimal but given the motivations above, it seems justified. I'd love to
have some feedback in this area.

@tjrivera Thoughts?

tjrivera:
> Following problems arise when we add RBAC resources into decK:
> 
> * Open-source doesn't have RBAC resources so we need to handle it transparently
> * With most enterprise setups, we highly encourage to manage RBAC resources
>   separately. There is a separation of concern and privilege when in it comes to
>   these two classes of resources. We would like this to be the case by default.
> * With most decK use-cases, the credentials being used will/should not have
>   access to CRUD RBAC resources. While this might not be the case always, we
>   would like to build workflows that promote such a setup.
> 
> Keeping these problems and motivations in mind, I've updated the proposed patch
> to add `--rbac-resources-only` flag to `diff`, `sync`, `dump` and `validate`
> sub-commands. RBAC resources are managed only when this flag is provided.
> In addition to that, when this flag is set, proxy configuration cannot be part
> of the input state to decK.
> 
> I anticipate certain users will want to manage RBAC and proxy configurations
> using the same credential. This workflow is possible by running `decK` twice,
> once with the `--rbac-resources-only` and once without it. This can look
> sub-optimal but given the motivations above, it seems justified. I'd love to
> have some feedback in this area.
> 
> @tjrivera Thoughts?

@hbagdi, thanks for the insight here. 

I agree with your points here. It makes sense to me to separate the management of RBAC resources. In our current setup our team, which is more focused on operations of our API Gateway than the deployment of services, utilizes deck to maintain a core set of configuration that assists us in operating the platform -- we don't have a solid, declarative way to manage RBAC at this time and our hope is to manage everything declaratively if we can.

I recognize that our model does not represent the majority use case of deck, and think that this flag-based approach still works great for us. Ideally, we'd also be able to manage admin creation declaratively, we leverage SSO in our setup, so we wouldn't expect to have to manage credentials. I mention this because, while out of scope for this PR, I think it also fits this flag-based model should we decide to go down that route.

Thanks as always for your work here. If there's anything I can do to assist in crafting the changes you suggest or to review, please let me know.

hbagdi:
@tjrivera Closing this as I've picked this patch up in #276. I've tried my best to attribute commit correctly. Please let me know if I've missed anything.

PR #276, opened 2021-02-25
feat: add support for Kong Enterprise RBAC resources

### LINKED ISSUE #270: feature: Manage RBAC Entities

hbagdi:
This takes the patch from #271 and adds in some machinery to avoid any potential breakage for existing users.
Please see https://github.com/Kong/deck/pull/271#issuecomment-785917323 for reasoning behind additional commits on top of the initial patch.
Please review commit-by-commit.

Fix #270

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"os"
 
@@ -145,6 +146,9 @@ func syncMain(filenames []string, dry bool, parallelism, delay int, workspace st
 	if err != nil {
 		return err
 	}
+	if err := checkForRBACResources(*rawState, dumpConfig.RBACResourcesOnly); err != nil {
+		return err
+	}
 	targetState, err := state.Get(rawState)
 	if err != nil {
 		return err
@@ -215,4 +219,35 @@ func validateNoArgs(cmd *cobra.Command, args []string) error {
 			"Run 'deck --help' for usage.")
 	}
 	return nil
+}
+
+func checkForRBACResources(content utils.KongRawState,
+	rbacResourcesOnly bool) error {
+	proxyConfig := containsProxyConfiguration(content)
+	rbacConfig := containsRBACConfiguration(content)
+	if proxyConfig && rbacConfig {
+		common := "At a time, state file(s) must entirely consist of either proxy " +
+			"configuration or RBAC configuration."
+		if rbacResourcesOnly {
+			return fmt.Errorf("When --rbac-resources-only is used, state file(s) " +
+				"cannot contain any resources other than RBAC resources. " + common)
+		}
+		return fmt.Errorf("State file(s) contains RBAC resources. " +
+			"Please use --rbac-resources-only flag to manage these resources. " + common)
+	}
+	return nil
+}
+
+func containsProxyConfiguration(content utils.KongRawState) bool {
+	return len(content.Services) != 0 ||
+		len(content.Routes) != 0 ||
+		len(content.Plugins) != 0 ||
+		len(content.Upstreams) != 0 ||
+		len(content.Certificates) != 0 ||
+		len(content.CACertificates) != 0 ||
+		len(content.Consumers) != 0
+}
+
+func containsRBACConfiguration(content utils.KongRawState) bool {
+	return len(content.RBACRoles) != 0
 }
--- a/cmd/diff.go
+++ b/cmd/diff.go
@@ -54,6 +54,8 @@ func init() {
 		"select-tag", []string{},
 		"only entities matching tags specified via this flag are diffed.\n"+
 			"Multiple tags are ANDed together.")
+	diffCmd.Flags().BoolVar(&dumpConfig.RBACResourcesOnly, "rbac-resources-only",
+		false, "sync only the RBAC resources (Kong Enterprise only)")
 	diffCmd.Flags().BoolVar(&diffCmdNonZeroExitCode, "non-zero-exit-code",
 		false, "return exit code 2 if there is a diff present,\n"+
 			"exit code 0 if no diff is found,\n"+
--- a/cmd/dump.go
+++ b/cmd/dump.go
@@ -168,5 +168,6 @@ func init() {
 		"select-tag", []string{},
 		"only entities matching tags specified via this flag are exported.\n"+
 			"Multiple tags are ANDed together.")
-
+	dumpCmd.Flags().BoolVar(&dumpConfig.RBACResourcesOnly, "rbac-resources-only",
+		false, "export only the RBAC resources (Kong Enterprise only)")
 }
--- a/cmd/reset.go
+++ b/cmd/reset.go
@@ -135,4 +135,6 @@ func init() {
 		"select-tag", []string{},
 		"only entities matching tags specified via this flag are deleted.\n"+
 			"Multiple tags are ANDed together.")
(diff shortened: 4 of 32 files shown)


hbagdi:
cc @tjrivera

rainest:
Issues and points of interest found in user/black box testing:
- Endpoint permissions with a non-existent workspace simply fail and no `_workspace` simply fail:
  ```
  while processing event: {Create} rbac-endpointpermission 22021df1-864f-42db-a477-74860a659c87-nonono-* failed: HTTP status 404 (message: "Workspace nonono does not exist")
  ```
  This is probably expected--decK normally requires you handle all transactions on a per-workspace basis. This poses a bit of a challenge since roles with permissions in multiple workspaces are supported, albeit not perfectly (the GUI doesn't really handle them effectively).
- `reset` lacks the `--rbac-resources-only` flag. It does not reset RBAC resources when run without it.
- Roles with no name cause a segfault: [no-name.yaml.txt](https://github.com/Kong/deck/files/6045989/no-name.yaml.txt)
- Permissions with no workspace cause a segfault: [no-workspace.yaml.txt](https://github.com/Kong/deck/files/6046010/no-workspace.yaml.txt)

rainest:
Some of the above are easy to fix: https://github.com/Kong/deck/pull/278

- Roles without names segfaulting isn't easy to address within the existing structure of `file/builder.go`. We don't have any existing pattern of logging and skipping invalid resources or returning errors when this happens, and other resources, e.g. [services](https://github.com/Kong/deck/blob/v1.4.0/file/builder.go#L420) appear to have a similar segfault path. We could do something like this:
  ```
          for _, r := range b.targetContent.RBACRoles {
                  r := r
                  if utils.Empty(r.ID) {
  +                       if utils.Empty(r.Name) {
  +                               b.err = errors.New("RBAC role lacks name")
  +                               return
  +                       }
                          role, err := b.currentState.RBACRoles.Get(*r.Name)
  ```
  though that's not particularly graceful, since it just fails state building entirely and doesn't proceed onward. Functionally that's equivalent to the segfault, though it's a bit less ugly.
- DELETEs not working is due to some odd admin API behavior. Permission admin API endpoints include the permission endpoint in the path, and the permission sometimes includes a leading slash (e.g. `/acls`) and sometimes does not (`*` is the only such endpoint AFAIK). Calls from Manager omit the leading slash, e.g. `DELETE /rbac/roles/a79bdf2b-8844-424d-b6fc-97490cc6fe1d/endpoints/default/acls/*`, whereas we include it `DELETE /rbac/roles/a79bdf2b-8844-424d-b6fc-97490cc6fe1d/endpoints/default//acls/*`. Our PATCHes also use the double slash, but for some reason, those succeed, whereas DELETEs do not (rather, they 204, but the permission remains). We can't just chop that leading slash off entirely with something like:
  ```
  @@ -18,6 +19,8 @@ func rbacEndpointPermissionFromStruct(arg diff.Event) *state.RBACEndpointPermiss
                  panic("unexpected type, expected *state.RBACEndpointPermission")
          }
 
  +       trimmed := strings.TrimLeft(*ep.Endpoint, "/")
  +       ep.Endpoint = &trimmed
          return ep
   }
  ```
  since that makes PATCHEs start 404ing--they _require_ the `//`. DELETEs succeed, though they do throw a spurious `while post processing event: entity not found` after. POSTs succeed, since Kong accepts a body like `{"workspace":"*","endpoint":"wat","actions":"delete,create,update,read"}` (it automatically appends the leading `/`). The discrepancy in the verb behaviors suggests we'd need to handle this in go-kong with endpoint-specific logic. We apparently already have that for [GETs](https://github.com/Kong/go-kong/blob/v0.15.0/kong/endpoint_permission_service.go#L51), but not for [DELETEs](https://github.com/Kong/go-kong/blob/v0.15.0/kong/endpoint_permission_service.go#L112).

hbagdi:
Thanks for the great feedback @rainest!

> Endpoint permissions with a non-existent workspace simply fail and no _workspace simply fail

As you said, we can't do anything about this.

> Roles with no name cause a segfault

> Permissions with no workspace cause a segfault

We can take care of this using https://github.com/Kong/deck/pull/278#discussion_r583425002

> Permission deletes do not work

This is solvable. We need to change the endpoint when it is passed on to go-kong and not in the FromStruct method. That will delete the permission in Kong and not cause the postProcess error.

hbagdi:
@tjrivera Would you mind taking this PR for a spin and letting us know if it works for your environment and use-case?
@rainest Any other comments?

rainest (approved):
Nah, other than the now-fixed issues I mentioned previously, everything worked as I'd expect. Good to go from my perspective.

tjrivera:
I was able to build and test this branch today. Everything works as expected. Thanks, @hbagdi, and @rainest for the help in getting this over the finish line!

hbagdi:
Amazing, thanks for all your hard work and making this possible, @tjrivera!

hbagdi:
@tjrivera FYI, we will have a release of decK with a few other features in a couple of weeks and this will be included then.

tjrivera:
Awesome! great to see! Thanks @hbagdi and @rainest!

PR #278, opened 2021-02-26
Address missing functionality in Enterprise RBAC PR

rainest:
Addresses two issues found during testing for #276:

- Adds `--rbac-resources-only` support to `deck reset`.
- Rejects endpoint permissions with no workspace.

### DIFF
--- a/cmd/reset.go
+++ b/cmd/reset.go
@@ -135,4 +135,6 @@ func init() {
 		"select-tag", []string{},
 		"only entities matching tags specified via this flag are deleted.\n"+
 			"Multiple tags are ANDed together.")
+	resetCmd.Flags().BoolVar(&dumpConfig.RBACResourcesOnly, "rbac-resources-only",
+		false, "sync only the RBAC resources (Kong Enterprise only)")
 }
--- a/cmd/reset.go
+++ b/cmd/reset.go
@@ -136,5 +136,5 @@ func init() {
 		"only entities matching tags specified via this flag are deleted.\n"+
 			"Multiple tags are ANDed together.")
 	resetCmd.Flags().BoolVar(&dumpConfig.RBACResourcesOnly, "rbac-resources-only",
-		false, "sync only the RBAC resources (Kong Enterprise only)")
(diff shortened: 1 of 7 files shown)


cmd/reset.go
```diff
@@ -135,4 +135,6 @@ func init() {
 		"select-tag", []string{},
 		"only entities matching tags specified via this flag are deleted.\n"+
 			"Multiple tags are ANDed together.")
+	resetCmd.Flags().BoolVar(&dumpConfig.RBACResourcesOnly, "rbac-resources-only",
+		false, "sync only the RBAC resources (Kong Enterprise only)")
```
hbagdi, line 139: Copy pasta: description of the flag needs an update.

state/rbac_endpoint_permission.go
```diff
@@ -60,6 +61,10 @@ func (k *RBACEndpointPermissionsCollection) Add(rbacEndpointPermission RBACEndpo
 		return errEndpointRequired
 	}
 
+	if utils.Empty(rbacEndpointPermission.Workspace) {
+		return errWorkspaceRequired
+	}
```
hbagdi, line 66: Please remove this and instead add workspace as a required field in https://github.com/Kong/deck/blob/main/file/codegen/main.go.
While you are at it, could you also make RBAC role's name field to be always required?
Feel free to add stricter validation for fields that are always required for RBAC role and RBAC endpoint permission.

hbagdi (changes requested):
Thanks for reviewing the patch and going the extra mile to send a fix for the issues you had raised!

rainest:
Given that we're checking workspace and endpoint in bc46bb0, it seemed like we should remove the similar existing endpoint check in the original PR also, so ddccda0 removes both. Wasn't positive that was what you wanted.

Leading slash in delete issue is now also fixed by c967616

hbagdi:
Incorporated this with some cleanup in #278, hence closing this.
I've done my best to attribute commits correctly. Please let me know if I missed something.

PR #282, opened 2021-03-07
feat(dump) apply select-tags to credetentials

### LINKED ISSUE #217: --select-tag includes Consumer Creds when Consumer is Tagged But Creds are not
decK version: 1.2.1
Kong version: 2.0.3 CE

Steps to repro:
* Create a consumer and tag it "foo"
* Create a set of creds for that consumer and **do not** tag them
* Using decK dump the configuration while selecting the tag "foo"  `deck dump --select-tag foo`
* The resulting deck config file will contain the consumer credentials that are not tagged
* An attempt to sync with --select-tag foo attempts to tag the credentials with 'foo'

Usecase:
* We managed almost all the entities with deck except the credentials. The consumers managed by decK also have plugins and other consumer specific config so --skip-consumers does not work for us.

Let me know if additioanal information is needed.

hbagdi:
Fix #217

### DIFF
--- a/dump/dump.go
+++ b/dump/dump.go
@@ -497,9 +497,7 @@ func GetAllTargets(ctx context.Context, client *kong.Client,
 func GetAllKeyAuths(ctx context.Context,
 	client *kong.Client, tags []string) ([]*kong.KeyAuth, error) {
 	var keyAuths []*kong.KeyAuth
-	// tags are not supported on credentials
-	// opt := newOpt(tags)
-	opt := newOpt(nil)
+	opt := newOpt(tags)
 
 	for {
 		s, nextopt, err := client.KeyAuths.List(ctx, opt)
@@ -525,9 +523,7 @@ func GetAllKeyAuths(ctx context.Context,
 func GetAllHMACAuths(ctx context.Context,
 	client *kong.Client, tags []string) ([]*kong.HMACAuth, error) {
 	var hmacAuths []*kong.HMACAuth
-	// tags are not supported on credentials
-	// opt := newOpt(tags)
-	opt := newOpt(nil)
+	opt := newOpt(tags)
 
 	for {
 		s, nextopt, err := client.HMACAuths.List(ctx, opt)
@@ -553,9 +549,7 @@ func GetAllHMACAuths(ctx context.Context,
 func GetAllJWTAuths(ctx context.Context,
 	client *kong.Client, tags []string) ([]*kong.JWTAuth, error) {
 	var jwtAuths []*kong.JWTAuth
-	// tags are not supported on credentials
-	// opt := newOpt(tags)
-	opt := newOpt(nil)
+	opt := newOpt(tags)
 
 	for {
 		s, nextopt, err := client.JWTAuths.List(ctx, opt)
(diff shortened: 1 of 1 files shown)


hbagdi:
The problem with this patch is that it can be a potential breaking change for some users and almost all users of Kong's Ingress Controller.

rainest:
What's the expected controller breakage here? It was already tagging its managed credentials, so adding tag filters to the GET requests shouldn't affect it, correct?

hbagdi:
> What's the expected controller breakage here? It was already tagging its managed credentials, so adding tag filters to the GET requests shouldn't affect it, correct?

This patch results in two problems which are caused by the same underlying problem:
- decK users: With almost complete certainty, we can say that existing credentials in Kong are not tagged. Imagine if a user is using `--select-tags` feature to manage some configuration which contains credentials. With this patch, the current state that decK builds won't contain those credentials, but in reality, those credentials do exist in Kong's database. decK sync will fail in this case
- KIC deployments backed by a database will have the exact same problem.


It is possible to solve this problem in KIC's case by tagging each credential that KIC is managing using some code in the controller and this can be done in a transparent fashion. I've not yet been able to come up with a solution for the standalone decK case.

rainest:
This shouldn't affect existing controller-managed credentials since sync already does apply `select_tags` content to credentials (per #217), and the controller [does add that when generating config](https://github.com/Kong/kubernetes-ingress-controller/blob/1.2.0/pkg/deckgen/generate.go#L151-L155).

The GETs to retrieve current config after on subsequent syncs do still retrieve those, since the lack of a filter in the current dump package doesn't _omit_ the properly-tagged entities, it just also retrieves credentials that lack the tag. That's fine for the controller: both the target and current state contain all credentials the controller wants to create.

What this would change for the controller is that it'd no longer see manually-created credentials without its tag and delete them, and that's good: you _don't_ want the controller to remove things it shouldn't manage.

This change shouldn't be breaking even for non-controller instances, since the effect of not having it is that all your credentials get tagged as soon as you perform a sync with `select_tag`. The only issue I can see is that there's no way to automatically strip tags from entities that sync shouldn't have updated, but we're stuck with that, since both intentionally and accidentally added tags look identical.

What would be a sequence of actions where this change would have a harmful effect? I've tried to think through possible scenarios where using a state file created before this change with a deck build after it would either delete entities or stop managing them, but can't see any where it would.

hbagdi:
I stand corrected.
3d9f90945d9d7d5369f53997592af548d8570405 ensured that tags are added to credentials as well and the thinking at the time was that proactively adding in tags will help with filtering based on tags in future i.e. this patch.

I tested the `main` HEAD of decK against Kong and it checks out as well.
@rainest Could you test out the KIC side once again to ensure this is the case indeed?

If that is the case, the next question to ask is if breaking compat with Kong < 1.4 is fine or not. I think it is but I'll check and get back.

rainest:
> Could you test out the KIC side once again to ensure this is the case indeed?

Smoke test looks fine. Built against decK at b857bb2, existing credentials remain in place, new credentials are created, deleted credentials are removed, and credentials with non-controller tags are untouched. Logs don't show repeated unnecessary changes to credentials.

rainest (approved):

hbagdi:
Awesome, the forward thinking in the past saved so much trouble here! More to come.

PR #285, opened 2021-03-07
feat(cmd) ask confirmation from user before overwriting a file

### LINKED ISSUE #184: Deck should not overwrite existing state file without asking
Currently, `deck dump` will overwrite an existing state file without asking - I have lost some configuration that way.

E.g.:
```ShellCommand
$ deck dump
$ ls -hl
total 248
-rw-------  1 cwurm  wheel   121K  7 Jul 10:43 kong.yaml

$ deck dump -w Shared
$ ls -hl
total 8
-rw-------  1 cwurm  wheel   2.9K  7 Jul 10:44 kong.yaml
```

I think that ideally, `deck dump` would ask before overwriting an existing file, similar to how `deck reset` asks, and with a `-f` parameter to force it.

hbagdi:
Fix #184

### DIFF
--- a/cmd/dump.go
+++ b/cmd/dump.go
@@ -107,6 +107,12 @@ configure Kong.`,
 			return nil
 		}
 
+		if yes, err := confirmFileOverwrite(dumpCmdKongStateFile, dumpCmdStateFormat, assumeYes); err != nil {
+			return err
+		} else if !yes {
+			return nil
+		}
+
 		// Kong OSS
 		// or Kong Enterprise single workspace
 		if dumpWorkspace != "" {
@@ -170,4 +176,6 @@ func init() {
 			"Multiple tags are ANDed together.")
 	dumpCmd.Flags().BoolVar(&dumpConfig.RBACResourcesOnly, "rbac-resources-only",
 		false, "export only the RBAC resources (Kong Enterprise only)")
+	dumpCmd.Flags().BoolVar(&assumeYes, "yes",
+		false, "Assume 'yes' to prompts and run non-interactively")
 }
--- a/cmd/konnect_dump.go
+++ b/cmd/konnect_dump.go
@@ -31,6 +31,12 @@ configure Konnect.` + konnectAlphaState,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		httpClient := http.DefaultClient
 
+		if yes, err := confirmFileOverwrite(dumpCmdKongStateFile, dumpCmdStateFormat, assumeYes); err != nil {
+			return err
+		} else if !yes {
+			return nil
+		}
+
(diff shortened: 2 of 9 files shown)


cmd/dump.go
```diff
@@ -107,6 +107,14 @@ configure Kong.`,
 			return nil
 		}
 
+		ok, err := confirmFileOverwrite(dumpCmdKongStateFile, dumpCmdStateFormat, assumeYes)
+		if err != nil {
+			return err
+		}
+		if !ok {
+			return nil
+		}
```
mflendrich, line 116:
```suggestion
		if ok, err := confirmFileOverwrite(dumpCmdKongStateFile, dumpCmdStateFormat, assumeYes); err != nil {
			return err
		} else if !ok {
			return nil
		}
```

nit - the variables don't need to be function-scoped

Also another nit: `ok` is a name in Go that has a well-defined meaning (meaning that a type assertion or key lookup succeeded). How about changing this to `hasConsent`?

Also applies to other repetitions.

mflendrich (approved):
Some non-blocking bikeshedding attached. LGTM, apply optionally

hbagdi:
@mflendrich I had to resolve a couple of conflicts too, mind doing another pass?

cmd/utils.go
```diff
@@ -0,0 +1,48 @@
...
+	"os"
+	"strings"
+
+	"github.com/kong/deck/utils"
+)
+
+// confirm prompts a user for a confirmation with message
+// and returns true with no error if input is "yes" or "y" (case-insensitive),
+// otherwise false.
+func confirm(message string) (bool, error) {
```
mflendrich, line 15: Have you considered using a library?

For example:

https://pkg.go.dev/gopkg.in/AlecAivazis/survey.v1#Confirm

cmd/konnect_dump.go
```diff
@@ -31,6 +31,14 @@ configure Konnect.` + konnectAlphaState,
...
 		httpClient := http.DefaultClient
 
+		ok, err := confirmFileOverwrite(dumpCmdKongStateFile, dumpCmdStateFormat, assumeYes)
+		if err != nil {
+			return err
+		}
+		if !ok {
+			return nil
+		}
+
```
mflendrich, line 41: Could this block be made identical to the equivalent one in dump.go?

cmd/utils.go
hbagdi: We don't do much here so I avoided adding any dependencies.

cmd/konnect_dump.go
hbagdi: Updated.

mflendrich (approved):

PR #286, opened 2021-03-07
feat(file) implement env var substitution for state files

### LINKED ISSUE #191: Deck support placeholder replace
Hi,
is it possible to apply a sort of placeholder substitution into deck sync operation to support the multi environment configuration via single branch of SCM like git?

Regards,
Marcello

hbagdi:
Introducing support for envirment variable substitution in state files.
This is primarily intended for injecting sensitive data at runtime.

Fix #191

### DIFF
--- a/file/readfile.go
+++ b/file/readfile.go
@@ -2,11 +2,14 @@ package file
 
 import (
 	"bufio"
+	"bytes"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
+	"text/template"
 
 	ghodss "github.com/ghodss/yaml"
 	"github.com/imdario/mergo"
@@ -80,22 +83,26 @@ func getReaders(fileOrDir string) ([]io.Reader, error) {
 // readContent reads all the byes until io.EOF and unmarshals the read
 // bytes into Content.
 func readContent(reader io.Reader) (*Content, error) {
-	var content Content
-	var bytes []byte
 	var err error
-	bytes, err = ioutil.ReadAll(reader)
+	contentBytes, err := ioutil.ReadAll(reader)
 	if err != nil {
 		return nil, err
 	}
-	err = validate(bytes)
+	renderedContent, err := renderTemplate(string(contentBytes))
+	if err != nil {
+		return nil, fmt.Errorf("parsing file: %w", err)
+	}
+	renderedContentBytes := []byte(renderedContent)
+	err = validate(renderedContentBytes)
 	if err != nil {
 		return nil, errors.Wrap(err, "validating file content")
 	}
-	err = yamlUnmarshal(bytes, &content)
+	var result Content
+	err = yamlUnmarshal(renderedContentBytes, &result)
 	if err != nil {
 		return nil, err
 	}
-	return &content, nil
+	return &result, nil
 }
 
 // yamlUnmarshal is a wrapper around yaml.Unmarshal to ensure that the right
@@ -131,4 +138,33 @@ func configFilesInDir(dir string) ([]string, error) {
 		return nil, errors.Wrap(err, "reading state directory")
 	}
 	return res, nil
+}
+
+func getPrefixedEnvVar(key string) (string, error) {
+	const envVarPrefix = "DECK_"
+	if !strings.HasPrefix(key, envVarPrefix) {
+		return "", fmt.Errorf("environment variables in the state file must "+
+			"be prefixed with 'DECK_', found: '%s'", key)
+	}
+	value, exists := os.LookupEnv(key)
+	if !exists {
+		return "", fmt.Errorf("environment variable '%s' present in state file but not set", key)
+	}
+	return value, nil
+}
+
+func renderTemplate(content string) (string, error) {
+	t := template.New("state").Funcs(template.FuncMap{
+		"env": getPrefixedEnvVar,
+	}).Delims("${{", "}}")
+	t, err := t.Parse(content)
+	if err != nil {
+		return "", err
+	}
+	var buffer bytes.Buffer
+	err = t.Execute(&buffer, nil)
+	if err != nil {
+		return "", err
+	}
+	return buffer.String(), nil
 }
--- a/file/readfile_test.go
+++ b/file/readfile_test.go
@@ -131,6 +131,7 @@ func Test_getContent(t *testing.T) {
 	tests := []struct {
 		name    string
 		args    args
+		envVars map[string]string
 		want    *Content
 		wantErr bool
 	}{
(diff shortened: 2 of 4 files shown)


mflendrich (changes requested):
I think this can be a security vulnerability (used to exfiltrate environment variables if the deck file is controlled by an actor interested in host's environment variables).

Let's discuss how to mitigate this risk. My first guess is that restricting available env vars e.g. to those sharing a predefined prefix (e.g. `DECK_ENV_`) might solve it.

hbagdi:
There are two ways to achieve this:
- Ask user to include `DECK_` prefix to all env vars in the state file.
- Implicitly append (and document accordingly) `DECK_` prefix to any invocation, eg: `${{ env "FOO" }}` will be substituted with the value of DECK_FOO.

mflendrich:
@hbagdi 
> Which one do you prefer?

I find the former less "magical" and less confusing. The latter approach seems acceptable too, if it gets documented well enough.

rainest:
I'm also in favor of the first option, where users must include the `DECK_` prefix on variables in the state file. Using the same variable name in both the state file and environment feels simplest, and users need to be aware of the requirement either way to set the properly-named variables in the environment.

hbagdi:
@mflendrich @rainest Perfect. I get an odd sense of happiness we all agree AND use the exact same argument, however small the issue maybe. 
I've updated the patch to match our discussion.
I'm not sure what is up with dependabot.

file/readfile.go
```diff
@@ -132,3 +140,26 @@ func configFilesInDir(dir string) ([]string, error) {
...
+func processEnvs(content string) (string, error) {
+	const envVarPrefix = "DECK_"
+	t := template.New("state").Funcs(template.FuncMap{
+		"env": func(key string) (string, error) {
+			if !strings.HasPrefix(key, envVarPrefix) {
+				return "", fmt.Errorf("environment variables in the state file must "+
+					"be prefixed with 'DECK_', found: '%s'", key)
+			}
+			return os.Getenv(key), nil
+		},
```
mflendrich, line 153: For readability and maintainability, please extract the lambda function into a named function, for example `func GetPrefixedEnvVar`

file/readfile.go
```diff
@@ -132,3 +140,26 @@ func configFilesInDir(dir string) ([]string, error) {
 	}
 	return res, nil
 }
+
+func processEnvs(content string) (string, error) {
```
mflendrich, line 144: This function is in fact a gotemplate rendering mapper. It happens to have `env` as its defined function, but may include other ones in the future.

I recommend that we redefine the responsibility of this function from "fill in env vars" to "render the deck file template" and rename it accordingly.

file/readfile.go
```diff
@@ -87,6 +90,11 @@ func readContent(reader io.Reader) (*Content, error) {
 	if err != nil {
 		return nil, err
 	}
+	processedContent, err := processEnvs(string(bytes))
+	if err != nil {
+		return nil, fmt.Errorf("parsing file: %w", err)
+	}
+	bytes = []byte(processedContent)
```
mflendrich, line 97: Please don't reuse the same `[]byte` for input and output of template rendering. From a strongly typed perspective, these two values represent different data types (unrendered template versus rendered document)

file/readfile.go
```diff
@@ -2,6 +2,9 @@ package file
 
 import (
 	"bufio"
+	"bytes"
+	"fmt"
+	"html/template"
```
mflendrich, line 7: You probably meant `text/template`. `html/template` will perform additional escaping which I think is undesirable (will break valid YAML) here.

Please add a test ensuring that the template renderer doesn't escape yaml characters like & or <.

file/readfile.go
hbagdi: I get the point somewhat but not enough.
Why does it really matter? Or maybe when?

I do agree that this way of thinking is correct, I just don't completely grasp it so please help me improve my understanding here.

And since I agree, I have already updated the patch,

file/readfile.go
hbagdi: Done

file/readfile.go
hbagdi: Done.

file/readfile.go
hbagdi: Done.

hbagdi:
@mflendrich @rainest Friendly ping to review this when you get a chance.

file/readfile_test.go
```diff
@@ -194,15 +199,24 @@ func Test_getContent(t *testing.T) {
 			},
 			wantErr: false,
 		},
+		{
+			name:    "file with bad environment variable",
+			args:    args{[]string{"testdata/bad-env-var/file.yaml"}},
+			wantErr: true,
+		},
```
rainest, line 206: Test for the unset var suggestion earlier:

```suggestion
		},
    {
        name:    "unset environment variable in file",
        args:    args{[]string{"testdata/file.yaml"}},
        wantErr: true,
    },
```

rainest (changes requested):
Using `<` in tags to check that we're not garbling YAML seems a bit off since it's not actually a valid tag, but ultimately that doesn't cause issues in unit tests.

I found one UX gotcha during practical testing: unset variables in the template will (probably, depending on the field type) result in an invalid state, and the resulting error doesn't make the issue cause clear. For example, attempting to apply `file/testdata/file.yaml` without setting `DECK_SVC2_HOST` results in:

```
	services.0.host: Invalid type. Expected: string, given: null
```
With the attached change:

```
Error: reading file: parsing file: template: state:3:12: executing "state" at <env "DECK_SVC2_HOST">: error calling env: environment variable 'DECK_SVC2_HOST' present in state file but not set
```
While the complete error unwrap looks a bit silly, it's at least clearer than the type error that you'll hit later.

Originally attempted to use PR suggestions, and then realized that works horribly when you need to include tabs, so `git apply /tmp/lookup.diff.txt`: [lookup.diff.txt](https://github.com/Kong/deck/files/6307170/lookup.diff.txt)

hbagdi:
> I found one UX gotcha during practical testing: unset variables in the template will (probably, depending on the field type) result in an invalid state, and the resulting error doesn't make the issue cause clear. 

Fair. What if we error out early whenever we encounter an unset environment variable?

hbagdi:
@rainest I updated the code to include your patch and also added a test case for it. PTAL.

ping @mflendrich

rainest (approved):

file/testdata/file.yaml
```diff
@@ -1,9 +1,11 @@
 services:
 - name: svc2
-  host: 2.example.com
+  host: ${{ env "DECK_SVC2_HOST" }}
   routes:
   - name: r2
     paths:
     - /r2
+  tags:
+    - '<'
```
mflendrich, line 9:
```suggestion
    - '<' # verifies that the templating engine does not perform character escaping
```

mflendrich (approved):

PR #287, opened 2021-03-07
chore: add a basic Makefile for the project

### LINKED ISSUE #277: Create a Makefile for the project
- [ ] Write a basic Makefile which includes targets for running test, updating auto-generated code, run linters
- [ ] Update Github Actions to use the Makefile

hbagdi:
Also updates github actions to use the Makefile

Fix #277


mflendrich (approved):
I'm concerned about the below, but that's non-blocking, so LGTM.

This PR contributes to https://github.com/Kong/kubernetes-ingress-controller/issues/934, though. We might want to find a way not to circumvent Make's prerequisite update detection.

But maybe it's not worth our time since builds are fast. I'll leave it to your best judgment. :shrug:

PR #290, opened 2021-03-11
fix(cmd) use separate variables for Konnect flags

### LINKED ISSUE #288: Deck always defaults to konnect.yaml instead of kong.yaml
With decK 1.5 the default Kong configuration file is `konnect.yaml` and no longer `kong.yaml`. Reading the code, I believe this is a mistake? The non-Konnect commands still have `kong.yaml` as the default, e.g. https://github.com/Kong/deck/blob/a640f10a2017d77eb7d88a283064a02427366017/cmd/dump.go#L152-L154

However, the same variable names are used in the Konnect commands, e.g. https://github.com/Kong/deck/blob/a640f10a2017d77eb7d88a283064a02427366017/cmd/konnect_dump.go#L80-L82

What I think is happening is that the first thing `StringVarP` does is set the variable to the default value in https://github.com/spf13/pflag/blob/85dd5c8bc61cfa382fecd072378089d4e856579d/string.go#L6-L9

And since the Konnect files are alphabetically after the non-Konnect files their default value will always overwrite the non-Konnect values during package initialization.

The different commands should probably just use different variables to avoid any crossover.

rainest:
Use separate variables to store Konnect command flag values where Konnect commands share flags with non-Konnect commands.

Cobra will otherwise update the variable contents with whichever flags it processes last, effectively overriding the non-Konnect defaults with the Konnect defaults.

I don't think anything wants a different default value currently other than the state file, but I think we should define new variables for everything that's not actually being inherited from a parent command for consistency and safety in the event of future changes to one command set's defaults.

Fix #288

### DIFF
--- a/cmd/konnect_diff.go
+++ b/cmd/konnect_diff.go
@@ -4,6 +4,12 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	konnectDiffCmdKongStateFile   []string
+	konnectDiffCmdParallelism     int
+	konnectDiffCmdNonZeroExitCode bool
+)
+
 // konnectDiffCmd represents the 'deck konnect diff' command.
 var konnectDiffCmd = &cobra.Command{
 	Use:   "diff",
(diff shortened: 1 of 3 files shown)


mflendrich (approved):
This problem is a perfect example of why we should avoid global state in decK.

Thanks @rainest for fixing this.

I recommend that we go further and isolate konnect commands into a subpackage.

PR #295, opened 2021-03-23
fix(build) update brew formula for 3.0.8

rainest:
Homebrew 3.0.8 enforces a new restriction on formula commands. They must
be separated into their components to avoid errors like the following:

```
Separate `system` commands into `"#{bin}/deck", "version"`
```
See https://travis-ci.com/github/Kong/homebrew-deck/builds/220979410 for an example failed build.

https://github.com/Homebrew/brew/pull/10873 is the related upstream change.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -36,4 +36,4 @@ brews:
   description: Declarative configuration for Kong
   skip_upload: true
   test: |
-      system "#{bin}/deck version"
+      system "#{bin}/deck", "version"


rainest:
Draft pending a successful homebrew-deck build with this change. CI is currently quite backlogged: https://travis-ci.com/github/Kong/homebrew-deck/builds/220985349

`audit` and `install` run successfully with this change applied in a local test environment, so AFAIK this is all that's needed, but the CI job is the source of truth.

shaneutt (approved):
LGTM, but note I don't have brew to test this myself.

PR #301, opened 2021-03-31
chore: temporarily remove analytics from being sent

hbagdi:


cmd/root.go
```diff
@@ -44,25 +42,7 @@ It can be used to export, import or sync entities to Kong.`,
...
-	}()
-
-	go func() {
-		defer wg.Done()
-		defer cancel()
-		err = rootCmd.Execute()
-	}()
-
-	wg.Wait()
+	err := rootCmd.Execute()
```
mflendrich, line 45: [optional]

Just realized that this `err` does not get printed anywhere. Shall we print it?

mflendrich (approved):

cmd/root.go
hbagdi: Good catch, added, Could you take another look?

cmd/root.go
```diff
@@ -43,29 +41,12 @@ It can be used to export, import or sync entities to Kong.`,
...
-	}()
-
-	go func() {
-		defer wg.Done()
-		defer cancel()
-		err = rootCmd.Execute()
-	}()
-
-	wg.Wait()
+func Execute() error {
```
mflendrich, line 44: Could this function be as simple as `return rootCmd.Execute()`?

mflendrich (approved):

PR #305, opened 2021-04-06
Use only one library for yaml handling : github.com/ghodss/yaml

mmorel-35:
Was there any particular reason for using different yaml libraries : 
* github.com/ghodss/yaml
* gopkg.in/yaml.v2

I replaced gopkg.in/yaml.v2 with github.com/ghodss/yaml as it was more frequently used


file/types_test.go
```diff
@@ -5,9 +5,9 @@ import (
 	"reflect"
 	"testing"
 
+	yaml "github.com/ghodss/yaml"
```
hbagdi, line 8:
```suggestion
	ghodss "github.com/ghodss/yaml"
```

file/writer.go
```diff
@@ -7,10 +7,10 @@ import (
 	"sort"
 	"strings"
 
+	yaml "github.com/ghodss/yaml"
```
hbagdi, line 10:
```suggestion
	ghodss "github.com/ghodss/yaml"
```

hbagdi (commented):
The reason for using ghodss/yaml is for compatiblitiy with JSON. This is especially important with plugin configuration which are free form JSON (from decK's perspective).

I think this is a good change overall. Although unlikely, it could potentially break something and we don't have enough tests to detect that.

I think we should make one change  to avoid mistakes in falling back to the more popular "gopkg.in/yaml.v2".

hbagdi (approved):
Yes to 1 less direct dependency! Thank you for your contribution!

hbagdi:
Thank you for your contribution. Please fill out the following form to claim your contributor swag:
https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#contributor-t-shirt.

PR #316, opened 2021-04-08
Keep actions up to date with Dependabot

mmorel-35:
Configure github-actions update with dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,8 +1,11 @@
 version: 2
-
 updates:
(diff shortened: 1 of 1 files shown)


mflendrich (approved):
Like I said on https://github.com/Kong/go-kong/pull/41 - this looks promising, let's see how this works in the wild.

mmorel-35:
@mflendrich , it seems like the indentation is not right, my bad, do you need a fixing PR?

PR #317, opened 2021-04-08
hotfix: fix double error message prints

hbagdi:
This incorrect change was introduced in 4717d53.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -41,8 +41,12 @@ It can be used to export, import or sync entities to Kong.`,
 // Execute adds all child commands to the root command and sets
 // sflags appropriately.
 // This is called by main.main(). It only needs to happen once to the rootCmd.
-func Execute() error {
-	return rootCmd.Execute()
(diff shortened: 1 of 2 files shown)


cmd/root.go
```diff
@@ -41,8 +41,12 @@ It can be used to export, import or sync entities to Kong.`,
 // Execute adds all child commands to the root command and sets
 // sflags appropriately.
 // This is called by main.main(). It only needs to happen once to the rootCmd.
-func Execute() error {
-	return rootCmd.Execute()
+func Execute() {
```
mflendrich, line 44: Did you consider setting `cobra.Command`'s `SilenceErrors` to `true` instead?

hbagdi: I did. I reverted to match the past to avoid further new surprises.

mflendrich (approved):
My preference is to go with `return err` and suppressing error display on the cobra level, but I understand your point and therefore won't block this PR.

PR #320, opened 2021-04-09
Separation of concern between lint and test

mmorel-35:


mflendrich:
In a project with more elaborate static analysis (or multiple unit testing workflows alongside static analysis) it would make sense to separate static analysis (incl. linting) out into a separate workflow. But at its current magnitude it seems like an overkill to me.

Thanks for this contribution @mmorel-35 - closing because I think it isn't a good time for this change.

PR #321, opened 2021-04-13
fix(dump) add workaround for free mode mtls-auths

### LINKED ISSUE #274: mTLS auth endpoint bubbles up error with Enterprise free mode
Enterprise images running in free/unlicensed mode return 403s for endpoints that are Enterprise-only.

`dump.GetAllMTLSAuths()` has logic to return an empty set of mTLS auth credentials when the response is a 404 (what OSS images return) but nothing for a 403, which bubbles back up to `dump.Get()` and upward: https://github.com/Kong/deck/blob/v1.4.0/dump/dump.go#L616-L622

Handling this is a bit tricky, as this endpoint can return a 403 in different situations. Either:
- Enterprise is running in free mode, and this endpoint is not available at all.
- Enterprise is running in licensed mode, but decK is using an RBAC user that lacks permissions for it.

Do we need to handle these differently? In free mode, the 403 is effectively equivalent to a 404, since the database can't possibly contain any mTLS auth entities. In licensed mode, mTLS auth entities may exist, but decK cannot see them.

Treating either type of 403 the same as a 404 in `dump.Get()` may be fine. An empty entity set isn't of much interest until you try to solve later. If you haven't asked decK to create any such entities, it should and will do nothing. If you have asked decK to create that entity type, that failure [will bubble up](https://github.com/Kong/deck/blob/v1.4.0/state/credentials.go#L81-L86), and should, as decK cannot reconcile the state.

rainest:
Treat a 403 on the mtls-auth endpoint the same as a 404. This handles the endpoint's behavior in Enterprise free mode gracefully, whereas previously the dump would fail entirely.

We eventually want a way to properly determine if an instance is in Enterprise free mode and avoid unavailable API calls entirely rather than relying on the return code.

Fix #274

### DIFF
--- a/dump/dump.go
+++ b/dump/dump.go
@@ -674,6 +674,17 @@ func GetAllMTLSAuths(ctx context.Context,
 		if kong.IsNotFoundErr(err) {
 			return mtlsAuths, nil
 		}
+		// TODO figure out a better way to handle unauthorized endpoints
+		// per https://github.com/Kong/deck/issues/274 we can't dump these resources
+		// from an Enterprise instance running in free mode, and the 403 results in a
+		// fatal error when running "deck dump". We don't want to just treat 403s the
+		// same as 404s because Kong also uses them to indicate missing RBAC permissions,
+		// but this is currently necessary for compatibility. We need a better approach
+		// before adding other Enterprise resources that decK handles by default (versus,
+		// for example, RBAC roles, which require the --rbac-resources-only flag).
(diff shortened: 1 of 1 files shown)


hbagdi (approved):

aaronhmiller:
y'all got a patch release for this planned?

ajeetshah:
I found it to be fixed in deck version 1.7.0 :)

PR #328, opened 2021-04-16
Add document type and service

rainest:
Adds a new Document type and API bindings for it.

Left type in the service source for now to allow me to keep an unrelated change to the base URL in types.go.


konnect/document_service.go
```diff
@@ -0,0 +1,197 @@
...
+	Published      *bool           `json:"published,omitempty"`
+	ServicePackage *ServicePackage `json:"-"`
+	ServiceVersion *ServiceVersion `json:"-"`
+}
+
+// TODO move to types.go
+
+// ParentEndpoint returns the path of the Service Package or Service Version a Document is attached to.
+// It returns an error if the Document has neither or both set
+func (d *Document) ParentEndpoint() (string, error) {
```
mflendrich, line 28: What do you think about an alternative/polymorphic API structure like below?

Assuming the following:
- a service package has a URL (`/api/service_packages/:id`)
- a service version has a URL (`/api/service_packages/:id`)
- a document can be a child of either a service package or a service version and its URL is the parent URL + `/documents/:id`

it stood out to me that `type Document` tries to do work that is the responsibility of its parents' types (by listing supported parent types, and needing to know the URL structure for each of these parent types)

If we make the following change:

- implement `func URL()` on the `ServicePackage` and `ServiceVersion` types:

    ```go
    func (p *ServicePackage) URL() string {
      return fmt.Sprintf("/api/service_packages/%s", p.ID)
    }
    
    func (v *ServiceVersion) URL() string {
      return fmt.Sprintf("/api/service_versions/%s", v.ID)
    }
    ```
- define an abstract type with method `URL`:
    ```go
    type URLer interface { // inspired by fmt.Stringer and the go tradition
    // of naming interfaces function name + 'er', there are certainly dozens
    // of better names out there
      URL() string // or maybe (string, error)
    }
    ```
- replace `Document`'s `ServicePackage` and `ServiceVersion` fields with
    ```go
    type Document struct {
      // ...
      Parent URLer `json:"-"`
    }
    ```
- implement `Document.URL`:
    ```go
    func (d *Document) URL() string {
      return fmt.Sprintf("%s/documents/%s", d.Parent.URL(), d.ID)
    }
    ```

we'll free `type Document` from the need to:
- know all the possible types of `Parent`
- know the _inner details_ of parent implementations (specifically: what URLs in the API parents reside under)
- have the logic to prevent two parents of different types being defined on a single Document

rainest:
Superseded by https://github.com/Kong/deck/pull/336

PR #330, opened 2021-04-20
feat: implement new convert command

hbagdi:


convert/convert.go
```diff
@@ -0,0 +1,118 @@
...
+	key := from + to
+	switch key {
+	case FormatKongGateway + FormatKonnect:
+		return convertKongGatewayToKonnect(inputFilename, outputFilename)
+	default:
+		return fmt.Errorf("cannot convert from '%s' to '%s' format", from, to)
+	}
+}
+
+func convertKongGatewayToKonnect(inputFilename, outputFilename string) error {
```
mflendrich, line 65: I'm unhappy that this function combines the responsibility with format conversion with the responsibility of storage, but I understand that refactoring the `package file` public interface is probably too much for this PR.

_Just a rant, no action required_

convert/convert.go
```diff
@@ -0,0 +1,118 @@
...
+		return err
+	}
+
+	for _, service := range inputContent.Services {
+		servicePackage, err := kongServiceToKonnectServicePackage(service)
+		if err != nil {
+			return err
+		}
+		inputContent.ServicePackages = append(inputContent.ServicePackages, servicePackage)
+	}
```
mflendrich, line 77: (blocker)
I think that we should decouple logic from presentation here.

requested action: Refactor this `for` and the rest of the conversion logic below into a `func` that takes a gateway `file.Content` as an argument and returns a konnect `file.Content` without changing the argument.

If this requires generating a deepcopy func for `type Content`, then it's probably the right time to do that.

convert/convert.go
```diff
@@ -0,0 +1,118 @@
...
+	// all other entities are left as is
+
+	if err := file.WriteContentToFile(inputContent, outputFilename, file.YAML); err != nil {
+		return err
+	}
+	return nil
+}
+
+func kongServiceToKonnectServicePackage(service file.FService) (file.FServicePackage, error) {
+	var serviceName string
```
mflendrich, line 91: nit: use `:=` in line 97

convert/convert.go
```diff
@@ -0,0 +1,118 @@
...
+
+	// all other entities are left as is
+
+	if err := file.WriteContentToFile(inputContent, outputFilename, file.YAML); err != nil {
+		return err
+	}
+	return nil
+}
+
+func kongServiceToKonnectServicePackage(service file.FService) (file.FServicePackage, error) {
```
mflendrich, line 90: one small quality improvement at a time will help us heal code quality issues: please ensure that this function (and other functions like this, in general) don't modify the input argument and return a copy instead.

convert/convert_test.go
```diff
@@ -0,0 +1,264 @@
...
+	"testing"
+
+	"github.com/kong/deck/file"
+	"github.com/kong/deck/utils"
+	"github.com/kong/go-kong/kong"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_validConversion(t *testing.T) {
+	type args struct {
```
mflendrich, line 14: nit: `args` feels a bit superfluous to me, I see this as a pattern in many tests but I don't see it as providing value over just inlining arguments in `tests`

Not a blocker or action item, just a scathing remark that I'd whisper if we were doing code review sitting in one room :joy:

convert/convert_test.go
```diff
@@ -0,0 +1,264 @@
...
+			}
+			zeroOutID(&got)
+			if !reflect.DeepEqual(got, tt.want) {
+				assert.Equal(t, tt.want, got)
+			}
+		})
+	}
+}
+
+func zeroOutID(sp *file.FServicePackage) {
```
mflendrich, line 180: Same comment as above about returning a copy instead of editing in place.

I know that this function is in a test (which, one might argue, limits the blast), but I don't feel that tests should have a lower quality bar than tested code.

Also, the rightmost part of the `a && b && c && d` is superfluous: if `v.Implementation.Kong.Service.ID` is nil, then setting it to nil won't hurt

mflendrich (commented):
This review is from a standpoint of concern that the quality of deck's codebase may be at risk of many small changes introducing small bits of tech debt and, as a reviewer, I'd like to deal with that by:
- preventing (or at least limiting) code's side effects (for example mutating arguments),
- enforcing that new code follows SOLID (most notably the "S") where practical

Several (non-binding non-blocking) style preferences attached as well.

Test coverage not reviewed - that will happen in another round.

convert/convert.go
hbagdi: Fair. `file `package certainly needs a bit of work and should be split into a couple of packages.

convert/convert_test.go
hbagdi: This piece is auto-generated. 🤷🏾

cmd/convert.go
```diff
@@ -0,0 +1,60 @@
...
+	Args: validateNoArgs,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		sourceFormat, err := convert.ParseFormat(convertCmdSourceFormat)
+		if err != nil {
+			return err
+		}
+		destinationFormat, err := convert.ParseFormat(convertCmdDestinationFormat)
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
hbagdi: Makes total sense.
I'm a bit torn: wouldn't that mean `convert` package is now aware of how `cmd` is using it? Maybe that's okay? How do you go about reasoning against/for this?

convert/convert_test.go
mflendrich: Interesting - what generator are you using?

convert/convert.go
hbagdi: SGTM. We will need to get https://github.com/Kong/deck/pull/332 merged in first.

cmd/convert.go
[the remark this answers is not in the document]
mflendrich: You're right - the orthodox way would IMO be to implement a wrapper type for Format that implements the interface. 

But Go's duck-typed interfaces allow us to take a shortcut here, and I think that shortcut is fine (because, worst case, somebody will need to implement that wrapper if they change the contract of `Format`)

In Java that wouldn't fly because it would need `Format` to explicitly implement that interface.

cmd/convert.go
```diff
@@ -0,0 +1,60 @@
...
+	convertCmdOutputFile        string
+)
+
+// convertCmd represents the convert command
+var convertCmd = &cobra.Command{
+	Use:   "convert",
+	Short: "Convert files in one format to another format",
+	Long: `Convert command converts files representing configuration in one format
+to another compatible format.
+and usage of using your command. For example: a configuration for 'kong-gateway'
```
rainest, line 23: Was something lost here? It looks like there was originally some more sentence before the "and usage..." here.

convert/convert_test.go
hbagdi: https://github.com/cweill/gotests via https://github.com/buoto/gotests-vim

cmd/convert.go
hbagdi: > and usage of using your command

This was a boilerplate that was left. Updated.

cmd/convert.go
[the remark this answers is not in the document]
hbagdi: This turned out to be significantly more complicated (layers of indirection, unnecessary FlagSet types) and I saw no real benefit of it so I gave up. Let me know if this is a blocker.

hbagdi:
@mflendrich I rebased this on top of #332 and ensured that input arguments are not being modified. PTAL.

hbagdi:
@mflendrich Friendly ping for another review.

cmd/convert.go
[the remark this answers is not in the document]
mflendrich: Let's pass on this.

convert/convert.go
```diff
@@ -0,0 +1,120 @@
...
+
+func validConversion(from, to Format) (bool, error) {
+	if fromMap, ok := conversions[from]; ok {
+		if _, ok := fromMap[to]; ok {
+			return ok, nil
+		}
+	}
+
+	return false, fmt.Errorf("cannot convert from '%s' to '%s' format", from, to)
+}
```
mflendrich, line 37: Comment part 1:

This duplicates the logic of the `switch` in `func Convert()`. Why do we need both? Why not just let `Convert()` fail if no suitable conversion is possible?

Comment part 2 (provided that we need to retain this function for some reason):

Each of the two return values of this function gives exact same information - the `error` additionally giving the error message. This makes the return value space unnecessarily complex: ((true, nil), (true, err), (false, nil), (false, err)).

If you rephrase this function as `func validateConversion(from, to Format) error`, you'll have only one return value. Easier to test and understand, more difficult to use in a buggy way.

Why: imagine that this function somehow ends up returning `false, nil` (because this is a natural way to say "the result evaluated to false with no runtime errors". In such case, the caller (`func Convert`) will erroneously `return nil` (saying "no error").

convert/convert.go
```diff
@@ -0,0 +1,120 @@
...
+	if valid, err := validConversion(from, to); !valid {
+		return err
+	}
+
+	switch {
+	case from == FormatKongGateway && to == FormatKonnect:
+		return convertKongGatewayToKonnect(inputFilename, outputFilename)
+	default:
+		return fmt.Errorf("cannot convert from '%s' to '%s' format", from, to)
+	}
```
mflendrich, line 61: We're passing `inputFilename` and `outputFilename` down the function calls several times without a clear reason, causing `convertKongGatewayToKonnect` to be concerned not only with conversion logic, but also with file handling.

Also, with more conversion format pairs added, file handling logic will need to be copied across all these format pair implementations.

How about doing file operations in `Convert` and having conversions as simple functional transformations (without the need to bloat actual conversion logic with the file handling concern)

```go
func Convert(inputFilename, outputFilename string, from, to Format) error {
  // probably unnecessary - see my comment above
	if valid, err := validConversion(from, to); !valid {
		return err
	}

	inputContent, err := file.GetContentFromFiles([]string{inputFilename})
	if err != nil {
		return err
	}

	var outputContent *file.Content
	switch {
	case from == FormatKongGateway && to == FormatKonnect:
		outputContent = convertKongGatewayToKonnect(inputContent)
	default:
		return fmt.Errorf("cannot convert from '%s' to '%s' format", from, to)
	}

	return file.WriteContentToFile(outputContent, outputFilename, file.YAML)
}
```

convert/convert.go
```diff
@@ -0,0 +1,120 @@
...
+	return file.FServicePackage{
+		Name:        &serviceName,
+		Description: kong.String("placeholder description for " + serviceName + " service package"),
+		Versions: []file.FServiceVersion{
+			{
+				Version: kong.String("v1"),
+				Implementation: &file.Implementation{
+					Type: utils.ImplementationTypeKongGateway,
+					Kong: &file.Kong{
+						Service: serviceCopy,
```
mflendrich, line 114:
```go
func wipeServiceName(s *file.Service) *file.Service {
  result := s.DeepCopy()
  result.Name = nil
  result.ID = kong.String(utils.UUID())
  return result
}
```

and
```suggestion
						Service: wipeServiceName(service),
```

hbagdi:
@mflendrich Resolved the comments. PTAL.

hbagdi:
ping for another review @mflendrich

mflendrich (approved):

PR #333, opened 2021-04-20
fix(dump) use konnect filename for konnect dump

### LINKED ISSUE #331: Wrong file prompt
`deck konnect dump` prompts to overwrite `kong.yaml`  while it actually should be `konnect.yaml`.

rainest:
Fix #331


mflendrich (approved):
Band-aid LGTM :adhesive_bandage: :+1: 

I propose #334 as a measure to prevent this problem from ever occuring again (and improving modularity of deck at the same time).

Issue #63, opened 2019-08-20
Figure out an end-to-end testing approach


hbagdi:
decK v0.5.0 had a very simple bug that rendered the release unusable.
This could have been easily could with an automated end-to-end testing setup, which we don't have at the moment.
As more changes are introduced, manual testing is becoming more expesive and proving not to be sufficient.
Figure out a path forward for how do we want to test decK in an automated fashion.

teunis90:
Going over your open tickets, I saw this one.

What about creating a full-featured `kong.yaml` file, provision Kong:
- Dump the kong file, and diff
- Call all HTTP endpoints manually.

That would not cover everything but would be a good start.

hbagdi:
That will cover part of it but not all. We need to test that decK is smart enough to handle cases where it can detect differences and out of sync cases, do reverse syncs too.

If you wish to contribute what you are suggesting, that would be a good addition. PR welcome!

hbagdi:
Some thoughts/questions?
- What do we test and what should take priority? There are numerous things that we need to include tests for:
  - *Diffing logic*: (mainly the diff package) This involves giving the diff engine two states and let decK traverse those states and output diff events. Verification involves ensure that the events generated are correct.
  - *e2e*: Given an state file, run decK and see what it outputs. This verifies if decK correct generates the right output on stdout. The verification trusts decK to do the right thing in this case if the text on stdout is right.
    - Something to add here would be to make API calls against Kong to ensure that the state that decK assumed that it successfully created actually matches the reality. This is meant to detect bugs inside the `solver` package and wiring of components inside decK.
    - This can be done against different versions of Kong.

Prashansa-K:
This is complete.

PR #335, opened 2021-04-21
fix(dump) address free mode workaround regression

rainest:
I goofed checking 8254879 and only confirmed the negative case :facepalm: 

We don't want this to happen if the mTLS auth credentials endpoint _is_ available:

```
$ /tmp/main-deck dump -o main.yaml
panic: interface conversion: error is nil, not *kong.APIError

goroutine 36 [running]:
github.com/kong/deck/dump.GetAllMTLSAuths(0xc80cb8, 0xc000139240, 0xc000172400, 0xf592b0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/home/traines/src/deck/dump/dump.go:671 +0x5b5
github.com/kong/deck/dump.getConsumerConfiguration.func8(0x0, 0x0)
	/home/traines/src/deck/dump/dump.go:128 +0xef
golang.org/x/sync/errgroup.(*Group).Go.func1(0xc0001dd2c0, 0xc000142f50)
	/home/traines/go/pkg/mod/golang.org/x/sync@v0.0.0-20201020160332-67f06af15bc9/errgroup/errgroup.go:57 +0x85
created by golang.org/x/sync/errgroup.(*Group).Go
	/home/traines/go/pkg/mod/golang.org/x/sync@v0.0.0-20201020160332-67f06af15bc9/errgroup/errgroup.go:54 +0x6b
10:16:08-0700 yagody $ /tmp/329-deck dump -o 329.yaml
panic: interface conversion: error is nil, not *kong.APIError
```

### DIFF
--- a/dump/dump.go
+++ b/dump/dump.go
@@ -660,18 +660,18 @@ func GetAllMTLSAuths(ctx context.Context,
 		if kong.IsNotFoundErr(err) {
 			return mtlsAuths, nil
 		}
-		// TODO figure out a better way to handle unauthorized endpoints
-		// per https://github.com/Kong/deck/issues/274 we can't dump these resources
-		// from an Enterprise instance running in free mode, and the 403 results in a
-		// fatal error when running "deck dump". We don't want to just treat 403s the
-		// same as 404s because Kong also uses them to indicate missing RBAC permissions,
-		// but this is currently necessary for compatibility. We need a better approach
-		// before adding other Enterprise resources that decK handles by default (versus,
-		// for example, RBAC roles, which require the --rbac-resources-only flag).
-		if err.(*kong.APIError).Code() == 403 {
-			return mtlsAuths, nil
-		}
 		if err != nil {
+			// TODO figure out a better way to handle unauthorized endpoints
+			// per https://github.com/Kong/deck/issues/274 we can't dump these resources
+			// from an Enterprise instance running in free mode, and the 403 results in a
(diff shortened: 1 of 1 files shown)


dump/dump.go
```diff
@@ -660,18 +660,18 @@ func GetAllMTLSAuths(ctx context.Context,
...
 		if err != nil {
+			// TODO figure out a better way to handle unauthorized endpoints
+			// per https://github.com/Kong/deck/issues/274 we can't dump these resources
+			// from an Enterprise instance running in free mode, and the 403 results in a
+			// fatal error when running "deck dump". We don't want to just treat 403s the
+			// same as 404s because Kong also uses them to indicate missing RBAC permissions,
+			// but this is currently necessary for compatibility. We need a better approach
+			// before adding other Enterprise resources that decK handles by default (versus,
+			// for example, RBAC roles, which require the --rbac-resources-only flag).
+			if err.(*kong.APIError).Code() == 403 {
```
hbagdi, line 672: I think this could still fail if err is not of type `*kong.APIError`.
Let's do a proper type assertion here:
```go
if kongErr, ok := err.(*kong.APIError); ok{
if kongErr.code() == 403 { return ... } 
}
```

mflendrich (commented):
+1 to Harry's comment, otherwise lgtm

Yet another reason why we need integration tests.

xref https://github.com/Kong/deck/issues/63 - to provide context about regressions we've encountered

mflendrich (approved):

PR #337, opened 2021-04-22
Order by name first and by id by default

mmorel-35:


file/types.go
```diff
@@ -253,12 +253,12 @@ type FTarget struct {
 
 // id is used for sorting.
 func (t FTarget) id() string {
```
mflendrich, line 255: Since the `id` function no longer returns the ID primarily, its name is misleading and it will lead to bugs.

Please rename:
- func `id()` to `sortKey()`
- interface `id` to `sortable`
- func `compareID` to `compareOrder`

mflendrich (changes requested):
The change looks good, but two more things are needed:
1. Please unit test the `id` (`sortKey`, as to be renamed to) functions on individual types, and the respective functions that sort their objects (in `file/writer.go`)
2. Please see the attached comment below

@hbagdi should we consider this a breaking change?

mflendrich (commented):
Renaming looks good - the other part of my review seems to remain unsolved:

> 1. Please unit test the `id` (`sortKey`, as to be renamed to) functions on individual types, and the respective functions that sort their objects (in `file/writer.go`)

hbagdi:
> @hbagdi should we consider this a breaking change?

This is borderline but I _feel_ this is NOT a breaking change:
- The schema of the state file is part of the [compat promise](https://docs.konghq.com/deck/compatibility-promise/). The ordering inside is not.
- This is a positive change that most users will welcome. I'm sure there are more people who want their git diffs of kong's configuration to look better.
- I can't think of a case where a change in ordering will break someone's workflow. The ordering was nondeterministic in the first place. I mean the ordering was based on ID but someone depending on UUID-based ordering sounds really odd.

Is that fair enough?

file/types.go
```diff
@@ -444,24 +444,27 @@ func (p *FPlugin) UnmarshalJSON(b []byte) error {
...
+func (p FPlugin) sortKey() string {
 	// concat plugin name and foreign relations
-	key := ""
-	key = *p.Name
-	if p.Consumer != nil {
-		key += *p.Consumer.ID
-	}
-	if p.Route != nil {
-		key += *p.Route.ID
+	if p.Name != nil {
```
hbagdi, line 450: non-blocker: you could simplify this. A plugin's name field can never be `nil`.

file/types.go
```diff
@@ -533,14 +536,14 @@ type FServicePackage struct {
...
+// sortKey is used for sorting.
+func (s FServicePackage) sortKey() string {
 	if s.Name != nil {
 		return *s.Name
 	}
+	if s.ID != nil {
+		return *s.ID
+	}
 	return ""
 }
```
hbagdi, line 548: The sort key for FServiceVersion seems to be missing and we should add that.
It can be done separately or part of this PR - either is fine.

hbagdi:
This looks good from a product perspective. Let's merge when technical review is complete.

file/types.go
mmorel-35: There was neither any  `compareID()` associated with this type, it seems like it was never sorted

file/types.go
mmorel-35: I thought so too, but then I realized that even if it's impossible, the `sortkey()` shouldn't trigger an error for that situation as it's responsability is to sort and not to validate what it is sorting. That's why I wrote it this way.

file/types.go
hbagdi: I've added https://github.com/Kong/deck/issues/348 to track it.
If you are interested, please take that on once you this PR is merged in.
Thanks you!

PR #338, opened 2021-04-23
refactor: use go-kong's header injection

hbagdi:
Why?
go-kong ships with header injection utility which can be used directly.
This was not possible when decK was first written.
This reduces some code duplication.

### DIFF
--- a/utils/types.go
+++ b/utils/types.go
@@ -109,31 +109,6 @@ func (kc *KongClientConfig) ForWorkspace(name string) KongClientConfig {
 	return result
 }
 
-// HeaderRoundTripper injects Headers into requests
-// made via RT.
-type HeaderRoundTripper struct {
-	headers []string
-	rt      http.RoundTripper
-}
-
-// RoundTrip satisfies the RoundTripper interface.
-func (t *HeaderRoundTripper) RoundTrip(req *http.Request) (*http.Response,
-	error) {
-	newRequest := new(http.Request)
-	*newRequest = *req
-	newRequest.Header = make(http.Header, len(req.Header))
-	for k, s := range req.Header {
-		newRequest.Header[k] = append([]string(nil), s...)
-	}
-	for _, s := range t.headers {
-		split := strings.SplitN(s, ":", 2)
-		if len(split) >= 2 {
-			newRequest.Header[split[0]] = append([]string(nil), split[1])
-		}
-	}
-	return t.rt.RoundTrip(newRequest)
-}
-
 // GetKongClient returns a Kong client
 func GetKongClient(opt KongClientConfig) (*kong.Client, error) {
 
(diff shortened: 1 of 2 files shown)


utils/types.go
```diff
@@ -156,10 +131,15 @@ func GetKongClient(opt KongClientConfig) (*kong.Client, error) {
...
 	if len(opt.Headers) > 0 {
-		c.Transport = &HeaderRoundTripper{
-			headers: opt.Headers,
-			rt:      defaultTransport,
+
+		headers := http.Header{}
+		for _, keyValue := range opt.Headers {
+			split := strings.SplitN(keyValue, ":", 2)
+			if len(split) >= 2 {
+				headers[split[0]] = []string{split[1]}
```
mflendrich, line 139: `http.Header` has an `Add()` function in its public interface.
Its behavior is slightly different than what is here now: `Add()` canonicalizes the header key format.

Should we use `Add()` instead of a map assignment here? If not, why not?

utils/types.go
```diff
@@ -156,10 +131,15 @@ func GetKongClient(opt KongClientConfig) (*kong.Client, error) {
 	defaultTransport.TLSClientConfig = &tlsConfig
 	c.Transport = defaultTransport
 	if len(opt.Headers) > 0 {
-		c.Transport = &HeaderRoundTripper{
-			headers: opt.Headers,
-			rt:      defaultTransport,
+
+		headers := http.Header{}
```
mflendrich, line 135: Please put this in a `func parseHeaders(raw []string) http.Header` and unit test that function.

utils/types.go
```diff
@@ -156,10 +131,15 @@ func GetKongClient(opt KongClientConfig) (*kong.Client, error) {
 	defaultTransport.TLSClientConfig = &tlsConfig
 	c.Transport = defaultTransport
 	if len(opt.Headers) > 0 {
```
mflendrich, line 133: This `if` seems superfluous at first glance. To keep code simple (and easier to test), could we always run with headers, even if the header list is empty?

utils/types.go
```diff
@@ -156,10 +131,15 @@ func GetKongClient(opt KongClientConfig) (*kong.Client, error) {
...
-		c.Transport = &HeaderRoundTripper{
-			headers: opt.Headers,
-			rt:      defaultTransport,
+
+		headers := http.Header{}
+		for _, keyValue := range opt.Headers {
+			split := strings.SplitN(keyValue, ":", 2)
+			if len(split) >= 2 {
```
mflendrich, line 138: I think we should return an `error` from `GetKongClient` when `len(split) < 2`.

utils/types.go
mflendrich: If we decide to implement the suggestion from https://github.com/Kong/deck/pull/338#discussion_r619089783, the return value will probably become `(http.Header, error)`

utils/types.go
hbagdi: Low level performance is the reason.
The implementation of the the RoundTrip will do unnecessary copies: https://github.com/Kong/go-kong/blob/main/kong/utils.go#L64-L75

In your opinion, what is the right place to have this guard? Inside RoundTrip or here? And why?
(I think inside the RoundTrip. Why? Make life easy for the caller. The caller shouldn't even know anything about the implementation.)

hbagdi:
@mflendrich This is ready for another round of review.

hbagdi:
Ping for another review @mflendrich.

utils/types.go
mflendrich: I think that the place for such optimizations is under the abstraction.

That said, I think that's too small of a detail to block this PR.

utils/types.go
```diff
@@ -188,6 +163,19 @@ func GetKongClient(opt KongClientConfig) (*kong.Client, error) {
...
 
+func parseHeaders(headers []string) (http.Header, error) {
+	res := http.Header{}
+	for _, keyValue := range headers {
+		split := strings.SplitN(keyValue, ":", 2)
+		if len(split) >= 2 {
+			res.Add(split[0], split[1])
+		} else {
+			return nil, fmt.Errorf("splitting header key-value '%s'", keyValue)
+		}
```
mflendrich, line 174: nit

https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

mflendrich (approved):

PR #343, opened 2021-04-23
fix:default value miss in route & upstream

w313062006:
add default value for PathHandling in route & HTTPSVerifyCertificate in upstream
set PathHandling to "v1" according https://docs.konghq.com/gateway-oss/2.3.x/admin-api/#add-route
set HTTPSVerifyCertificate to true according  https://docs.konghq.com/gateway-oss/2.3.x/admin-api/#add-upstream
fix https://github.com/Kong/kubernetes-ingress-controller/issues/1212
It's my first pr, thx review

### DIFF
--- a/utils/constants.go
+++ b/utils/constants.go
@@ -23,6 +23,7 @@ var (
 		RegexPriority: kong.Int(0),
 		StripPath:     kong.Bool(false),
 		Protocols:     kong.StringSlice("http", "https"),
+		PathHandling:  kong.String("v1"),
 	}
 	targetDefaults = kong.Target{
 		Weight: kong.Int(defaultWeight),
@@ -47,6 +48,7 @@ var (
 					Interval:     kong.Int(0),
 					HTTPStatuses: []int{429, 404, 500, 501, 502, 503, 504, 505},
 				},
+				HTTPSVerifyCertificate: kong.Bool(true),
 			},
 			Passive: &kong.PassiveHealthcheck{
 				Healthy: &kong.Healthy{
--- a/file/builder_test.go
+++ b/file/builder_test.go
@@ -395,6 +395,7 @@ func Test_stateBuilder_ingestRoute(t *testing.T) {
 						ID:            kong.String("538c7f96-b164-4f1b-97bb-9f4bb472e89f"),
 						Name:          kong.String("foo"),
 						PreserveHost:  kong.Bool(false),
+						PathHandling:  kong.String("v1"),
 						RegexPriority: kong.Int(0),
 						StripPath:     kong.Bool(false),
 						Protocols:     kong.StringSlice("http", "https"),
@@ -421,6 +422,7 @@ func Test_stateBuilder_ingestRoute(t *testing.T) {
 						ID:            kong.String("4bfcb11f-c962-4817-83e5-9433cf20b663"),
 						Name:          kong.String("foo"),
 						PreserveHost:  kong.Bool(false),
+						PathHandling:  kong.String("v1"),
 						RegexPriority: kong.Int(0),
 						StripPath:     kong.Bool(false),
 						Protocols:     kong.StringSlice("http", "https"),
@@ -1488,7 +1490,8 @@ func Test_stateBuilder_upstream(t *testing.T) {
 						Slots: kong.Int(42),
 						Healthchecks: &kong.Healthcheck{
 							Active: &kong.ActiveHealthcheck{
-								Concurrency: kong.Int(10),
+								HTTPSVerifyCertificate: kong.Bool(true),
+								Concurrency:            kong.Int(10),
 								Healthy: &kong.Healthy{
 									HTTPStatuses: []int{200, 302},
 									Interval:     kong.Int(0),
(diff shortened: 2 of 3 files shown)


rainest:
There are currently a number of places where the current tests don't expect these default values in the generated structs: https://github.com/Kong/deck/pull/343/checks?check_run_id=2421864715

[tests.diff.txt](https://github.com/Kong/deck/files/6367231/tests.diff.txt) adds them. @w313062006 can you download that diff, `git apply tests.diff.txt`, and push a new commit that includes those changes?


```
         	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -11,3 +11,3 @@
        	            	    Paths: ([]*string) <nil>,
        	            	-   PathHandling: (*string)(<nil>),
        	            	+   PathHandling: (*string)((len=2) "v1"),
        	            	    PreserveHost: (*bool)(false),
        	Test:       	Test_stateBuilder_ingestRoute
```
```
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -25,3 +25,3 @@
        	            	      HTTPSSni: (*string)(<nil>),
        	            	-     HTTPSVerifyCertificate: (*bool)(<nil>),
        	            	+     HTTPSVerifyCertificate: (*bool)(true),
        	            	      Type: (*string)((len=4) "http"),
        	Test:       	Test_stateBuilder_upstream
```

utils/constants.go
```diff
@@ -47,6 +48,7 @@ var (
 					Interval:     kong.Int(0),
 					HTTPStatuses: []int{429, 404, 500, 501, 502, 503, 504, 505},
 				},
+				HTTPSVerifyCertificate: kong.Bool(true),
```
hbagdi, line 51: (Blocker comment)
Please verify in test that the user can set this value to false in the state file and that it doesn't get set to `true` i.e. this value can be set to `false` in the file and that gets reflected in Kong's database.

w313062006: when set HTTPSVerifyCertificate true in constants.go,  https://github.com/Kong/deck/blob/19205e5408661377f079960f198128743d716b36/utils/defaulter_test.go#L355 test can't pass, when set HTTPSVerifyCertificate false in arg and want, mergo.Merge will always set HTTPSVerifyCertificate to true according https://github.com/imdario/mergo/issues/24

hbagdi:
There is another problem here.
With this patch, we are essentially breaking compatibility with older versions of Kong. 
The last version of Kong that will remain compatible will be the one which has all the fields that are being defaulted now.
IOW, a version of Kong which doesn't understand a specific field and if that field is being defaulted in decK will not be compatible anymore.
We could make this happen with decK 2.0.

utils/constants.go
w313062006: @hbagdi  I don't know "set this to false in the state file" exact  meaning and kongingress crd don't have HTTPSVerifyCertificateso param, so I code in getUpstreams func.
 https://github.com/Kong/kubernetes-ingress-controller/blob/2caa5240c2f10ade2781b0edf37250e3f38ab927/internal/ingress/controller/parser/parser.go#L1291
like 
image
getUpstream called in Build func, Build func creates a Kong configuration from Ingress and Custom resources defined in Kuberentes, so I think this change can simulate user change the  HTTPSVerifyCertificate value in state file?
then I print upstream data before and after file.Get func in https://github.com/Kong/kubernetes-ingress-controller/blob/2caa5240c2f10ade2781b0edf37250e3f38ab927/internal/ingress/controller/kong.go#L265
 
image

HTTPSVerifyCertificate really set to false. and I check kong config by admin api 
image

so this value can be set to false and reflected in Kong's database, right?

w313062006: @hbagdi  the next step is check and get this "Healthchecks.Active.HTTPSVerifyCertificate can be set to false"  test pass?

w313062006: > (Blocker comment)
> Please verify in test that the user can set this value to false in the state file and that it doesn't get set to `true` i.e. this value can be set to `false` in the file and that gets reflected in Kong's database.

I have test deck v1.1.0 which kic 0.9.1 used, I add default value logic to constants.go, I don't understand "set this value to false in the state file" exact meaning, so I add code in getUpstreams func
https://github.com/Kong/kubernetes-ingress-controller/blob/2caa5240c2f10ade2781b0edf37250e3f38ab927/internal/ingress/controller/parser/parser.go#L1291
image


this func called by Build func in https://github.com/Kong/kubernetes-ingress-controller/blob/0.9.x/internal/ingress/controller/parser/parser.go, Build func creates a Kong configuration from Ingress and Custom resources defined in Kuberentes, So I think this operator simulate user set value in the state file.

when I print data before and after file.Get func in https://github.com/Kong/kubernetes-ingress-controller/blob/2caa5240c2f10ade2781b0edf37250e3f38ab927/internal/ingress/controller/kong.go#L265
 
the same upstream is change, yes, current is false and after file.Get func is become true
image

so I think when no default value logic, the HTTPSVerifyCertificate value in state file will remain, file.Get has no bug. but when add default value logic, file.Get always overridden  HTTPSVerifyCertificate value. that's why "Healthchecks.Active.HTTPSVerifyCertificate can be set to false" test can't pass

GGabriele:
Since we merged https://github.com/Kong/deck/pull/573, we populate defaults directly from the Admin API, so this is no longer needed. Thanks for your contribution though!

Issue #325, opened 2021-04-15
Refactor plumbing to build a go-kong client


hbagdi:
In the cmd package, a kong client is instantiated using different inputs - sometimes with a workspace, sometimes not and this is a source of confusion and maintenance burden.

Look for ways to simplify this. We can even push changes to kong/go-kong if required.

Prashansa-K:
Thank you for filing this issue. We're keeping our focus on stability and customer facing fixes right now, so we're not planning to pick up this refactor in the near future. Closing for now — happy to review a PR if you'd like to contribute to the repo. Feel free to reopen if this turns out to be blocking something concrete.

PR #350, opened 2021-04-26
refactor: plumb a single context throughout the code

hbagdi:
This patch is an effort to improve the context propagation in decK, down
to every single operation.
In order to do that, a single root context is now being created in the
main function, and all contexts should be derived from it.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"math/rand"
 	"os"
@@ -11,22 +12,22 @@ import (
 	"github.com/kong/deck/cmd"
 )
 
-func registerSignalHandler() {
+func registerSignalHandler() context.Context {
+	ctx, cancel := context.WithCancel(context.Background())
 	sigs := make(chan os.Signal, 1)
-	done := make(chan struct{})
-	cmd.SetStopCh(done)
 	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
 
 	go func() {
 		sig := <-sigs
 		fmt.Println("received", sig, ", terminating...")
-		close(done)
+		cancel()
 	}()
+	return ctx
 }
 
 func main() {
-	registerSignalHandler()
-	cmd.Execute()
+	ctx := registerSignalHandler()
+	cmd.Execute(ctx)
 }
 
 func init() {
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -25,20 +25,12 @@ const (
 )
 
 var (
-	stopChannel chan struct{}
-	dumpConfig  dump.Config
-	assumeYes   bool
+	dumpConfig dump.Config
+	assumeYes  bool
 )
 
-// SetStopCh sets the stop channel for long running commands.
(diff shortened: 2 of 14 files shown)


rainest (dismissed):

hbagdi:
Will wait for @mflendrich for a review on this one.

mmorel-35:
The contexts of ping command seems to be missing is it deliberately?

hbagdi:
> The contexts of ping command seems to be missing is it deliberately?

Good catch. Added that.

mmorel-35:
https://github.com/Kong/deck/blob/76d64ae1a05bc1d4611a6dbe58a42c9bf386a78e/cmd/dump.go#L38

mmorel-35:
Shall those lines be removed with this pr?
https://github.com/Kong/deck/blob/64a98cc76ce956ba9d3901e97a211e7744b1f47a/.golangci.yml#L18-L20

hbagdi:
> Shall those lines be removed with this pr?

We can't because solver package has a bunch of nil ctx invocations.
This PR wasn't meant to be an exhaustive squash of all issues with ctx.

hbagdi:
Conflicts have been fixed. PTAL.

main.go
```diff
@@ -11,22 +12,22 @@ import (
...
-		close(done)
+		cancel()
 	}()
+	return ctx
 }
 
 func main() {
-	registerSignalHandler()
-	cmd.Execute()
+	ctx := registerSignalHandler()
```
mflendrich, line 29: Consider the following:

https://pkg.go.dev/os/signal#NotifyContext

`ctx, _ := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)`

mflendrich (approved):
I'm happy to see this change :+1:

LGTM! With one optional comment.

main.go
hbagdi: included this, this is a handy simplification

hbagdi: Said too soon, @mflendrich the only problem here is that when the signal arrives, there wouldn't be a message to stdout stating that the signal was actually received. This is quite helpful in debugging. Thoughts?

mflendrich: Hacky way (not recommended):

```go
go func() {
  select {
    case <-ctx.Done():
      fmt.Printf("context cancelled (%w), terminating...", ctx.Err())
  }
}
```

Clean way:
- Replace `os.Exit(1)` throughout the codebase with errors plumbed all the way through to `RunE` and to the return value of `cmd.Execute()`
- After `cmd.Execute()` returns, check if `ctx.Err()` is not nil. If it isn't, then the context got cancelled by the signal.

Both are somewhat tedious and may defeat the purpose of this simplification.

hbagdi: Thought and explored a few different ways to do it but I haven't found a good way to deal with it in this PR.
This also feels minor enough that we can leave it be for now. If it is a bigger issue, please LMK and I'll address with a quick follow up PR.
I'm going to go ahead and merge this in.

mmorel-35:
Part 1of #325

PR #358, opened 2021-04-30
Add diff/sync support for Konnect documents

rainest:
Follow-up to #336. This adds support for diff and sync on top of dump, along with the associated necessary code in the differ, file builder, and solver. Some notes:

- This _is_ prescriptive about the layout of the document content files on disk. They must match the structure used by dump. My rationale despite previous discussion otherwise was that (a) we basically have to be prescriptive about the service package and service version structure without some sort of manifest file and (b) that paths are really names, and there shouldn't be any reason to actually handle any more complex directory tree structure below the package and version directories.
- Originally, we intended to assume that the content directories would be under the same directory as konnect.yaml. In the course of writing this, I found that we actually support multiple state files, which is at odds with our original intent. The implementation in this PR uses the "first" state file's directory, but that's a cop-out (and I'm not even sure the order is consistent). How do we want to handle this? My first thought is that we use the state file directory if there's only a single state file, but if there are multiple, we have an additional flag to specify the content directories' location and require it when you provide multiple states.
- In some manual test runs, I found that `DeepEqual()` finds a spurious difference in documents attached to service packages under some unknown circumstances. I believe this is somehow related to the versions attached to the parent package, since [their addresses](https://gist.github.com/rainest/63f6f0dc01d87deb66e1f0d614bd4fee) are the only difference I spotted walking through the code, although the values within them are the same and they _shouldn't_ result in inequality detection. I disabled foreign checks, which clears that issue, and doesn't present any obvious issue--you can't change the foreign relationships of documents since they're embedded within their parent in the target state.


state/konnect_types.go
```diff
@@ -12,9 +12,41 @@ type Document struct {
...
+
+// Equal returns true if s1 and s2 are equal.
+func (d1 *Document) Equal(d2 *Document) bool {
+	return d1.EqualWithOpts(d2, false, false, false)
+}
+
+// EqualWithOpts returns true if d1 and d2 are equal.
+// If ignoreID is set to true, IDs will be ignored while comparison.
+// If ignoreTS is set to true, timestamp fields will be ignored.
+func (d1 *Document) EqualWithOpts(d2 *Document,
```
mflendrich, line 36: The problem of comparing structs with certain fields ignored has been solved already: [go-cmp](https://pkg.go.dev/github.com/google/go-cmp/cmp) with `FilterPath` and `Ignore` can do that and I've successfully used that in other projects.

Not requesting changes here - sharing for visibility of the alternative, potentially better, approach.

mflendrich (approved):
LGTM but see comments.

konnect/document_service.go
```diff
@@ -50,7 +50,7 @@ func (d *DocumentService) Delete(ctx context.Context, doc *Document) error {
 		return fmt.Errorf("document must have a Parent")
 	}
 
-	endpoint := doc.Parent.URL() + "/" + *doc.ID
+	endpoint := doc.Parent.URL() + "/documents/" + *doc.ID
```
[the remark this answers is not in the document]
mflendrich: This is a feature of Go that is different than, for example, in C++.

hbagdi:
Please hold from merging this.
I'll work with @rainest to get this merged.

hbagdi:
> Originally, we intended to assume that the content directories would be under the same directory as konnect.yaml. In the course of writing this, I found that we actually support multiple state files, which is at odds with our original intent. The implementation in this PR uses the "first" state file's directory, but that's a cop-out (and I'm not even sure the order is consistent). How do we want to handle this? My first thought is that we use the state file directory if there's only a single state file, but if there are multiple, we have an additional flag to specify the content directories' location and require it when you provide multiple states.

I think the state file will in most cases point to a directly rather than multiple files.
We also support a hierarchy of files, so multiple files in different directories are also supported.

> This is prescriptive about the layout of the document content files on disk. They must match the structure used by dump. My rationale despite previous discussion otherwise was that (a) we basically have to be prescriptive about the service package and service version structure without some sort of manifest file and (b) that paths are really names, and there shouldn't be any reason to actually handle any more complex directory tree structure below the package and version directories.

The point is that we don't want users to follow the specific structure. If they want to have a different structure for any reason, they should be allowed. Is there a technical blocker to implement that?

hbagdi:
BUG:
1.I created a service package and a version. Added a service document and an OAS spec.
2.`deck konnect dump`
3. Delete package from the UI
4. `deck konnect sync` -> throws 500 for the create document for version

rainest:
> I think the state file will in most cases point to a directory rather than multiple files.

We do totally support multiple files in different directories though. This is valid:

```
/tmp/kdeck konnect diff --konnect-email "traines+konnect@konghq.com" --konnect-password-file pass \
    -s /tmp/wat/konnect.yaml \
    -s /tmp/why/konnect.yaml
```

> The point is that we don't want users to follow the specific structure. If they want to have a different structure for any reason, they should be allowed. Is there a technical blocker to implement that?

From a design perspective, I'm inclined to use the single format for both dump (where we have to choose the layout anyway) and sync unless we have a known use case that would require the additional flexibility already. Limiting the decisions users have to make is beneficial if we don't know of a reason they'd actually need the flexibility.

If we do receive feedback that users need more flexibility here, I don't think we're locked in to this layout: we can add an alternative means of specifying the path while keeping this as a default, similar to how we default to kong.yaml/konnect.yaml but use your chosen state file if you specify another. We'd further have the benefit of understanding why users need the flexibility, allowing us to design around that need and bounce prototypes off interested users.

From a technical perspective, the more flexible approach requires some sort of manifest indicating the file locations AFAICT. IMO the easiest option for that is reintroducing the `LocalPath` field I had on Documents originally, and having them point to an absolute location. Earlier discussion indicated we didn't want this.

f9d1f12461a9423914ea8df551738f9bf1142078 shortens the content diffs. Content is stripped from the objects we pass to `getDiff()` (we still need to run that for the other fields) and passed through a unified diff instead, which is printed after the `getDiff()` output. It uses a Document-specific path in the `crud.Update` clause. If we need something similar for other types in the future, we should rework that into an interface.

7fb98fe891741d30e1fa85c8fe1c52cae24d0d4d makes Documents not arrays in the file. This aligns with what the GUI does, though not with the API, which does support multiple Documents as long as only one is published (they're just entirely unused in the GUI).

file/types.go
```diff
@@ -523,7 +523,7 @@ type FServiceVersion struct {
 	ID             *string         `json:"id,omitempty" yaml:"id,omitempty"`
 	Version        *string         `json:"version,omitempty" yaml:"version,omitempty"`
 	Implementation *Implementation `json:"implementation,omitempty" yaml:"implementation,omitempty"`
-	Documents      []FDocument     `json:"documents,omitempty" yaml:"documents,omitempty"`
+	Document       *FDocument      `json:"document,omitempty" yaml:"documents,omitempty"`
```
hbagdi, line 526: yaml tag needs to be updated to `documents`.

cmd/common_konnect.go
```diff
@@ -28,6 +28,11 @@ func syncKonnect(ctx context.Context,
 		return err
 	}
 
+	err = targetContent.PopulateDocumentContent(filenames)
```
hbagdi, line 31: Question for my understanding: Why is this a separate function and not not done when the content is being read/built?

konnect/document_service.go
```diff
@@ -74,14 +74,25 @@ func (d *DocumentService) Update(ctx context.Context, doc *Document) (*Document,
...
-	req, err := d.client.NewRequest("PATCH", endpoint, nil, doc)
+	// Document PATCHes run through POST validation logic. Attempting to PATCH a Published: true
+	// document without toggling Published results in a 400, as if you'd tried to POST another
+	// Published: true document under the same resource.
+	endpoint := fmt.Sprintf("%s/documents/%s", doc.Parent.URL(), *doc.ID)
+	delReq, err := d.client.NewRequest("DELETE", endpoint, nil, doc)
+	if err != nil {
+		return nil, err
+	}
+	putReq, err := d.client.NewRequest("PUT", endpoint, nil, doc)
```
hbagdi, line 85: Why can't we just do a PUT (avoid the delete)?

solver/utils.go
```diff
@@ -14,6 +19,45 @@ var (
 	differ = gojsondiff.New()
 )
 
+func getDocumentDiff(a, b *state.Document) (string, error) {
```
hbagdi, line 22: Note to self: study and evaluate this

solver/utils.go
```diff
@@ -14,6 +19,45 @@ var (
...
+		if err != nil {
+			return "", err
+		}
+		aContent = string(aBytes)
+		bContent = string(bBytes)
+	}
+	edits := myers.ComputeEdits(span.URIFromPath("old"), aContent, bContent)
+	contentDiff = fmt.Sprint(gotextdiff.ToUnified("old", "new", aContent, edits))
+
+	return objDiff + contentDiff, nil
```
hbagdi, line 58: Can we unit test this function?

hbagdi (changes requested):

rainest:
In retrospect, the f9d1f12 explanation would have benefited from an example. In this case, `swagger.json` is a single-line JSON file (the Swagger pet store). Performing any sort of naive text diff against this would print a large amount of text, since all changes would be within the one line. The JSON document diff uses `MarshalIndent()` to pretty-print the JSON blobs first and then performs a unified diff on the results. This only shows the section that changes:

```
$ /tmp/kdeck konnect diff --konnect-email "traines+konnect@konghq.com" --konnect-password-file pass
updating document /swagger.json  {        
   "id": "084708f1-9af4-4a38-b6d3-6c593b87c552",
   "path": "/swagger.json",
   "published": true
 }
--- old
+++ new
@@ -229,7 +229,7 @@
 						]
 					}
 				],
-				"summary": "Add a new pet bloo bloo bloo to the store",
+				"summary": "Add a new pet to the store",
 				"tags": [
 					"pet"
 				]
```
Alternative would be some sort of JSON structure-aware diff that can instead show diffs at an object path or a word diff that can better handle small changes in a long line. I don't know of a readily available Go library for these.

19ddb57 handles using the Document Path field as both the Path value when interacting with the Konnect API and a filepath relative to the state file directory. Having now implemented this, I'm further convinced that we should use a separate state file field (`localPath` or `filepath` or whatever) instead of a field that serves both purposes.

With a single field, we effectively discard the original path value retrieved from the API when dumping, and must rebuild it using `StripLocalDocumentPath()` and `FilenameToName()`. This is additional complexity, breaks with other decK objects (which we can marshal and unmarshal without additional transformations), and imposes the same sort of hidden requirements we'd wanted to avoid by not enforcing the dump filestructure, just elsewhere, and worse, as it overlaps with content that's actually sent to Konnect (whereas the enforced filenames were internal to decK).

cmd/common_konnect.go
rainest: The preceding function, `file.GetContentFromFiles()`, is a basic validation check before we call `getContent()`. `getContent()` iterates over all files and and [calls `readContent()` on each](https://github.com/Kong/deck/blob/v1.6.0/file/readfile.go#L30), which is in turn basically just a [passthrough to a YAML unmarshaler](https://github.com/Kong/deck/blob/v1.6.0/file/readfile.go#L94).

We can't handle reading Document content within the YAML unmarshaler, so we need some sort of separate function for that. We could place it at the end of `getContent()`, but this seems a better home for it as it's Konnect-specific, and `getContent()` runs for everything.

hbagdi:
@rainest you might want to rebase the feat/konnect-documents branch on top of main and resolve all the conflicts and then rebase wip/konnect-documents on top of feat/konnect-documents. I expect there to be significant merge conflicts.

cmd/common_konnect.go
```diff
@@ -28,6 +28,13 @@ func syncKonnect(ctx context.Context,
...
+	if err != nil {
+		return err
```
hbagdi, line 33:
```suggestion
		return fmt.Errorf("reading documents: %w", err)
```

solver/utils.go
```diff
@@ -14,6 +19,45 @@ var (
...
+	aContent := *a.Content
+	bContent := *b.Content
+	aCopy.Content = nil
+	bCopy.Content = nil
+	objDiff, err := getDiff(aCopy, bCopy)
+	if err != nil {
+		return "", err
+	}
+	var contentDiff string
+	if json.Valid([]byte(aContent)) {
```
hbagdi, line 34: Please add a check for bContent being a valid JSON as well.

solver/utils.go
```diff
@@ -14,6 +19,45 @@ var (
...
+		}
+		bBytes, err := json.MarshalIndent(bObj, "", "\t")
+		if err != nil {
+			return "", err
+		}
+		aContent = string(aBytes)
+		bContent = string(bBytes)
+	}
+	edits := myers.ComputeEdits(span.URIFromPath("old"), aContent, bContent)
+	contentDiff = fmt.Sprint(gotextdiff.ToUnified("old", "new", aContent, edits))
```
hbagdi, line 56: Can we please rename `old` and `new` to actual path to the file?

solver/utils.go
hbagdi: At the minimum, please abstract out the json content diff part into a separate function and unit test that.
This is a blocker for this merge.

utils/utils.go
```diff
@@ -79,3 +79,10 @@ func NameToFilename(name string) string {
 	s = strings.ReplaceAll(s, string(os.PathSeparator), url.PathEscape(string(os.PathSeparator)))
 	return s
 }
+
+// FilenameToName (partially) reverses NameToFilename, replacing all URL-encoded path separator characters
+// with the path separator character. It does not re-add a leading separator, because there is no way to know
+// if that separator was included originally, and only some names (document paths) typically include one.
+func FilenameToName(filename string) string {
+	return strings.ReplaceAll(filename, url.PathEscape(string(os.PathSeparator)), string(os.PathSeparator))
+}
```
hbagdi, line 88: Please unit test this function.

hbagdi:
> With a single field, we effectively discard the original path value retrieved from the API when dumping, and must rebuild it using StripLocalDocumentPath() and FilenameToName(). This is additional complexity, breaks with other decK objects (which we can marshal and unmarshal without additional transformations), and imposes the same sort of hidden requirements we'd wanted to avoid by not enforcing the dump filestructure, just elsewhere, and worse, as it overlaps with content that's actually sent to Konnect (whereas the enforced filenames were internal to decK).

This is certainly more complex in decK's code but it makes the end-user experience better by a) being flexible (flexibility is good ro not is a separate discussion) b) hiding away the wrongs of the API of Konnect.

> the same sort of hidden requirements we'd wanted to avoid by not enforcing the dump filestructure

Can you elaborate? How does it impose hidden requirements?
I do see the point that it overlaps with the setting in Konnect API but I think that's on Konnect API to solve and not decK.

Kong's API was mature enough that we could simply dump configuration as we saw in the API. This is not the case with Konnect's API so we can't treat both as the same thing. The argument of "this is how it was done for Kong" (unfortunately) doesn't apply to Konnect objects. This is sadly the case because of realities we live in. I do not like what we are doing here either but we have to put the end-user workflow and experience above code complexities in this area.

solver/utils.go
rainest: In the strictest sense of test, yes. 5327795 is uh... not great because multiline text wants are no fun.

solver/utils.go
rainest: Both versions share the same local path--sort of, anyway, since one was pulled from the Konnect API. There aren't actually separate local filepaths to show.

We potentially could reconstruct the local paths and then append suffixes to get something like `servicePackage/serviceVersion/foo.json.old` and `servicePackage/serviceVersion/foo.json.new`, but it's rather convoluted, since these are state.Document structs (which contain the path value we actually get from/push to Konnect) rather than file.Document structs (where the path is the local path, which you see in konnect.yaml). IMO trying to pass that along from the konnect.yaml read onward into the state struct and hiding it from the diff mechanism probably isn't worth the effort at present.

rainest:
> you might want to rebase the feat/konnect-documents branch on top of main and resolve all the conflicts and then rebase wip/konnect-documents on top of feat/konnect-documents. I expect there to be significant merge conflicts.

IMO merge this into feat/konnect-documents and then merge main into that before PRing feat/konnect-documents into main. There are a few merge conflicts, but nothing difficult to resolve based on an initial trial run. Rebasing would probably overcomplicate things because history is pretty long at this point.

Re https://github.com/Kong/deck/pull/358#issuecomment-840008314 the issue I have now is with the dual usage of the `path` field. We'd still offer the same level of flexibility to arrange content files as you like with separate `path` (for the Konnect API object) and `localPath` fields.

The hidden limitation arises from the translation between the value in the state file and the value sent to Konnect. If, for instance, you edit the state file to add a new version and document, you have to know that decK will strip off all but the filename from the path you use when setting the path value in Konnect. There's a tradeoff between economizing on state file fields and fields that have a distinct purpose that they satisfy well. The latter also being cleaner to implement is an additional point for that option in a holistic evaluation of the feature from both a UX and code maintainability/readability perspective.

hbagdi:
> IMO merge this into feat/konnect-documents and then merge main into that before PRing feat/konnect-documents into main. There are a few merge conflicts, but nothing difficult to resolve based on an initial trial run. Rebasing would probably overcomplicate things because history is pretty long at this point.

That will complicate the history a bit but that's fine too then.
Please organize the commits properly in that case. All the wip commits need to go away - commits should tell a story and not contain retakes of various scenes. Please consult the history of this repo for an idea if you need one.

solver/utils_test.go
```diff
@@ -0,0 +1,118 @@
...
+		{
+			name: "JSON",
+			args: args{
+				docA: &state.Document{
+					Document: konnect.Document{
+						Path: kong.String("foo"),
+						Parent: &konnect.ServiceVersion{
+							ID: kong.String("abc"),
+						},
+						Content: kong.String("{\"foo\":\"foo\",\"bar\":\"bar\"}"),
```
hbagdi, line 71: Please replace this and all other JSON strings with unescaped versions.
```suggestion
						Content: kong.String(`{"foo":"foo","bar":"bar"}`),
```

solver/utils.go
```diff
@@ -14,6 +19,48 @@ var (
...
+			return "", err
+		}
+	}
+	edits := myers.ComputeEdits(span.URIFromPath("old"), aContent, bContent)
+	contentDiff = fmt.Sprint(gotextdiff.ToUnified("old", "new", aContent, edits))
+
+	return objDiff + contentDiff, nil
+}
+
+func pprintJSONString(JSONString string) (string, error) {
```
hbagdi, line 50: Better to name functions what they do.
```suggestion
func prettyPrintJSONString(JSONString string) (string, error) {
```

solver/utils_test.go
```diff
@@ -0,0 +1,118 @@
...
+					Document: konnect.Document{
+						Path: kong.String("foo"),
+						Parent: &konnect.ServiceVersion{
+							ID: kong.String("abc"),
+						},
+						Content: kong.String("{\"foo\":\"foo\",\"bar\":\"bar\",\"baz\":\"baz\"}"),
+					},
+				},
+			},
+			want: " {\n   \"path\": \"foo\"\n }\n--- old\n+++ new\n@@ -1,4 +1,5 @@\n {\n \t\"bar\": \"bar\",\n+\t\"baz\": \"baz\",\n \t\"foo\": \"foo\"\n }\n\\ No newline at end of file\n",
```
hbagdi, line 84: This is extremely hard to read and much more hard to reason about.
Did you consider using backtick strings for this? It helps you with new lines and also avoids the unnecessary escaping.
See: https://yourbasic.org/golang/multiline-string/

hbagdi (changes requested):
Almost there.

.golangci.yml
```diff
@@ -34,3 +34,7 @@ issues:
     - gosec
     text:  "Expect WriteFile permissions to be 0600 or less"
     path: file/codegen/main.go
+  - linters:
+    - lll
+    text: "line is 180 characters"
+    path: solver/utils_test.go
```
hbagdi, line 40: Please use backticks for strings and then remove this linter exception.

solver/utils_test.go
rainest: I'm quite aware of those. Getting the want string indents to match output with literals rather than escape codes is unfortunately also quite annoying.

solver/utils_test.go
```diff
@@ -21,9 +21,15 @@ func Test_PrettyPrintJSONString(t *testing.T) {
...
+				jstring: `{"foo":"foo","bar":{"a": 1, "b": 2}}`,
 			},
-			want:    "{\n\t\"bar\": {\n\t\t\"a\": 1,\n\t\t\"b\": 2\n\t},\n\t\"foo\": \"foo\"\n}",
+			want: `{
+	"bar": {
+		"a": 1,
+		"b": 2
+	},
+	"foo": "foo"
+}`,
```
hbagdi, line 32: If you find this ugly, you could use https://github.com/lithammer/dedent to make this better. We should absolutely do this, not a blocker for merge.

hbagdi (approved):
Thank you for your work here!
I'll leave merge to you as per your convenience.

PR #363, opened 2021-05-03
Strip patch version from ping output

### LINKED ISSUE #361: deck ping returns bad version number
I have Kong Enterprise 2.3.3.1 running. 

```shell
$ deck ping
Successfully connected to Kong!
Kong version:  2.3.0
$
```

rainest:
Strip the patch version from ping output. We already strip it when generating a semver `Version` in `CleanKongVersion`, so it's always 0 if we print the full version.

Fix #361

### DIFF
--- a/cmd/ping.go
+++ b/cmd/ping.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


hbagdi:
I ran into this a couple times during refactor and I think we should pass `http :8001/ | jq ".version"` as is to the output here.
Users are free to do all the greps that they want. `CleanKongVersion()` exists for an internal reason to do defaulting properly and there is no need to use that for user-facing output.

hbagdi:
So, in essence, we should break up `kongVersion()` to output `version` as is. That also improves code-health a bit (SRP).

rainest:
Looked at adding an argument to `kongVersion()` to toggle cleaning on or off, but I'm not sure how much the non-semver stuff in our versions would be handled by ParseTolerant (https://pkg.go.dev/github.com/blang/semver#ParseTolerant suggests its transforms probably won't handle our extra version info).

At present, `kongVersion()` is also tasked with finding the root, and we have work in flight to split that out. If we don't want to go with the original change proposed here I think this should wait on https://github.com/Kong/go-kong/pull/48 or similar, so closing this for now. This fix isn't particularly urgent.

PR #372, opened 2021-05-07
feat(print) refactor and introduce DisableOutput

hbagdi:
- This packages refactors code in the print package to DRY the code out
- It also introduces a way to silence output (prep for #152)

### DIFF
--- a/print/color.go
+++ b/print/color.go
@@ -9,47 +9,64 @@ import (
 var (
 	// mu is used to synchronize writes from multiple goroutines.
 	mu sync.Mutex
+	// DisableOutput disables all output.
+	DisableOutput bool
+)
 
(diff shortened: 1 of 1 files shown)


print/color.go
```diff
@@ -9,47 +9,64 @@ import (
 var (
 	// mu is used to synchronize writes from multiple goroutines.
 	mu sync.Mutex
+	// DisableOutput disables all output.
+	DisableOutput bool
+)
+
+func printf(fn func(string, ...interface{}), format string, a ...interface{}) {
```
mflendrich, line 16: Please rename to `maybePrintf` and `maybePrintln`. Otherwise there'll be hell of a head-scratching when `printf` does not print :joy:

mflendrich (approved):

PR #373, opened 2021-05-07
feat: plumb context further into diff and solver packages

### LINKED ISSUE #90: fix: pass along context based on os signal to each request
If a user sends SIGINTERM to decK, then it should be able to exit quickly.
This can be achieved by passing along context down to requests.

hbagdi:
This is continuation of the effort started in 7b85526.

Fix #90

### DIFF
--- a/crud/registry_test.go
+++ b/crud/registry_test.go
@@ -1,6 +1,7 @@
 package crud
 
 import (
+	"context"
 	"fmt"
 	"testing"
 
@@ -28,15 +29,15 @@ func (t testActionFixture) invoke(op string, inputs ...Arg) (Arg, error) {
 	return res, nil
 }
 
-func (t testActionFixture) Create(input ...Arg) (Arg, error) {
+func (t testActionFixture) Create(ctx context.Context, input ...Arg) (Arg, error) {
 	return t.invoke("create", input...)
 }
 
-func (t testActionFixture) Delete(input ...Arg) (Arg, error) {
+func (t testActionFixture) Delete(ctx context.Context, input ...Arg) (Arg, error) {
 	return t.invoke("delete", input...)
 }
 
-func (t testActionFixture) Update(input ...Arg) (Arg, error) {
+func (t testActionFixture) Update(ctx context.Context, input ...Arg) (Arg, error) {
 	return t.invoke("update", input...)
 }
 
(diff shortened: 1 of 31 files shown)


crud/registry_test.go
```diff
@@ -100,26 +101,26 @@ func TestRegistryCreate(t *testing.T) {
 	err := r.Register("foo", a)
 	assert.Nil(err)
 
-	res, err := r.Create("foo", "yolo")
+	res, err := r.Create(nil, "foo", "yolo")
```
mflendrich, line 104: Why `nil`?

https://golang.org/pkg/context/ says
> Do not pass a nil Context, even if a function permits it. Pass context.TODO if you are unsure about which Context to use.

hbagdi: These are unit tests, does it matter?
Maybe from a referential code perspective...

hbagdi: ping @mflendrich

hbagdi:
Update:
The linter has now been turned on for all non-test code to catch for cases where we might be passing a nil context.

mmorel-35:
Part 2 of #325

crud/registry_test.go
mflendrich: The documentation for `context.Background` says

> It is typically used by the main function, initialization, and tests

Given that:
- the documentation says to never pass `nil`
- the documentation suggests that `context.Background` is intended for use in tests
- it's a matter of a single line `ctx := context.Background` and replacing `nil` by `ctx`
- functions that get a nil `context` will panic if they try to call any function on a nil context (and from the perspective of tests we shouldn't be assuming that they don't)

I can see strong evidence convincing that we should pass a valid context also in tests.

The cost is one `ctx := context.Background()` line per function and replacement of `nil`s with `ctx`s which doesn't look like a big deal to me.

mflendrich (approved):
Please see the comment

crud/registry_test.go
hbagdi: Okay. I've updated the code. The most convincing part was `nil` context point.

hbagdi:
I've update the PR to include changes that happen in main due to merge of Konnect documents work. Please take another look.

mflendrich (commented):
please resolve the conflicts before the final LGTM because the nature of these conflicts seems particularly bug-prone to me, so a second pair of eyes could help.

hbagdi:
@mflendrich That merge conflict took a while to resolve. PTAL.

hbagdi:
@mflendrich Again resolved the conflicts, please review at your earliest convenience.

mflendrich (approved):

PR #374, opened 2021-05-07
feat: configurable value of konnect API endpoint

### LINKED ISSUE #354: feature: a way to dynamically change the address of Konnect API endpoint
Konnect API's endpoint is currently hardcoded to `konnect.konghq.com` [here](https://github.com/Kong/deck/blob/19205e5408661377f079960f198128743d716b36/konnect/types.go#L7-L11).
This needs to be changed dynamically:
- internally at Kong Inc, various environments exists and Kong employees want to use decK with them
- Konnect has separate environments on different continents. These are available on different URLs, which means the endpoint address needs to change. 

The scope of this ticket is to:
- Figure out how to implement this, (a flag in cobra?)
- Implement it

hbagdi:
Fix #354

### DIFF
--- a/cmd/common_konnect.go
+++ b/cmd/common_konnect.go
@@ -36,7 +36,7 @@ func syncKonnect(ctx context.Context,
 	targetContent.StripLocalDocumentPath()
 
 	// get Konnect client
-	konnectClient, err := utils.GetKonnectClient(httpClient, konnectConfig.Debug)
+	konnectClient, err := utils.GetKonnectClient(httpClient, konnectConfig)
 	if err != nil {
 		return err
 	}
@@ -60,7 +60,7 @@ func syncKonnect(ctx context.Context,
 
 	// initialize kong client
 	kongClient, err := utils.GetKongClient(utils.KongClientConfig{
-		Address:    konnect.BaseURL() + "/api/control_planes/" + kongCPID,
+		Address:    konnectConfig.Address + "/api/control_planes/" + kongCPID,
 		HTTPClient: httpClient,
 		Debug:      konnectConfig.Debug,
 	})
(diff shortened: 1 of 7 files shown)


cmd/root.go
```diff
@@ -126,6 +126,11 @@ func init() {
 		"File containing password to your Konnect account")
 	viper.BindPFlag("konnect-password-file",
 		rootCmd.PersistentFlags().Lookup("konnect-password-file"))
+
+	rootCmd.PersistentFlags().String("konnect-addr", "https://konnect.konghq.com",
```
mflendrich, line 130: Why do these flags need to be on `rootCmd` and not on `konnectCmd`?

konnect/client.go
```diff
@@ -34,14 +34,18 @@ type Client struct {
 	debug                 bool
 }
 
+type ClientOpts struct {
+	BaseURL string
+}
```
mflendrich, line 39: please add docstrings

cmd/common_konnect.go
```diff
@@ -52,7 +52,7 @@ func syncKonnect(ctx context.Context,
 
 	// initialize kong client
 	kongClient, err := utils.GetKongClient(utils.KongClientConfig{
-		Address:    konnect.BaseURL() + "/api/control_planes/" + kongCPID,
+		Address:    konnectConfig.Address + "/api/control_planes/" + kongCPID,
```
mflendrich, line 55: In the past I got bitten by kong.Client exploding when the kong URL provided by the user ends with a `/`.

Could we make sure that kong url and konnect URL are resilient to a trailing `/` provided by the user?

cmd/root.go
hbagdi: Because `konnect-email` and `konnect-password` are at root-level.
The question then comes up: why are these at root-level?
Because we want users to have a `~/.deck.yml` file that contains these credentials and not supply them on the CLI. So, because of viper working at the root-level only.

cmd/common_konnect.go
mflendrich: Filed https://github.com/Kong/deck/issues/390

mflendrich (approved):

PR #385, opened 2021-05-14
Refactor  workspaceExists

mmorel-35:


cmd/common.go
```diff
@@ -28,31 +28,27 @@ var (
 )
 
 // workspaceExists checks if workspace exists in Kong.
-func workspaceExists(ctx context.Context, config utils.KongClientConfig) (bool, error) {
-	if config.Workspace == "" {
+func workspaceExists(ctx context.Context, rootConfig utils.KongClientConfig, workspaceName string) (bool, error) {
```
mflendrich, line 31: To avoid bugs, we can make this function work on any `KongClientConfig`, not only one with workspace unset.

To do that:
- rename the argument `rootConfig` to `config`, and pass it by pointer (`*utils.KongClientConfig`)
- in the function, make the config a root config explicitly:
    ```go
    rootConfig := config.ForWorkspace("")
    ```

mmorel-35:
Hi @mflendrich , anything left do here ?

rainest:
Michal is out this week, I'm combing through some of the outstanding PRs.

@mmorel-35 this looks like it's probably fine. Can you provide a brief blurb about the purpose of the refactor? Not immediately clear to me from the code alone, and I'd like to have a more descriptive explanation for posterity.

mmorel-35:
It's using go-kong services to check the existence of a workspace instead of the previous logic of verification

PR #386, opened 2021-05-14
Do not destroy SNIs when updating unrelated certificate fields

### LINKED ISSUE #356: Lose snis when updating certificate
I have a certificate with a sni (*.example.xzy):
Screenshot from 2021-04-28 09-40-29

But it loses snis when I update certificate (for renewal):
Screenshot from 2021-04-28 09-44-09

I use the commands below to update certificate:
```bash
deck ping
deck -s kong/ validate
deck -s kong/ diff
deck -s kong/ sync
```

Here is my Kong Declarative configuration for certificate:
```yaml
certificates:
- id: ecd612c1-ec71-4311-85b3-96226ae79e6c
  cert: |-
    -----BEGIN CERTIFICATE-----
    #hidden
    -----END CERTIFICATE-----
    -----BEGIN CERTIFICATE-----
    #hidden
    -----END CERTIFICATE-----
  key: |-
    -----BEGIN PRIVATE KEY-----
    #hidden
    -----END PRIVATE KEY-----
  snis:
  - name: '*.example.xzy'
```

What I missed? Thanks in advance!

rainest:
The first commit here is a noisy chore. Review 6de9ee254abadc2c4b9fa0987f49844208b4239c for the meat of this change.

#356 occurs when updating a certificate without modifying its SNIs. We strip SNIs from certificate objects to handle SNIs separately, but the Kong admin API behavior for certificate PUTs handles an empty SNI set on certificates as an instruction to _delete_ those SNIs (probably because SNIs can't exist independent of certificates and because there's a strict one-to-one certificate-to-SNI relationship).

To work around this, this change copies the existing certificate's SNIs out of state during updates and sets them on the updated certificate object before sending it. This preserves the SNI list if it is unchanged while still updating SNIs if they do change (subsequent diffs on the SNIs themselves handle that).

Fix #356

### DIFF
--- a/diff/cert.go
+++ b/diff/cert.go
@@ -85,6 +85,26 @@ func (sc *Syncer) createUpdateCertificate(
 
 	// found, check if update needed
 	if !currentCertificate.EqualWithOpts(certificateCopy, false, true) {
+		// Certificate and SNI objects have a special relationship. A PUT request
+		// (which we use for updates) with a certificate that contains no SNI
+		// children will in fact delete any existing SNI objects associated with
+		// that certificate, rather than leaving them as-is.
+
+		// To work around this issues, we set SNIs on certificates here using the
+		// current certificate's SNI list. If there are changes to the SNIs,
+		// subsequent actions on the SNI objects will handle those.
+		currentSNIs, err := sc.currentState.SNIs.GetAllByCertID(*currentCertificate.ID)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error looking up current certificate SNIs %v",
+				certificate.Identifier())
+		}
+		sniNames := make([]*string, 0)
+		for _, s := range currentSNIs {
+			sniNames = append(sniNames, s.Name)
+		}
+
+		certificateCopy.SNIs = sniNames
+		currentCertificate.SNIs = sniNames
 		return &Event{
 			Op:     crud.Update,
 			Kind:   "certificate",


rainest:
[cert_test.txt](https://github.com/Kong/deck/files/6480899/cert_test.txt) shows some example syncs with the changes against [testcerts.tar.gz](https://github.com/Kong/deck/files/6480903/testcerts.tar.gz)

Note that there was one small change after the test run: originally, I only copied the current SNIs onto the new certificate, so this would spuriously report that it was adding SNIs even if they were unchanged (this was only a cosmetic issue with the diff--the PUT had effect it should). I later amended the commit to copy it onto both, since the current certificate in the update is only used in the diff, and doing so resolves that diff issue.

diff/cert.go
```diff
@@ -85,6 +85,18 @@ func (sc *Syncer) createUpdateCertificate(
 
 	// found, check if update needed
 	if !currentCertificate.EqualWithOpts(certificateCopy, false, true) {
+		currentSNIs, err := sc.currentState.SNIs.GetAllByCertID(*currentCertificate.ID)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error looking up current certificate SNIs %v",
+				certificate.Identifier())
```
shaneutt, line 91: Personally since wrapping became available in `fmt` in Go `v1.14` I would take this opportunity to use the standard library wrapping instead, to stop the growth of the 3rd party library which seems to be outmoded at this point:

```suggestion
			return nil, fmt.Errorf("error looking up current certificate SNIs %v: %w", certificate.Identifier(), err)
```

I don't consider this a blocker, but just for your considerations :thinking:

diff/cert.go
```diff
@@ -85,6 +85,18 @@ func (sc *Syncer) createUpdateCertificate(
...
+		}
+		sniNames := make([]*string, 0)
```
shaneutt, line 93:
```suggestion
		sniNames := make([]*string, 0, len(currentSNIs))
```
:point_up: just a micro-optimization: since we know the length this slice is going to end up needing to be anyway, this will heap alloc the right size array behind the slice instead of making the runtime do it dynamically.

shaneutt (approved):
Ultimately my read of this patch based on the description is that this will take care of it, I have no blockers (though please do see my comments below) however, due to the lack of testing for this code path I would appreciate if we could please write a small report about our manual testing of this patch before we merge this.

hbagdi:
In the spirit of atomic and smaller PRs, please send a different PR for the typo and remove that commit from this PR.

diff/cert.go
```diff
@@ -85,6 +85,18 @@ func (sc *Syncer) createUpdateCertificate(
 
 	// found, check if update needed
 	if !currentCertificate.EqualWithOpts(certificateCopy, false, true) {
+		currentSNIs, err := sc.currentState.SNIs.GetAllByCertID(*currentCertificate.ID)
```
hbagdi, line 88: This is a fairly special case. Do you mind copying the git commit body in here as a code comment to help future maintainers?

hbagdi (approved):
The special handling of SNIs in Kong continues to be a pain point in decK. Another lesson learned with API design.

hbagdi:
Please only merge the relevant patch in this PR. The typo fix should be its own PR.

Issue #232, opened 2020-11-13
[Feature] - Allow the ability to mute or not display plugin config when syncing


DeclanPaulKeane:
**Request Feature**

Can we add the ability to mute or not display plugin or general config when syncing changes to Kong.

**Current Behaviour**

When you make any change to the configuration of Kong, the change is outputted to standard out.

**Reproduction** 

1. Create a file like below. Filename "add-header.yaml":

`_format_version: "1.1"
services:
- connect_timeout: 60000
  host: mockbin.org
  name: Test
  path: /request
  port: 80
  protocol: http
  read_timeout: 60000
  retries: 5
  write_timeout: 60000
- connect_timeout: 60000
  host: httpbin.org
  name: test
  port: 80
  protocol: http
  read_timeout: 60000
  retries: 5
  write_timeout: 60000
  routes:
  - name: test
    paths:
    - /some/path
    path_handling: v0
    preserve_host: false
    protocols:
    - http
    - https
    regex_priority: 0
    strip_path: true
    https_redirect_status_code: 426
  plugins:
  - name: pre-function
    config:
      access: []
      body_filter: []
      certificate: []
      functions:
      - |-
        local original_path = ngx.var.request_uri
        ngx.req.set_header("Kong-Original-Path-1", original_path)
      header_filter: []
      log: []
      rewrite: []
    enabled: true
    protocols:
    - grpc
    - grpcs
    - http
    - https
`

2. deck sync -s add-header.yaml 
3. Make a change to the config. Change "Kong-Original-Path-1" to "Kong-Original-Path-2"
4. deck sync -s add-header.yaml 
5. Changes that are outputted to the screen:

`updating plugin pre-function for service 951c1f16-79f8-48d2-9186-5e7d3a53beb9  {
   "config": {
     "access": [
     ],
     "body_filter": [
     ],
     "certificate": [
     ],
     "functions": [
     "local original_path = ngx.var.request_uri
ngx.req.set_header("Kong-Original-Path-1", original_path)"
      "local original_path = ngx.var.request_uri
ngx.req.set_header("Kong-Original-Path-2", original_path)"
     ],
     "header_filter": [
     ],
     "log": [
     ],
     "rewrite": [
     ]
   },
 "created_at": 1.605227664e+09,
   "enabled": true,
   "id": "30095a96-5186-40a1-9751-63fb902503b7",
   "name": "pre-function",
   "protocols": [
     "grpc",
     "grpcs",
     "http",
     "https"
   ],
[...]

PR #393, opened 2021-05-18
Feat/silence print

### LINKED ISSUE #152: implemnt -s/--silent flag to silence all event logs

hbagdi:
Fixes #152 #232

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -1,0 +1,16 @@
+package cmd
+
+import (
+	"github.com/fatih/color"
+	"github.com/kong/deck/solver"
+)
+
+func printStats(stats solver.Stats) {
+	// do not use github.com/kong/deck/print because that package
+	// is used only for events logs
+	printFn := color.New(color.FgGreen, color.Bold).PrintfFunc()
+	printFn("Summary:\n")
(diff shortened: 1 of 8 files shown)


hbagdi:
Please review commit by commit. 
I'm looking for feedback on:
- flag naming
- how to lay out code (continued effort to de-duplicate code)

shaneutt (approved):

cmd/utils.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+func preRunSilenceEventsFlag() error {
+	printPkg.DisableOutput = true
+	return nil
+}
+
+func addSilenceEventsFlag(set *pflag.FlagSet) {
+	set.BoolVar(&silenceEvents, "disable printing events to stdout",
+		false, "silent events")
```
rainest, line 31: Is the arg order here correct? It looks like you have two usage strings and no name: https://pkg.go.dev/github.com/spf13/pflag#FlagSet.BoolVar

`silence-events` seems fine if this is only a partial silence. If it's actually everything, then `silence`.

hbagdi: Thanks for catching that, the last refactor gone wrong. Updated.

PR #394, opened 2021-05-18
feat: implement `file` template function

### LINKED ISSUE #91: feature: add support to read certificate private keys from file-system
Certificates are sensitive and can't be stored in plain-text.
In most use-cases, decK runs in a CI pipeline. In such a system, certificate private key and certificate can be provided to decK during runtime by CI via a file.
This avoids storing the certificate's private key in plaint-text in decK's state file.

hbagdi:
Fix #91

### DIFF
--- a/file/readfile.go
+++ b/file/readfile.go
@@ -153,9 +153,18 @@ func getPrefixedEnvVar(key string) (string, error) {
 	return value, nil
 }
 
+func getFileContent(path string) (string, error) {
+	content, err := ioutil.ReadFile(path)
+	if err != nil {
+		return "", fmt.Errorf("reading file: %v", err)
+	}
+	return string(content), nil
+}
+
 func renderTemplate(content string) (string, error) {
 	t := template.New("state").Funcs(template.FuncMap{
-		"env": getPrefixedEnvVar,
+		"env":  getPrefixedEnvVar,
+		"file": getFileContent,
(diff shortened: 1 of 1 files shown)


hbagdi:
@mflendrich @rainest How does this implementation look to you? I'm looking for some early feedback. I intend to add tests for this once this approach looks good.

mflendrich (commented):
I think that allowing arbitrary FS access is a big deal from the security perspective: I am not convinced that benefits outweigh the security opportunity cost, considering the following alternative solution:

You can achieve the equivalent result using #286 by reading the file into an environment variable

```bash
env "DECK_FILE_AAA=$(cat /my/file.pem)" "DECK_FILE_BBB=$(cat /my/other/file.pem)" deck sync ...
```

and then using the `env` substitution function. This should be enough for CI use.

The solution described above seems to solve the problem pointed out in #91, at the same time not permitting unbounded FS access for a user who controls the decK file.

That's a -1 from me for implementing of `file` template function, because of that security concern.

rainest:
Does loading them into environment variables change anything? The user running deck needs read permissions to the file whether they're cat-ing them into the environment before invoking deck or having deck load them in via the template.

While env loading can achieve the same, it could be a bit unwieldy with large numbers of certs.

hbagdi:
Holding off on this for now.
We will revisit this later.

PR #397, opened 2021-05-20
fix: use a document check that works

### LINKED ISSUE #396: deck sync results in a panic if all entities exist
state file:
```yaml
_format_version: "1.1"
services:
- connect_timeout: 60000
  host: httpbin.org
  name: svc1
  port: 80
  protocol: http
  read_timeout: 60000
  retries: 5
  routes:
  - https_redirect_status_code: 301
    name: r1
    path_handling: v0
    paths:
    - /r1
    preserve_host: false
    protocols:
    - http
    - https
    regex_priority: 0
    request_buffering: true
    response_buffering: true
    strip_path: false
    tags:
    - managed-by-deck
    - org-unit-42
  tags:
  - team-svc1
  - managed-by-deck
  - org-unit-42
  write_timeout: 60000
```

Steps to reproduce:
```shell
➜  deck git:(main) ✗ go run main.go sync  --kong-addr http://localhost:7001 --headers 'kong-admin-token:CHANGE_ME'                                                                                                 
creating service svc1                                                                                    
creating route r1                                                                                                                                                                                                  
Summary:                                                                                                 
  Created: 2                                        
  Updated: 0                                                                                                                                                                                                       
  Deleted: 0                                                                                             
➜  deck git:(main) ✗ go run main.go sync  --kong-addr http://localhost:7001 --headers 'kong-admin-token:CHANGE_ME'                                                                                                 
panic: interface conversion: interface {} is *state.Service, not *state.Document                                                                                                                                   
                                                                                                                                                                                                                   
goroutine 135 [running]:                                                                                                                                                                                           
github.com/kong/deck/solver.Solve.func2(0xa3684f, 0x6, 0xa37255, 0x7, 0x9f8f20, 0xc00027b380, 0x9f8f20, 0xc00027b440, 0x1fa6ff510040d1aa, 0x60a6a301, ...)                                                         
        /home/hbagdi/deck/solver/solver.go:53 +0xb7f                                   
github.com/kong/deck/diff.(*Syncer).handleEvent.func1(0xc000697960, 0xc000697960)                                                                                                                                  
        /home/hbagdi/deck/diff/diff.go:472 +0xaf                                                         
github.com/cenkalti/backoff/v4.RetryNotifyWithTimer(0xc000274e28, 0xaee688, 0xc000697960, 0x0, 0xaf1c30, 0xc000232230, 0x0, 0x0)                                                                                   
        /home/hbagdi/go/pkg/mod/github.com/cenkalti/backoff/v4@v4.1.0/retry.go:55 +0x102                 
github.com/cenkalti/backoff/v4.RetryNotify(...)                                                          
        /home/hbagdi/go/pkg/mod/github.com/cenkalti/backoff/v4@v4.1.0/retry.go:34                        
github.com/cenkalti/backoff/v4.Retry(...)                                                                
        /home/hbagdi/go/pkg/mod/github.com/cenkalti/backoff/v4@v4.1.0/retry.go:28                        
github.com/kong/deck/diff.(*Syncer).handleEvent(0xc00023f040, 0xc000697740, 0xa3684f, 0x6, 0xa37255, 0x7, 0x9f8f20, 0xc00027b380, 0x9f8f20, 0xc00027b440, ...)                                                     
        /home/hbagdi/deck/diff/diff.go:471 +0xfa                                                         
github.com/kong/deck/diff.(*Syncer).eventLoop(0xc00023f040, 0xc000697740, 0x0, 0x0)                      
        /home/hbagdi/deck/diff/diff.go:461 +0x165                                                                                                                                                                  
github.com/kong/deck/diff.(*Syncer).Run.func1(0xc00023f040, 0xc000697740, 0xc000528330)
        /home/hbagdi/deck/diff/diff.go:398 +0x39    
created by github.com/kong/deck/diff.(*Syncer).Run                                                                                                                                                                 
        /home/hbagdi/deck/diff/diff.go:397 +0x185                                                        
```

rainest:
reflect.Type.Name() has limitations:

Name returns the type's name within its package for a defined type.
For other (non-defined) types it returns the empty string.

Not clear exactly why these types /aren't/ considered defined, but using
that did return empty string results, at least sometimes. The
convertible check should be more stringent, since that can only succeed
if the subsequent conversion will also.

Fix #396


solver/solver.go
```diff
@@ -49,7 +49,7 @@ func Solve(ctx context.Context, syncer *diff.Syncer,
 			print.CreatePrintln("creating", e.Kind, c.Console())
 		case crud.Update:
 			var diffString string
-			if reflect.TypeOf(e.OldObj).Name() == reflect.TypeOf(&state.Document{}).Name() {
+			if reflect.TypeOf(e.OldObj).ConvertibleTo(reflect.TypeOf(&state.Document{})) {
```
hbagdi, line 52:
```patch
-                       if reflect.TypeOf(e.OldObj).Name() == reflect.TypeOf(&state.Document{}).Name() {
+                       // TODO handle nil?
+                       oldName := reflect.Indirect(reflect.ValueOf(e.OldObj)).Type().Name()
+                       documentTypeName := reflect.TypeOf(state.Document{}).Name()
+                       if oldName == documentTypeName {

```
I think you need to compare types of values and not pointers to values.

hbagdi: Let's unit test this. Please move that comparison to a func isDocument(oldobj interface{}) bool.
Documentation says that reflect.Type are comparable using ==. Let's use that for equality checking.

(copied from internal slack)

solver/utils_test.go
```diff
@@ -145,3 +145,36 @@ bar
...
+			args: args{
+				obj: &state.Document{},
+			},
+			want: true,
+		},
+		{
+			name: "not document",
+			args: args{
+				obj: &state.Service{},
+			},
```
hbagdi, line 169: Can we check file.Document or FDocument as well?

rainest: cb80ca8

hbagdi (approved):
Let's wait for a review from @mflendrich on this one.

solver/utils.go
```diff
@@ -95,3 +96,9 @@ func eventFromArg(arg crud.Arg) diff.Event {
 	}
 	return event
 }
+
+func isDocument(obj interface{}) bool {
+	a := reflect.TypeOf(obj)
+	b := reflect.TypeOf(&state.Document{})
+	return a == b
```
mflendrich, line 103: Would simple type assertion not work?

```go
func isDocument(obj interface{}) bool {
	_, ok := obj.(*state.Document)
	return ok
}
```

mflendrich (changes requested):
I may be gravely mistaken about requesting changes because I don't know the context, but if what we're trying to do is nothing more than "test if object X is of type `struct Y`", then a type assertion should be the most robust way to do exactly that.

solver/utils.go
mflendrich: Or we could even do this inline in the `if` which calls `isDocument` today:

```go
      if oldObj, ok := e.OldObj.(*state.Document); ok {
        diffString, err = getDocumentDiff(oldObj, e.Obj.(*state.Document))
      } else {
        diffString, err = getDiff(e.OldObj, e.Obj)
      }
```

without the need to define a function and write a unit test.

rainest: It doesn't need a function; we broke it out specifically because we [wanted to unit test that part of the code](https://github.com/Kong/deck/pull/397#discussion_r636492718), and the calling function lacks its own test.

mflendrich: The problem we're solving is *exactly* an instance of the pattern below:

```
if x is of type A
  make a type assertion obvious from the condition above, and do something with x
else
  do something else
```

By hiding "x is of type A" we're making the operation more complex:
```
if (x is a document, whatever this means)
  make an unchecked type assertion not obviously valid from the
    condition above (without looking into the implementation details of the function), 
    and do something with x
else
  do something else
```

Also we have a unit test that verifies that:
1. a type assertion in Go works as intended
2. that the abstraction implements a type assertion

There is no point in testing (1). We don't test the plus operator by checking that `2+2 == 4`; we don't test the `if` expression by checking that if the condition is false, the block won't get executed. Why would we check that a type assertion works as expected?

Testing (2) would make sense if the abstraction provided value. I think it does not. It introduces some concept of "an untyped variable being _a document_", and leads to a type assertion being made in the conditional block that is not checked for validity locally. This will make the life of a person debugging this in the future harder.

mflendrich (commented):
See my response to the comment thread. The comment is no longer blocking, but I **strongly recommend** that the type assertion check happen immediately before the conversion.

That means less code, more readable logic, and no need for the test to exist. And less code is better in general.

solver/utils.go
mflendrich: Also, with the current implementation, the type assertion happens twice (first in `isDocument` and then in the conditional block). After the proposed change gets made, it will happen only once (in the condition).

solver/utils_test.go
```diff
@@ -3,6 +3,7 @@ package solver
 import (
 	"testing"
 
+	"github.com/kong/deck/file"
```
mflendrich, line 6: leftover import?

mflendrich (approved):

PR #400, opened 2021-05-21
Enable more linters to track and improve health

hbagdi:


hbagdi:
This is not meant to be an exhaustive list.
There are a bunch of other linters that I wish to enable, doing it gradually to make reviews easily.
Please review commit by commit.

The linter list is not sorted alphabetically currently. That will happen separately at a later time as well.

rainest (dismissed):
The heck does stylecheck actually do? https://golangci-lint.run/usage/linters/ doesn't really say much, and neither does the linked project.

rainest:
I'm against gofumpt:
- gofmt is standard and all Go dev environments have it+typically run it, so lower friction for community contributions.
- It looks like popular/well-tested rules from it eventually percolate into gofmt anyway.

Not sure if the same can be said for stylecheck--still not clear exactly how it differs from golint.

konnect/document_service.go
```diff
@@ -136,7 +134,7 @@ func (d *DocumentService) listAllByPath(ctx context.Context, path string) ([]*Do
 // ListAllForParent fetches all Documents in Konnect for a parent entity.
 func (d *DocumentService) ListAllForParent(ctx context.Context, parent ParentInfoer) ([]*Document, error) {
 	if parent == nil {
-		return nil, fmt.Errorf("Parent cannot be nil")
+		return nil, fmt.Errorf("rarent cannot be nil")
```
mmorel-35, line 137: Typo? 
```suggestion
		return nil, fmt.Errorf("parent cannot be nil")
```

.golangci.yml
```diff
@@ -1,16 +1,30 @@
 linters:
   enable:
+  - wastedassign
```
mmorel-35, line 3: Would it be more readable to sort alphabetically the linters?

mflendrich (approved):

konnect/document_service.go
mflendrich: Taken care of in https://github.com/Kong/deck/pull/402

.golangci.yml
mflendrich: Taken care of in https://github.com/Kong/deck/pull/402

mflendrich:
> I'm against gofumpt:
> 
>     * gofmt is standard and all Go dev environments have it+typically run it, so lower friction for community contributions.

Many of the checks performed by this linter would actually bubble up in the actual review. So having the static check return an issue rather than the reviewer doing that and requiring extra review churn is a reduction of friction to me. We could also have less stringent rules, but to me these don't seem unreasonable.

> 
>     * It looks like popular/well-tested rules from it eventually percolate into gofmt anyway.

It's more opinionated than gofmt.

PR #404, opened 2021-05-24
chore: bump up linter, more linters, conditionally skip bodyclose

hbagdi:
The bodyclose linter is going off incorrectly. The body is actually
closed before returning so we are disabling the linter.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,17 +1,18 @@
 linters:
   enable:
   - asciicheck
-  - bodyclose
   - depguard
   - dogsled
   - durationcheck
   - exhaustive
   - exportloopref
   - gci
+  - goconst
   - gofmt
   - gofumpt
   - goimports
   - gomnd
+  - gomodguard
   - gosec
   - govet
   - importas
--- a/konnect/control_plane_relations_service.go
+++ b/konnect/control_plane_relations_service.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/http"
 )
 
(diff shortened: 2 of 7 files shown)


mflendrich (commented):
I've read the sources of the bodyclose linter and it seems to expect that if a function returns a `http.Response` then the caller closes it.

This is not compatible with how `konnect.Client`'s `Do()` works: it makes a http call under the hood, then closes the response body and returns the `http.Response` to the caller.

To me, that feature of `konnect.Client` as well as `kong.Client` can be considered a leaky abstraction, but it feels completely impractical to try to do something about it now.

My recommendation would be to:
- remove the `bodyclose` check altogether (because it won't work in this codebase due to how `http.Response` gets passed outside of its context)
- create a code health issue to replace the returned `http.Response` with a struct with information of interest obtained from the response

Makes sense?

hbagdi:
I didn't open an issue because I'm conlicted if http.Response is a leaky abstraction or not and I can live without this liner. Is that okay?

mflendrich (approved):
> I'm conlicted if http.Response is a leaky abstraction

What I mean is the following:

`kong.Client` as well as `konnect.Client` are (similar) abstractions. Both hide an API behind their interface. Both use HTTP as the transport.
Both take some inputs and, internally, bake an HTTP request, and get a response.

The "leaky abstraction" part IMO is that kong.Client on one hand abstracts away creation of the http request (thus making HTTP an implementation detail of some kind), and on the other hand expose parts of that implementation detail (the `http.Response` type being an example of that).

PR #413, opened 2021-05-26
refactor: use go's embedding feature for embedding json schema

hbagdi:


file/schema.go
```diff
@@ -1,1507 +1,6 @@
...
-      },
-      "additionalProperties": false,
-      "type": "object"
-    }
-  }
-}`
+import _ "embed" // for embedding only
+
+//go:embed kong_json_schema.json
+var kongJSONSchema []byte
```
mmorel-35, line 6: You can also embed directly as a string, see https://golang.org/pkg/embed. Out of curiosity, why the array of bytes?

mflendrich (approved):
@jakubdyszkiewicz MBOI for you

LGTM

file/schema.go
hbagdi: Done!

PR #414, opened 2021-05-26
fix: drop conditional tagging

hbagdi:
When decK v1.7.0 or later is used with select-tag feature and
credentials in state file, decK will error out in most sync runs because
it can't locate the credentials in the current state.
This happens because #282 changed the conditions under which credentials
were exported: only tagged credentials are now exported.

On the other hand, `kongVersion` field in the state builder is not set.
This leads to credentials not being tagged appropriately when they are
read from the state files.

This patch unconditionally tags all credentials and ensures that
credentials are created and updated appropriately.

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -6,7 +6,6 @@ import (
 	"net/http"
 	"os"
 
-	"github.com/blang/semver/v4"
 	"github.com/kong/deck/cprint"
 	"github.com/kong/deck/diff"
 	"github.com/kong/deck/dump"
@@ -83,10 +82,6 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 	if err != nil {
 		return fmt.Errorf("reading Kong version: %w", err)
 	}
-	parsedKongVersion, err := parseKongVersion(kongVersion)
-	if err != nil {
-		return fmt.Errorf("parsing Kong version: %w", err)
-	}
 
 	// TODO: instead of guessing the cobra command here, move the sendAnalytics
 	// call to the RunE function. That is not trivial because it requires the
@@ -137,7 +132,6 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 	// read the target state
 	rawState, err := file.Get(targetContent, file.RenderConfig{
 		CurrentState: currentState,
-		KongVersion:  parsedKongVersion,
 	})
 	if err != nil {
 		return err
@@ -229,14 +223,6 @@ func fetchKongVersion(ctx context.Context, config utils.KongClientConfig) (strin
 		version = root["version"].(string)
 	}
 	return version, nil
-}
-
-func parseKongVersion(version string) (semver.Version, error) {
-	v, err := utils.CleanKongVersion(version)
-	if err != nil {
-		return semver.Version{}, err
-	}
-	return semver.ParseTolerant(v)
 }
 
 func validateNoArgs(cmd *cobra.Command, args []string) error {
--- a/file/builder.go
+++ b/file/builder.go
@@ -3,7 +3,6 @@ package file
 import (
 	"fmt"
 
-	"github.com/blang/semver/v4"
 	"github.com/kong/deck/konnect"
 	"github.com/kong/deck/state"
 	"github.com/kong/deck/utils"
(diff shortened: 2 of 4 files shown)


mflendrich (approved):

hbagdi:
This fixes an unreported bug in decK now. @rainest could you review my logical thinking here (and only then make sure that the code is updated accordingly)?

rainest:
This drops Kong version checking entirely. However, we still have an outstanding credential type without tags, because its history was weird: https://github.com/Kong/deck/blob/c77fa4d56b8b361e361b2d7c9d64b4a49baf48e2/file/builder.go#L391-L406

Per https://docs.konghq.com/enterprise/2.4.x/support-policy/#version-support-for-kong-gateway-enterprise we are covered to drop support for <=1.5 in newer decK versions, but the EOL date for 2.1 is still a ways off. We'd need to preserve the deck version awareness to properly fix that.

The second thing I'm unclear on is how this would fix the bug described/where the bug appears. The change here is to just remove the conditional version behavior, correct? It seems like if this bug affects versions that do support credential tags that we aren't actually changing their behavior: they'd have already passed the version gate and have already run `MustMergeTags(&cred, b.selectTags)`.

mmorel-35:
Is there still any use for github.com/blang/semver/v4 in go.mod after this?

hbagdi:
> The second thing I'm unclear on is how this would fix the bug described/where the bug appears. The change here is to just remove the conditional version behavior, correct? It seems like if this bug affects versions that do support credential tags that we aren't actually changing their behavior: they'd have already passed the version gate and have already run MustMergeTags(&cred, b.selectTags).

The bug comes into play because we no longer set `b.kongVersion`. Since no version is set, `MustMergeTags()` is not executed which results in plugins not loading correctly.

The mtls-auth credentials should remain unaffected here since we don't do any sort of merging or de-duplication for them. No?

rainest:
> The bug comes into play because we no longer set b.kongVersion

Alright, that makes sense then.

> The mtls-auth credentials should remain unaffected here since we don't do any sort of merging or de-duplication for them. No?

Well, we don't currently, but we should: we didn't enable the commented `if b.kongVersion.GTE(kong220Version)` clause to tag it because the fix wasn't actually ready yet when it was added. It is available now. However, we can't make use of it unless we keep the version checks or drop support for 2.1 and older instead of 1.4 and older.

hbagdi:
> Well, we don't currently, but we should: we didn't enable the commented if b.kongVersion.GTE(kong220Version) clause to tag it because the fix wasn't actually ready yet when it was added. It is available now. However, we can't make use of it unless we keep the version checks or drop support for 2.1 and older instead of 1.4 and older.

Thinking out loud:
Would enabling that break existing users? If a user migrates from 1.5 to 2.1, decK will tag these.
To further complicate things, we are exporting credentials only when they are tagged: https://github.com/Kong/deck/blob/aec7c176eecff877972fdd6e72dcac499f8beb18/dump/dump.go#L128.
I _think_ the tags are ignored when an entity is not tagged.

rainest:
Revisiting this again:

> The bug comes into play because we no longer set b.kongVersion

But this is the PR [that removes this](https://github.com/Kong/deck/pull/414/files#diff-2b0546a527b3560f4e7f8501ef7907999b2e1a7f2498a60719630310f7778956L86-L141), correct? It's fixing a bug that it introduces, but it's not clear to me what reason we have to drop version parsing originally.

> Would enabling [mtls-auth credential tag handling] break existing users?

If they're on a version that doesn't support it and we use the pattern introduced here (unconditionally tagging credentials), then yes. If we keep version info/ability to gate tagging, we can add it without issues.

As noted, the complication is the use of tag filters [in the credential dump as of #282](https://github.com/Kong/deck/blame/v1.7.0/dump/dump.go#L656). If we wanted to gate mtls-auth tagging to supported versions, we'd need to expose version info to `GetAllMTLSAuths()` and gate its filter also: supported versions do set the tag option, older versions do not (they still use the unconditional dump we had prior to 1.7).

hbagdi:
> But this is the PR that removes this, correct? It's fixing a bug that it introduces, but it's not clear to me what reason we have to drop version parsing originally.

@rainest No. It was not being set at the caller location.


This has all been a bit confusing. I took some time to walk through the possible scenarios and came to the solution in #455.
Please verify my assumptions, approach and solution in there.
I'm going to close this PR.

PR #416, opened 2021-05-26
feat/completion

hbagdi:


hbagdi:
This builds on top of #415. Will need a rebase once that's merged in. Please review only the last commit.

cmd/completion.go
```diff
@@ -0,0 +1,71 @@
...
+PowerShell:
+
+  PS> deck completion powershell | Out-String | Invoke-Expression
+
+  # To load completions for every new session, run:
+  PS> deck completion powershell > deck.ps1
+  # and source this file from your PowerShell profile.
+`,
+	DisableFlagsInUseLine: true,
+	ValidArgs:             []string{"bash", "zsh", "fish", "powershell"},
```
mflendrich, line 52: optional:

The list of valid choices is defined twice: once in `ValidArgs`, then in the `switch`.

Suggestion to avoid duplication: remove `ValidArgs`, replace `cobra.ExactValidArgs` with `cobra.ExactArgs` and add a `default` case to the `switch`.

mflendrich (approved):
LGTM but see the optional comment.

cmd/completion.go
hbagdi: Good call. Incorporated this change!

PR #418, opened 2021-05-26
tests: add test for cprint package

hbagdi:


cprint/color_test.go
```diff
@@ -0,0 +1,105 @@
...
+	go func() {
+		var buf bytes.Buffer
+		defer wg.Done()
+		io.Copy(&buf, reader)
+		out <- buf.String()
+	}()
+	defer wg.Wait()
+	f()
+	writer.Close()
+	return <-out
```
mflendrich, line 33: Won't this work:

```go
var out bytes.Buffer
color.Output = out
f()
return out.String()
```

?

hbagdi: Works!
It is amazing how my tunnel vision led me to this ugly monster.

cprint/color_test.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"bytes"
+	"os"
+	"testing"
+
+	"github.com/fatih/color"
+	"github.com/stretchr/testify/assert"
+)
+
+func captureOutput(f func()) string {
```
mflendrich, line 12: Sorry about bugging you again - forgot about this. We should probably revert the prior setting of `color.Output` and `defer` setting it back.

mflendrich (approved):
LGTM but see the comment I didn't think of during the first round of review.

cprint/color_test.go
hbagdi: Updated!

PR #424, opened 2021-05-27
tests: add tests and refactor utils package

hbagdi:


utils/uuid.go
```diff
@@ -0,0 +1,37 @@
+package utils
+
+import (
+	"crypto/rand"
+	"encoding/hex"
+)
+
+// UUID will generate a random v14 unique identifier based upon random numbers
+// nolint:gomnd
+func UUID() string {
```
mmorel-35, line 10: Why was `github.com/google/uuid` not used for uuid generation by the way?

hbagdi: I think it was primarily to avoid an additional dependency.

mmorel-35: This implementation could be simply replaced by `uuid.New().String()`, that might worth a new PR

utils/prompt.go
```diff
@@ -0,0 +1,46 @@
...
+	"os"
+	"strings"
+)
+
+// confirm prompts a user for a confirmation with message
+// and returns true with no error if input is "yes" or "y" (case-insensitive),
+// otherwise false.
+func Confirm(message string) (bool, error) {
+	fmt.Print(message)
+	yes := []string{"yes", "y"}
```
shaneutt, line 15: nit: I would prefer to call something like this `validOptions`, and then below instead of `for_, valid := range yes` I think it could be a little easier to read `for _, validOption := range validOptions`? Let me know what you think, not a blocker.
```suggestion
	validOptions := []string{"yes", "y"}
```

utils/prompt.go
```diff
@@ -0,0 +1,46 @@
...
+	input = strings.ToLower(input)
+	for _, valid := range yes {
+		if input == valid {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func ConfirmFileOverwrite(filename string, ext string, assumeYes bool) (bool, error) {
```
shaneutt, line 30: lint: public symbol needs godoc comment

```suggestion
// ConfirmFileOverwrite is a helper function to determine whether or not the program should
// truncate and overwrite a file given its name and extension. If the file doesn't already exist
// in the filesystem, then this will return true, otherwise it will prompt the user for confirmation.
func ConfirmFileOverwrite(filename string, ext string, assumeYes bool) (bool, error) {
```

utils/prompt.go
```diff
@@ -0,0 +1,46 @@
...
+	if assumeYes {
+		return true, nil
+	}
+
+	filename = AddExtToFilename(filename, ext)
```
shaneutt, line 35: q: just wondering why we don't just take `filename` and remove the `ext` var here, is there some context I'm missing where the extension is important?

utils/zero_test.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	t.Parallel()
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			ZeroOutTimestamps(tt.args.obj)
+			assert.Equal(t, tt.expectedObj, tt.args.obj)
+		})
+	}
+}
```
shaneutt, line 136: tests++

shaneutt (approved):
I have no blockers, but please consider my comments and suggestions before you merge. :+1:

utils/uuid.go
hbagdi: @mmorel-35 A follow up PR after this to change that is welcome.

utils/prompt.go
```diff
@@ -0,0 +1,46 @@
+package utils
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+)
+
+// confirm prompts a user for a confirmation with message
```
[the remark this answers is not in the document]
hbagdi: Surprised why linter didn't catch this.

utils/prompt.go
hbagdi: The reason is 'I wrote bad code' and hence this is the way it is. It will go away in a future refactor.
This has weird implications otherwise.

PR #425, opened 2021-05-28
Feature/cache modules

mmorel-35:


hbagdi:
The CI says cache hit occured in the first run. That seems odd, any idea whats going on?

mmorel-35:
which run ? 
image

hbagdi:
The post run of that same build.
Actually, there is a bug here. You want to populate cache _after_ you check out the repo. There is no go.sum file when you populate the cache here.

mflendrich (commented):
As Harry said above :point_up_2:, please fix the job order and re-request review

mmorel-35:
@mflendrich, that's what I just did with 6ee7fd5

mflendrich (approved):
LGTM. Seems that I reviewed an old version when I was making my previous comment.

mmorel-35:
This incompatible with Golangci

PR #431, opened 2021-06-04
fix: document creation and deletion race with parent objects

hbagdi:
Document object in Konnect are always associated with either a service
package or a service version.
- If a service version or package is deleted, deleting a document
results in a 404, an error in decK
- If a document is created before the corresponding service package or
version is created, it resulted in 400, again an error

This patch adds the appropriate barriers to prevent such races.

### DIFF
--- a/diff/diff.go
+++ b/diff/diff.go
@@ -251,6 +251,10 @@ func (sc *Syncer) delete() error {
 		return err
 	}
 
+	// barrier for foreign relations
+	// Documents must be deleted before ServiceVersions and Service packages
+	sc.wait()
+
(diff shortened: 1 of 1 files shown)


hbagdi:
ping @rainest specifically

rainest (approved):

hbagdi:
holding so that #435 and #436 get in. 
merging this in before them will result in a lot of unnecessary merge conflict pain

rainest:
#435 and #436 are now in; anything other than the conflict needed before proceeding with this?

hbagdi:
No. Rebased and resolved the conflict.

PR #435, opened 2021-06-08
Refactor/consolidation

hbagdi:
This PR refactors the entity-specific diff, sync, post-process logic into `types` package.
This patch is part of a larger set where we are consolidating the two
tightly coupled `diff` and `solver` packages into a single `diff`
package. As part of the refactor, we are also moving out entity-specific
logic into a separate package that will be injected into diff package as
needed. The intention is to make adding new entities into decK easier
and more maintainable and make the core engine of decK testable and test
it.

The refactor seems large but largely involves moving files around with some minor consistent editing.
One commit will need to be squash before merging. It is laid out the way it is to make reviewing easy.

Because this is a large PR, I'm seeking a review for correctness rather than stylistic comments as they can get harder to incorporate.
Please review commit-by-commit.


diff/solver.go
```diff
@@ -0,0 +1,106 @@
...
+		return result, nil
+	})
+	return stats, errs
+}
+
+func buildRegistry(client *kong.Client, konnectClient *konnect.Client) *crud.Registry {
+	opts := types.EntityOpts{
+		KongClient:    client,
+		KonnectClient: konnectClient,
+	}
```
hbagdi, line 95: This duplicate opts and call to NewEntity below will be removed in a following PR. I didn't include it in this PR because that resulted in another larger refactor and also touched cmd package heavily.

types/core.go
```diff
@@ -0,0 +1,335 @@
...
+func (e entityImpl) PostProcessActions() crud.Actions {
+	return e.postProcessActions
+}
+
+type EntityOpts struct {
+	CurrentState  *state.KongState
+	TargetState   *state.KongState
+	KongClient    *kong.Client
+	KonnectClient *konnect.Client
+}
```
rainest, line 49: This seems like it's named oddly: the state and client objects aren't _options_, correct? The comment indicating that the instances of this type will be de-duplicated in the future and the contents of the struct suggests that this is more connection config and states for the differ, not the individual entities it handles.

types/core.go
```diff
@@ -0,0 +1,335 @@
...
+
+	ServicePackage, ServiceVersion, Document,
+}
+
+func NewEntity(t string, opts EntityOpts) (Entity, error) {
+	switch t {
+	case Service:
+		return entityImpl{
+			typ: "service",
+			cRUDActions: &serviceCRUD{
```
rainest, line 125: Minor readability thing, but case should be consistent across the acronym, i.e. `crudActions` rather than `cRUDActions`. If the first letter needs to be lowercase to not expose it, the rest should be also. "CRUD" is a common-enough acronym that it's recognizable as such even when not capitalized, thankfully.

rainest (commented):
Only minor nomenclature comments so far.

I think this is definitely a step in the right direction, knowing that decK currently has a lot of same-ish functions across the various entities it works with and that defining interfaces for those will allow us to test important functionality without re-testing each implementation, as we'd need to with the current interface-less architecture.

Thanks for the initial commit that changes this for just Services before doing the same again for the rest of the entities--it helps me to see how the change works in isolation.

I'm still struggling to grasp the refactor entirely without tests for the new interface--I think seeing those will help me wrap my brain around it better. Without those or a walkthrough of the design process my brain is saying "this is good, but I can't say for certain if there's some issue with the design or a way it could be improved further". Do you think it's possible to add those partway through, or will those be hard to write until after additional refactoring? Wasn't sure if those were waiting on some additional changes beyond #436.

types/core.go
```diff
@@ -0,0 +1,335 @@
...
+	"github.com/kong/go-kong/kong"
+)
+
+func eventFromArg(arg crud.Arg) crud.Event {
+	event, ok := arg.(crud.Event)
+	if !ok {
+		panic("unexpected type, expected diff.Event")
+	}
+	return event
+}
```
mflendrich, line 18: this is a constructor for a type in package `crud` that takes an argument which lives in package `crud` as well.

Probably makes more sense in that package, no?

types/core.go
```diff
@@ -0,0 +1,335 @@
...
+
+type EntityOpts struct {
+	CurrentState  *state.KongState
+	TargetState   *state.KongState
+	KongClient    *kong.Client
+	KonnectClient *konnect.Client
+}
+
+const (
```
mflendrich, line 51: Entity types feel like a good case for an enumeration type to me. That might be better than an opaque `string`.

Suggestion: `type EntityType string` and define these consts below as `EntityType`.

types/core.go
```diff
@@ -0,0 +1,335 @@
...
+	RBACRole, RBACEndpointPermission,
+
+	ServicePackage, ServiceVersion, Document,
+}
+
+func NewEntity(t string, opts EntityOpts) (Entity, error) {
+	switch t {
+	case Service:
+		return entityImpl{
+			typ: "service",
```
mflendrich, line 124: Why use literals and not consts for the value of `typ`? (DRY)

types/core.go
mflendrich: +1 to that, there's an explicit recommendation in the canon:

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

types/core.go
hbagdi: The type used to live in the diff package and was refactored into the crud package as part of this refactor. What you suggest makes sense, will do.

types/core.go
hbagdi: `XxOpts` is a way to name a configuration struct to an abstraction, that's all. I can rename to `XxConfig` if that improves readability.

types/core.go
hbagdi: Did it here already: https://github.com/Kong/deck/pull/436/commits/3bcc059d5b99fc5897c9e42d1153688f0fe5caa9

hbagdi: Pulling it up here will result in a lot of conflict. Is it okay to keep the order the way it is right now?

types/core.go
hbagdi: Suggestion accepted. Is it okay if I do it in a follow up refactor after this and #436 PRs merge in? Tracking in https://github.com/Kong/deck/issues/440.

hbagdi:
@mflendrich @rainest Thanks for the review on this large diff. I've pushed one commit to take care of the reviews and added comments as responses where needed. Please take another look when you can.

types/core.go
rainest: Nah, if that's a standard naming pattern I should just get used to it then; disregard.

rainest (approved):

hbagdi:
ping @mflendrich for a review. Would like to get this in as fast as possible to avoid conflicts.

mflendrich (approved):

PR #436, opened 2021-06-08
Refactor/consolidate event gen

hbagdi:
This PR builds on top of #435.
It goes a step ahead and moves a lot more entity specific logic into the types package to make it easy to test existing code and add new entities in future.

Please review the last 4 commits only. The rest of them are part of #435.

Why two separate PRs? To make code reviewing a bit friendly. This PR will be rebased once #435 merges in.


mflendrich:
A small usability tip/request for future PRs building on top of other PRs:

For PRs where there's a custom diff to be reviewed, consider adding a specific "commit compare" link like this to make life easier for reviewers: https://github.com/Kong/deck/compare/refactor/consolidation...refactor/consolidate-event-gen

Also the "Please review the last 4 commits only" comment ages badly in case new commits get added to a PR :joy:

Some other CR systems have that built in. Too bad GitHub does not.

types/basicauth.go
```diff
@@ -0,0 +1,199 @@
...
+	return &state.BasicAuth{BasicAuth: *updatedBasicAuth}, nil
+}
+
+type basicAuthDiffer struct {
+	kind crud.Kind
+
+	currentState, targetState *state.KongState
+}
+
+func (d *basicAuthDiffer) warnBasicAuth() {
```
mflendrich, line 98: This function does not do anything (and has a TODO therein), so calls to it may be misleading to the reader.

Suggestion:
- delete this function altogether
- remove the calls to this function
- create a GH issue that this is a regression
- in the places of the removed calls to `warnBasicAuth`, put TODOs with a link to the GH issue

mflendrich (commented):
Definitely a change in the good direction.

I don't feel confident about stability after this change, because the size of this diff (even diff-based) makes it challenging to review each bit of the logic carefully without repetition blindness, and testing is limited here.

Added one comment for the TODO about the warning regression, otherwise the diff looks good (with the reservation expressed above) :slightly_smiling_face:

hbagdi:
> I don't feel confident about stability after this change, because the size of this diff (even diff-based) makes it challenging to review each bit of the logic carefully without repetition blindness, and testing is limited here.

Agree 100%. I did some manual testing to ensure some basic workflows but I share the concerns.
#63 is making progress which will help us gain confidence in future.

types/basicauth.go
hbagdi: Fixed the feature in 1c5df1184a94653bf745202558028904112159b1.

mflendrich (approved):

PR #437, opened 2021-06-08
refactor: remove unused reset package

hbagdi:
PR #409 simplified code to use existing diff engine to power the reset
functionality instead of having another untested reset funtion.
This PR simply drops the package from the source tree.


hbagdi:
The package was not left out on purpose but rather was an oversight. This PR fixes that.

ccfishk (approved):
LGTM

PR #441, opened 2021-06-10
prevent user overwrite default configuration

### LINKED ISSUE #421: restrict some fields from being set in default values
#419 implements defaulting logic in decK.
A user can set all properties of Service/Route/Target/Upstream but not all of them actually make sense.
The goal of this issue is to add validation and error out if the following fields are specified in data structure KongDefaults (Under Info of Content):
- service.id
- service.host
- service.name
- service.port
- route.name
- route.id
- target.target4
- target.id
- upstream.id
- upstream.name

ccfishk:
prevent user overwrite some default configuration.

fixes #421


file/validate.go
```diff
@@ -1,20 +1,70 @@
...
 	"fmt"
 
 	ghodss "github.com/ghodss/yaml"
 	"github.com/kong/deck/utils"
 	"github.com/xeipuuv/gojsonschema"
 )
 
+func check(item string, t *string) bool {
+	if t != nil || len(*t) > 0 {
+		fmt.Println(item + "[err] is not allowed to be specified.")
```
rainest, line 14: I think we'd want to instead aggregate the unacceptable fields `checkDefaults()` sees and return an error if we see more than 0. Something like:

```
func checkDefaults(kd KongDefaults) error {
    var invalid string[]
    if kd.Service.ID != nil || len(*kd.Service.ID) > 0 {
        invalid = append(invalid, "Service.ID")
    }
    ...
    if len(invalid) > 0 {
        return fmt.Errorf("unacceptable fields in defaults: %s", strings.Join(invalid, ", "))
    }
    return nil
}
```

rainest (changes requested):

file/validate.go
```diff
@@ -29,5 +82,16 @@ func validate(content []byte) error {
 		err := fmt.Errorf(desc.String())
 		errs.Errors = append(errs.Errors, err)
 	}
+
+	var kongdefaults *KongDefaults
+	if err := json.Unmarshal(content, kongdefaults); err != nil {
```
hbagdi, line 87: This won't work. The code should unmarshal this into [Content](https://github.com/Kong/deck/blob/53b3f95b3a991161535a865515dc5f28f6c2b249/file/types.go#L599) struct.

file/validate.go
```diff
@@ -1,13 +1,66 @@
...
 import (
+	"encoding/json"
 	"fmt"
 
 	ghodss "github.com/ghodss/yaml"
 	"github.com/kong/deck/utils"
 	"github.com/xeipuuv/gojsonschema"
 )
 
+func check(item string, t *string) bool {
```
hbagdi, line 12: Please return an `error` instead of boolean.

file/validate.go
```diff
@@ -1,13 +1,66 @@
...
+	if t != nil || len(*t) > 0 {
+		fmt.Println(item + "[err] is not allowed to be specified.")
+		return false
+	}
+	return true
+}
+
+func checkDefaults(kd KongDefaults) bool {
+	if ret := check("Service.ID", kd.Service.ID); !ret {
```
hbagdi, line 21: This and every invocation to check has a nil exception. kd.Service is a pointer, which can be nil. Please add appropriate nil checks

file/validate.go
```diff
@@ -1,13 +1,66 @@
...
 
+func check(item string, t *string) bool {
+	if t != nil || len(*t) > 0 {
+		fmt.Println(item + "[err] is not allowed to be specified.")
+		return false
+	}
+	return true
+}
+
+func checkDefaults(kd KongDefaults) bool {
```
hbagdi, line 20: Please change this function to return []error. In the body, please collect all the errors and return the result.

file/validate.go
```diff
@@ -29,5 +82,16 @@ func validate(content []byte) error {
...
+
+	var kongdefaults *KongDefaults
+	if err := json.Unmarshal(content, kongdefaults); err != nil {
+		return fmt.Errorf("unmarshaling file into KongDefaults: %w", err)
+	}
+	if kongdefaults != nil {
+		if res := checkDefaults(*kongdefaults); !res {
+			return fmt.Errorf("fields are not allowed to specify in defaults")
+		}
+	}
```
hbagdi, line 94: Once the `checkDefaults` function returns []error, please merge errors returned from jsonSchema library with this result and return the whole thing.

hbagdi (changes requested):
Thanks for working on this, Jimin.
The approach looks good and is in the right direction but requires some additional work to actually make it work.
We will need unit tests for all the new code to get this merged in.

file/validate.go
rainest: Yeah the example was written directly into the comment without the benefit of a syntax checker :)

Is the gist of the change clear though?

hbagdi:
@ccfishk There are already existing reviews that have not been addressed.

file/validate.go
ccfishk: we need validate and error out each item within the struct, right. If so, we need pass in the specific DS as a parameter if I understand correctly.

ccfishk:
@hbagdi  done addressing comments.

file/validate.go
```diff
@@ -1,13 +1,71 @@
...
+	"strings"
 
 	ghodss "github.com/ghodss/yaml"
 	"github.com/kong/deck/utils"
 	"github.com/xeipuuv/gojsonschema"
 )
 
+func check(item string, t *string) string {
+	if t != nil || len(*t) > 0 {
+		fmt.Println(item + "[err] is not allowed to be specified.")
```
rainest, line 15: With the problem items now collected in the error, you'll want to remove this to avoid printing duplicate information:
```suggestion
```

file/validate.go
```diff
@@ -29,5 +87,16 @@ func validate(content []byte) error {
...
 		errs.Errors = append(errs.Errors, err)
 	}
+
+	var kongdefaults *KongDefaults
+	if err := json.Unmarshal(content, kongdefaults); err != nil {
+		return fmt.Errorf("unmarshaling file into KongDefaults: %w", err)
+	}
+	if kongdefaults != nil {
+		if err := checkDefaults(*kongdefaults); err != nil {
+			return fmt.Errorf("fields are not allowed to specify in defaults")
```
rainest, line 97: We do want to include the downstream error text here though:
```suggestion
			return fmt.Errorf("default values are not allowed for these fields: %w", err)
```
Also some rephrasing--the transitive "to specify" with no object in the original was confusing.

rainest (changes requested):

PR #443, opened 2021-06-16
refactor: simplify dependency definitions

hbagdi:
Kong and Konnect's data-model is relational in nature. This requires
decK to order its operations in a certain fashion. Creating an entity
without first ensuring that the present is present results in
violations. Deleting on the other hand, requires deleting every single
child before a parent is deleted to properly account for edges of the
entity relationship graph.

This was historically done by ordering entities in a particular order.
This was prone to some error, but even more so, very hard to grok for
anyone maintaining or reading the code.

This change builds upon the previous refactors(#435,#436) and decouples
the ordering of entities.

The dependency graph has been built up from ground up again to account
for all of 21 entities that decK can manage today.
Most of these entities were added organically and there dependencies
were haphazardly figured out. With this patch, the number of barriers
(`sc.wait()` calls) have been reduced to 4. With some inputs, users
will observe significant speed ups in calls to `decK sync` as
parallelism will be used more effectively.

With implementation, an additional Direct Acyclic Graph (DAG) library was
considered. An initial version even used Terraform's
[DAG](https://pkg.go.dev/github.com/hashicorp/terraform/dag#AcyclicGraph) but had
to be dropped for licensing reasons. The implemenetation is simple
without a DAG and doesn't have any limitations at the moment.

Please note that this change has large consequences since a missing dependency will
cause. Alas, there is no automated testing suite that can catch errors between
dependencies here. Some basic manual testing has been performed.

### DIFF
--- a/diff/order.go
+++ b/diff/order.go
@@ -1,0 +1,90 @@
+package diff
+
+import "github.com/kong/deck/types"
+
+/*
+                                       Root
+                                         |
+         +----------+----------+---------+------------+---------------+
+         |          |          |         |            |               |
+         v          v          v         v            v               v
+L1    Service    RbacRole  Upstream  Certificate  CACertificate  Consumer ---+
+      Package        |         |        |     |      |                |      |
+        |            v         v        v     |      v                v      |
+L2      |        RBACRole   Target     SNI    +-> Service       Credentials  |
+        |        Endpoint                         |  |              (7)      |
+        |                                         |  |                       |
+        |                                         |  |                       |
+L3      +---------------------------> Service <---+  +-> Route               |
+        |                             Version        |     |                 |
+        |                                 |          |     |                 |
+        |                                 |          |     v                 |
+L4      +----------> Document   <---------+          +-> Plugins  <----------+
+*/
+
+// dependencyOrder defines the order in which entities will be synced by decK.
+// Entities at the same level are processed concurrently.
+// Entities at level n will only be processed after all entities at level n-1
+// have been processed.
+// The processing order for create and update stage is top-down while that
+// for delete stage is bottom-up.
+var dependencyOrder = [][]types.EntityType{
+	{
+		types.ServicePackage,
+		types.RBACRole,
+		types.Upstream,
+		types.Certificate,
+		types.CACertificate,
+		types.Consumer,
+	},
+	{
+		types.RBACEndpointPermission,
+		types.Target,
+		types.SNI,
+		types.Service,
+
+		types.KeyAuth, types.HMACAuth, types.JWTAuth,
+		types.BasicAuth, types.OAuth2Cred, types.ACLGroup,
+		types.MTLSAuth,
+	},
+	{
+		types.ServiceVersion,
+		types.Route,
+	},
+	{
+		types.Plugin,
+		types.Document,
+	},
+}
+
+func order() [][]types.EntityType {
+	return deepCopy(dependencyOrder)
+}
+
+func reverseOrder() [][]types.EntityType {
+	order := deepCopy(dependencyOrder)
+	return reverse(order)
+}
+
+func reverse(src [][]types.EntityType) [][]types.EntityType {
+	src = deepCopy(src)
+	i := 0
+	j := len(src) - 1
+	for i < j {
+		temp := src[i]
+		src[i] = src[j]
+		src[j] = temp
+		i++
+		j--
+	}
+	return src
+}
+
+func deepCopy(src [][]types.EntityType) [][]types.EntityType {
+	res := make([][]types.EntityType, len(src))
+	for i := range src {
+		res[i] = make([]types.EntityType, len(src[i]))
+		copy(res[i], src[i])
+	}
+	return res
+}
--- a/diff/diff.go
+++ b/diff/diff.go
@@ -141,266 +141,28 @@ func (sc *Syncer) diff() error {
 }
 
 func (sc *Syncer) delete() error {
-	var err error
-
-	err = sc.entityDiffers[types.Plugin].Deletes(sc.queueEvent)
-	if err != nil {
-		return err
+	for _, types := range reverseOrder() {
+		for _, entityType := range types {
+			err := sc.entityDiffers[entityType].Deletes(sc.queueEvent)
+			if err != nil {
+				return err
+			}
+			sc.wait()
+		}
 	}
-	err = sc.entityDiffers[types.KeyAuth].Deletes(sc.queueEvent)
-	if err != nil {
-		return err
-	}
-	err = sc.entityDiffers[types.HMACAuth].Deletes(sc.queueEvent)
-	if err != nil {
-		return err
-	}
-	err = sc.entityDiffers[types.JWTAuth].Deletes(sc.queueEvent)
-	if err != nil {
-		return err
-	}
-	err = sc.entityDiffers[types.BasicAuth].Deletes(sc.queueEvent)
-	if err != nil {
-		return err
-	}
-	err = sc.entityDiffers[types.OAuth2Cred].Deletes(sc.queueEvent)
-	if err != nil {
-		return err
-	}
-	err = sc.entityDiffers[types.ACLGroup].Deletes(sc.queueEvent)
-	if err != nil {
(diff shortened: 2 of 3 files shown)


diff/order.go
```diff
@@ -0,0 +1,84 @@
...
+L2      |        RBACRole   Target     SNI    +-> Service       Credentials  |
+        |        Endpoint                         |  |              (7)      |
+        |                                         |  |                       |
+        |                                         |  |                       |
+L3      +---------------------------> Service <---+  +-> Route               |
+        |                             Version        |     |                 |
+        |                                 |          |     |                 |
+        |                                 |          |     v                 |
+L4      +----------> Document   <---------+          +-> Plugins  <----------+
+*/
```
hbagdi, line 23: Note to reviewers, please make sure that this dependency graph is accurate. The entire patch has been built based upon this.

diff/order.go
```diff
@@ -0,0 +1,84 @@
...
+		types.MTLSAuth,
+	},
+	{
+		types.ServiceVersion,
+		types.Route,
+	},
+	{
+		types.Plugin,
+		types.Document,
+	},
```
hbagdi, line 51: Note to reviewers, please then verify if the above ascii is reflected in this data structure correctly.
I took all entities at the same level and put them together.

diff/order.go
ccfishk: I like the gragh which brings in whole picture and clarity -- the message flow among dependencies. It would help to  understand if you add some concrete resource and actions in the picture. 

i.e. root (user) --> Certificate (Private, Public) --> SNI (authentication)

diff/order.go
ccfishk: I think this DS reflect the vertical relationship, the dependencies relationship is not that clear, how do you think ?

ccfishk (commented):
Add a couple of comments

diff/order.go
rainest: This looks accurate, but was it written (drawn?) by hand? That seems like it'd be quite a pain to update--I already have nightmares enough from editing big Markdown tables by hand, and this is more complex :)

IMO we'd want to use some sort of metadata format that we can then generate an SVG from for a visual representation. There are several standard graph formats that we can use an off-the-shelf SVG generator for. https://networkx.org/documentation/stable/reference/readwrite/index.html covers most of them (along with NetworkX's proprietary formats).

Only issue there is that the standard formats are all XML, which isn't my favorite for writing by hand (though it's still better than hand-formatted ASCII). However, [GEXF](https://gephi.org/gexf/format/data.html) isn't particularly complex XML, and once we have an initial version available adding additional relationships should be easy, since you'd just copy-pasta node and edge entries and edit the contents.

To Jimin's point, I don't know that we'd need much in terms of edge labels, since these all just reflect foreign key relationships in the Kong DB schema rather than methods that interact with the other objects. The only one that's a bit non-obvious is the Certificate->Service relationship, which exists because of the client_certificate field. GEXF does support edge labels if we want those though (the example doesn't show them, but they are in the spec).

diff/order.go
rainest: Yes and no--this correctly reflects the levels, but the levels alone aren't the dependency relationship, and this isn't actually a graph.

Whether that matters or not is "probably not" though, since simply grouping things into levels is sufficient for achieving the goal of the refactor, i.e. that we no longer use an absolute order for all entities and waits, and instead simply place entities into a group below all their required parents, with waits between the groups.

I don't see anything that an actual DAG would get us over this approach for that need, though it may have other uses--in the KIC context a per-entity DAG would provide us more rich status information and allow us to selectively drop entities and their children from the config source (e.g. do not create a route from an Ingress when its target Service is not present, mark the Ingress unsatisfied because the Service is missing). However, that's a much more radical change because it's entity relationships rather than entity class relationships, and is more complicated to implement (we'd need an actual graph type and BFS that waits between the levels).

I don't think there'd be much benefit to a graph with the entity class approach yet. It'd mean we could handle, say, credential creation independent of upstream creation, but in practice our graph is simple enough at present (and unlikely to get hugely more complex) that I think the simple level approach is fine. It may matter because of type cardinality: for example, if you have a very large number of consumers checking all those blocks creation of new targets, though I don't know whether that has a meaningful UX impact--that block _already existed_ in the approach prior to this refactor, and I don't think anyone was complaining about it.

diff/order.go
mmorel-35: Have you tried mermaid, [for example](https://mermaid.cloudinary.us/#/edit/eyJjb2RlIjoiZ3JhcGggVERcbkFbUm9vdF0gLS4tPiBCW1NlcnZpY2UgUGFja2FnZV0gXG5BIC0uLT4gQ1tSYmFjUm9sZV1cbkEgLS4tPiBEW1Vwc3RyZWFtXVxuQSAtLi0-IEVbQ2VydGlmaWNhdGVdXG5BIC0uLT4gRltDQUNlcnRpZmljYXRlXVxuQSAtLi0-IEdbQ29uc3VtZXJdXG5DIC0uLT4gSFtSYmFjUm9sZUVuZHBvaW50XVxuRCAtLi0-IElbVGFyZ2V0XVxuRSAtLi0-IEpbU05JXVxuRSAtLi0-IEtbU2VydmljZV1cbkYgLS4tPiBLXG5HIC0uLT4gTFtDcmVkZW50aWFsc11cbkIgLS4tPiBNW1NlcnZpY2VWZXJzaW9uXVxuSyAtLi0-IE1cbksgLS4tPiBOW1JvdXRlXVxuQiAtLi0-IE9bRG9jdW1lbnRdXG5NIC0uLT4gT1xuSyAtLi0-IFBbUGx1Z2luXVxuTiAtLi0-IFBcbkcgLS4tPiBQIiwibWVybWFpZCI6eyJ0aGVtZSI6ImRlZmF1bHQifX0)?

diff/order.go
```diff
@@ -0,0 +1,84 @@
...
+        |                                         |  |                       |
+        |                                         |  |                       |
+L3      +---------------------------> Service <---+  +-> Route               |
+        |                             Version        |     |                 |
+        |                                 |          |     |                 |
+        |                                 |          |     v                 |
+L4      +----------> Document   <---------+          +-> Plugins  <----------+
+*/
+
+var dependencyOrder = [][]string{
```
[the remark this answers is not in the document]
hbagdi: > Creating deep copies and implementing reversal just to make it possible to iterate over an array in a reverse order seems superfluous. Consider the following alternative that means less code (and less memory copying):

Although this is more code, I want to keep the ordering separate. The motivation here is that the ordering will be injected into the Syncer rather than hard-coding it in there.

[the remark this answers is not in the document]
hbagdi: I took your suggestion and reworded it a little bit.

diff/order.go
hbagdi: > This looks accurate, but was it written (drawn?) by hand? That seems like it'd be quite a pain to update--I already have nightmares enough from editing big Markdown tables by hand, and this is more complex :)

I used https://asciiflow.com to generate this diagram.

hbagdi: >  It would help to understand if you add some concrete resource and actions in the picture.

All of these are entities in Kong and Konnect's data-model and nothing here is abstract.
I could be misunderstanding you, in which case please elaborate on what you mean.

diff/order.go
hbagdi: > I think this DS reflect the vertical relationship, the dependencies relationship is not that clear, how do you think ?

This DS doesn't track the dependency relationships because that is not required for execution. We simplify the graph into levels as that is enough information to correctly execute.

hbagdi:
@rainest We can use something more funky but I think that gets harder to maintain. 
We do not regularly add entities to decK so maintaining an ASCII diagram shouldn't be very cumbersome.

@mflendrich @ccfishk I've incorporated your comments and responded to your comments, please take another look.

diff/order.go
rainest: > I used https://asciiflow.com to generate this diagram.

Did you happen to save the share link for it? It looks like it can save the drawing using some custom URL data format, but doesn't have any way to import existing stuff.

Issue #449, opened 2021-06-24
Timeout using dump


fjaenni:
Hi!

I have a problem with **dump** use. I call to a platform wich have Kong version 1.3.X, and when sends operations in parallel to obtain all services, routes, plugins... it takes more than 10 seconds (deffined in defaultClientTimeout property) and then returns the following error:

`context deadline exceeded (Client.Timeout exceeded while awaiting headers)`

A single call (one of the parallels previous calls) takes over 3 seconds.

Exists some way to set the timeout?. If this is not possible, exist some way to set a limit of parallels calls?

Thank you very much!!
Best regards

PR #450, opened 2021-06-25
Set client timeout https://github.com/Kong/deck/issues/449

fjaenni:
#449

### DIFF
--- a/utils/types.go
+++ b/utils/types.go
@@ -19,11 +19,7 @@ import (
 	"github.com/kong/go-kong/kong/custom"
 )
 
-const (
-	defaultClientTimeout = 10 * time.Second
-)
-
-var clientTimeout  time.Duration
+var ClientTimeout time.Duration
 
 // KongRawState contains all of Kong Data
 type KongRawState struct {
@@ -133,12 +129,8 @@ func GetKongClient(opt KongClientConfig) (*kong.Client, error) {
 		}
 		tlsConfig.RootCAs = certPool
 	}
-	
-	clientTimeout = defaultClientTimeout
-	if opt.Timeout > 0{
-		clientTimeout =  time.Duration(opt.Timeout) * time.Second
-	}
 
+	ClientTimeout = time.Duration(opt.Timeout) * time.Second
 	c := opt.HTTPClient
 	if c == nil {
 		c = HTTPClient()
@@ -213,12 +205,12 @@ func CleanAddress(address string) string {
 // sane default timeouts.
 func HTTPClient() *http.Client {
 	return &http.Client{
-		Timeout: clientTimeout,
+		Timeout: ClientTimeout,
 		Transport: &http.Transport{
 			DialContext: (&net.Dialer{
-				Timeout: clientTimeout,
+				Timeout: ClientTimeout,
 			}).DialContext,
-			TLSHandshakeTimeout: clientTimeout,
+			TLSHandshakeTimeout: ClientTimeout,
 		},
 	}
 }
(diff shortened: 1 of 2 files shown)


utils/types.go
```diff
@@ -23,6 +23,8 @@ const (
 	defaultClientTimeout = 10 * time.Second
 )
 
+var ClientTimeout  time.Duration
```
mmorel-35, line 26:
```suggestion
var clientTimeout  time.Duration
```

utils/types.go
```diff
@@ -129,6 +133,11 @@ func GetKongClient(opt KongClientConfig) (*kong.Client, error) {
 		}
 		tlsConfig.RootCAs = certPool
 	}
+	
+	ClientTimeout = defaultClientTimeout
```
mmorel-35, line 137:
```suggestion
	clientTimeout = defaultClientTimeout
```

utils/types.go
```diff
@@ -129,6 +133,11 @@ func GetKongClient(opt KongClientConfig) (*kong.Client, error) {
...
+	if opt.Timeout > 0{
+		ClientTimeout =  time.Duration(opt.Timeout) * time.Second
```
mmorel-35, line 139:
```suggestion
		clientTimeout =  time.Duration(opt.Timeout) * time.Second
```

utils/types.go
```diff
@@ -204,12 +213,12 @@ func CleanAddress(address string) string {
 // sane default timeouts.
 func HTTPClient() *http.Client {
 	return &http.Client{
-		Timeout: defaultClientTimeout,
+		Timeout: ClientTimeout,
```
mmorel-35, line 216:
```suggestion
		Timeout: clientTimeout,
```

utils/types.go
```diff
@@ -204,12 +213,12 @@ func CleanAddress(address string) string {
...
 		Transport: &http.Transport{
 			DialContext: (&net.Dialer{
-				Timeout: defaultClientTimeout,
+				Timeout: ClientTimeout,
```
mmorel-35, line 219:
```suggestion
				Timeout: clientTimeout,
```

utils/types.go
```diff
@@ -204,12 +213,12 @@ func CleanAddress(address string) string {
...
 			}).DialContext,
-			TLSHandshakeTimeout: defaultClientTimeout,
+			TLSHandshakeTimeout: ClientTimeout,
```
mmorel-35, line 221:
```suggestion
			TLSHandshakeTimeout: clientTimeout,
```

mmorel-35 (commented):
Does it really needs to be a public variable? Lowercase shall be enough?

franciscojn (dismissed):

rainest (changes requested):
@fjaenni some minor linter issues that need addressing.

I'd recommend placing the default timeout in the flag (`rootCmd.PersistentFlags().Int("timeout", 10, ...)`) and using that value throughout. Otherwise it's a bit confusing that the value shown in help doesn't reflect the actual default timeout--it suggests that the default is instead no timeout.

fjaenni:
> @fjaenni some minor linter issues that need addressing.
> 
> I'd recommend placing the default timeout in the flag (`rootCmd.PersistentFlags().Int("timeout", 10, ...)`) and using that value throughout. Otherwise it's a bit confusing that the value shown in help doesn't reflect the actual default timeout--it suggests that the default is instead no timeout.

hi,
I agree, I thought about doing it like this but finally ...
Thank you so much

rainest (approved):

PR #455, opened 2021-07-06
fix: multiple bugs with credentials and select-tags

hbagdi:
Please see #414 for some background and an alternate solution that was
considered but had to be dropped for backwards-compatibility reasons.

This patch fixes the following bugs:
- file package
  - statebuilder sets tags on credentials conditionally but the variable
  containing the version of kong was never being set in the first place.
  We do that correctly now. Please note that we don't set these
  variables for the code-path for Konnect because control-plane in
  Konnect is not-versioned from our customer's perspective. That and
  select-tag feature is not available in Konnect.
  - mTLS auth credentials received tagging support in Kong Enterprise
  2.3. decK will now set the select-tag in mtls-auth credentials when
  Kong version is compatible.
- dump package:
  - #282 incorrectly only exported tagged mtls-auth credentials even
  though they weren't tagged (by decK or otherwise). We have resorted to
  always exporting all mtls-auth credentials to avoid a breaking change
  for our existing users. Please see the code comment for how this
  choice shouldn't be a problem for most users.

### DIFF
--- a/file/builder.go
+++ b/file/builder.go
@@ -405,6 +405,7 @@ func (b *stateBuilder) ingestACLGroups(creds []kong.ACLGroup) error {
 }
 
 func (b *stateBuilder) ingestMTLSAuths(creds []kong.MTLSAuth) {
+	kong230Version := semver.MustParse("2.3.0")
 	for _, cred := range creds {
 		cred := cred
 		// normally, we'd want to look up existing resources in this case
@@ -412,11 +413,9 @@ func (b *stateBuilder) ingestMTLSAuths(creds []kong.MTLSAuth) {
 		// so we don't--schema validation requires the ID
 		// there's nothing more to do here
 
-		// TODO: this is stub code, since mtls-auth doesn't actually have tag support yet
-		// They probably should, FTI-1706 tracks that request with the Kong Enterprise team
-		//if b.kongVersion.GTE(kong220Version) {
-		//	utils.MustMergeTags(&cred, b.selectTags)
-		//}
+		if b.kongVersion.GTE(kong230Version) {
+			utils.MustMergeTags(&cred, b.selectTags)
+		}
 		b.rawState.MTLSAuths = append(b.rawState.MTLSAuths, &cred)
 	}
 }
--- a/file/reader.go
+++ b/file/reader.go
@@ -36,6 +36,7 @@ func GetForKonnect(fileContent *Content, opt RenderConfig) (*utils.KongRawState,
 	// setup
 	builder.targetContent = fileContent
 	builder.currentState = opt.CurrentState
+	builder.kongVersion = opt.KongVersion
 
 	kongState, konnectState, err := builder.build()
 	if err != nil {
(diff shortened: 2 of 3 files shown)


file/builder.go
```diff
@@ -405,18 +405,17 @@ func (b *stateBuilder) ingestACLGroups(creds []kong.ACLGroup) error {
...
 		// however, this is impossible here: mtls-auth simply has no unique fields other than ID,
 		// so we don't--schema validation requires the ID
 		// there's nothing more to do here
 
-		// TODO: this is stub code, since mtls-auth doesn't actually have tag support yet
-		// They probably should, FTI-1706 tracks that request with the Kong Enterprise team
-		//if b.kongVersion.GTE(kong220Version) {
-		//	utils.MustMergeTags(&cred, b.selectTags)
-		//}
+		if b.kongVersion.GTE(kong230Version) {
```
mflendrich, line 416: _this is not an action item for this PR, but something to consider for the future regarding code health_

`stateBuilder` is a function object that takes some input data (current state and target state, and kong version) and produces some output data.

1. The distinction between inputs and outputs is not very clear in the implementation of `stateBuilder`. I would like to put on the table (for consideration) the possibility of refactoring stateBuilder into pure mathematical functions with no preserved state - similar to what happened to `parser` in KIC. I believe that this would help with testability and readability. Of course, let's be mindful of priorities in our projects - I only want these considerations to contribute to the vision of where we want to take the codebase to in small steps.
2. When we think of stateBuilder as a mathematical function, like described above: A mathematical function changing its outputs based on version of some software dependency (here: gateway version) suffers from two problems:
  2.1. keeps stateBuilder (a mathematical function processing X into Y) concerned with a version of "some external" component - which feels at odds with SRP
  2.2. in tests, couples logic with software versions. I believe that the role of a unit test here would be to check that "state builder properly adds tags when asked to, and does not when not asked to". "Asking" the state builder to add tags or not depending on Gateway version is a separate concern imo, and should happen on a higher level (and would be tested on a higher level too, in an integration test).

To make the above considerations actionable, I believe that the following would help improve the code health:
- kick `kongVersion` out of `stateBuilder` to a higher level, and replace it with knobs that control whether selectTags get applied to all creds, or all creds but mtls-auth, or to none.
- refactor stateBuilder into pure stateless functions

mflendrich (commented):
Leaving the ultimate review to @rainest, but added some thoughts. No action required from my side.

file/reader.go
```diff
@@ -36,6 +36,7 @@ func GetForKonnect(fileContent *Content, opt RenderConfig) (*utils.KongRawState,
 	// setup
 	builder.targetContent = fileContent
 	builder.currentState = opt.CurrentState
+	builder.kongVersion = opt.KongVersion
```
rainest, line 39: To confirm, we just set this to avoid a segfault; it won't actually have a value? Looks like that's the case.

rainest (approved):
One minor question to confirm, but I'm pretty sure that's indeed the answer, so :lgtm:

file/reader.go
hbagdi: The empty value for semver.Version is 0.0 so there won't be a segfault. This is to guard against the problem from cropping again if there comes a point that we care about the version of Kong Control-Plane in the Konnect world.

file/builder.go
hbagdi: Tracking in https://github.com/Kong/deck/issues/457.

PR #460, opened 2021-07-12
fix: correct formats where conversion is allowed

hbagdi:
Previously, KongGateway was listed in --to and Konnect was listed in
--to, which wasn't supported.
This fixes the documentation so that the user isn't confused anymore.


rainest (approved):
IMO including the `[]convert.Format{...}` inline with the Sprint would be more readable if we don't need to reuse the format lists, but that's a minor style nit that doesn't really matter--the variable definitions are close by, so vov.

Issue #462, opened 2021-07-14
Redact value from placeholders


carnei-ro:
The feature from this PR is really awesome https://github.com/Kong/deck/pull/286, although I have some concerns in here... Using env var as placeholder for values may imply the value is "secret". It would be nice if we could have a flag to display those values as `[masked]` at diffs or applies outputs.

carnei-ro:
the idea is in the PR https://github.com/Kong/deck/pull/463

PR #463, opened 2021-07-14
feat: mask environment variables from outputs

carnei-ro:
fixes https://github.com/Kong/deck/issues/462

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -81,7 +81,7 @@ func getWorkspaceName(workspaceFlag string, targetContent *file.Content) string 
 }
 
 func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
-	delay int, workspace string,
+	delay int, workspace string, noMaskValues bool,
 ) error {
 	// read target file
 	targetContent, err := file.GetContentFromFiles(filenames)
@@ -225,7 +225,7 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 		return err
 	}
 
-	totalOps, err := performDiff(ctx, currentState, targetState, dry, parallelism, delay, kongClient)
+	totalOps, err := performDiff(ctx, currentState, targetState, dry, parallelism, delay, kongClient, noMaskValues)
 	if err != nil {
 		return err
 	}
@@ -273,7 +273,7 @@ func fetchCurrentState(ctx context.Context, client *kong.Client, dumpConfig dump
 }
 
 func performDiff(ctx context.Context, currentState, targetState *state.KongState,
-	dry bool, parallelism int, delay int, client *kong.Client,
+	dry bool, parallelism int, delay int, client *kong.Client, noMaskValues bool,
 ) (int, error) {
 	s, err := diff.NewSyncer(diff.SyncerOpts{
 		CurrentState:  currentState,
@@ -285,7 +285,7 @@ func performDiff(ctx context.Context, currentState, targetState *state.KongState
 		return 0, err
 	}
 
-	stats, errs := s.Solve(ctx, parallelism, dry)
+	stats, errs := s.Solve(ctx, parallelism, dry, noMaskValues)
 	// print stats before error to report completed operations
 	printStats(stats)
 	if errs != nil {
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -27,8 +27,9 @@ const (
 )
 
 var (
-	dumpConfig dump.Config
-	assumeYes  bool
+	dumpConfig   dump.Config
+	assumeYes    bool
+	noMaskValues bool
 )
 
 type mode int
@@ -81,7 +82,7 @@ func getWorkspaceName(workspaceFlag string, targetContent *file.Content) string 
 }
 
 func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
-	delay int, workspace string, noMaskValues bool,
+	delay int, workspace string,
 ) error {
 	// read target file
 	targetContent, err := file.GetContentFromFiles(filenames)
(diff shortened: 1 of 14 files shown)


dancallaghan:
Just wondering what the status was on this PR? We noticed that the `diff` command was resulting in secrets being leaked in our CI.

thesheps:
> Just wondering what the status was on this PR? We noticed that the `diff` command was resulting in secrets being leaked in our CI.

Can confirm this is also causing us problem across all our environments. More than happy to contribute or review any supporting PRs if that's of any use?

hbagdi:
This looks promising. Could you please add an integration test for this?

carnei-ro:
@hbagdi TBH I have no idea how to create a "diff" integration test that actually looks to the diff output :disappointed:  
Is there any "CONTRIBUTING" doc to help me out?  
I've rebased, created a simple unit test, and edited the existing integration test. It is masking the value.

```
=== RUN   Test_Diff_Workspace
=== RUN   Test_Diff_Workspace/diff_with_not_existent_workspace_doesn't_error_out
creating workspace test
creating service svc1  {
+  "connect_timeout": 60000
+  "host": "mockbin.org"
+  "id": "d0d2ec9b-fe41-4516-be12-50766a770523"
+  "name": "[masked]"
+  "protocol": "http"
+  "read_timeout": 60000
+  "write_timeout": 60000
 }
```

hbagdi:
@GGabriele Could you help out, please?

carnei-ro:
Thanks @GGabriele it really did the job. Masked and Unmasked for Kong < 3 and Kong >= 3 tests are there!

diff/diff_helpers_test.go
```diff
@@ -145,3 +146,35 @@ bar
...
+
+func Test_MaskEnvVarsValues(t *testing.T) {
+	tests := []struct {
+		name    string
+		args    string
+		want    string
+		envVars map[string]string
+	}{
+		{
+			name: "JSON",
```
GGabriele, line 158: Can you please add a test cases making sure that:
1. an unused env var doesn't result in any masked field (like having `DECK_FOO=foo` with no `foo` field existing)
2. an env var only partially matching a value doesn't result in any masked field (like having `DECK_FOO=foo` with some value being `foo_test`)
3. an env var used multiple times results in multiple masked values

carnei-ro: 1 and 2 won't work as you mentioned because of the implementation:

```golang
func maskEnvVarValue(diffString string) string {
	for _, envVar := range parseDeckEnvVars() {
		diffString = strings.Replace(diffString, envVar.Value, "[masked]", -1)
	}
	return diffString
}
```

This is the same behavior of Jenkins or Gitlab uses IIRC.

3 will work and I can implement it. 

Just let me know if you think 1 & 2 is a blocker

GGabriele: Yeah I think something like this would be good enough:

```
		{
			name: "JSON",
			envVars: map[string]string{
				"DECK_BARR": "barbar",
				"DECK_BAZZ": "bazbaz",   // used more than once
				"DECK_FUB":  "fubfub",   // unused
				"DECK_FOO":  "foo_test", // unused, partial match
			},
			args: `{"foo":"foo","bar":"barbar","baz":"bazbaz","another":"bazbaz"}`,
			want: `{"foo":"foo","bar":"[masked]","baz":"[masked]","another":"[masked]"}`,
		},
```

Other than that, this look great already!

diff/diff.go
```diff
@@ -322,8 +322,34 @@ type Stats struct {
...
+	}
+	if !noMaskValues {
+		diffString = maskEnvVarValue(diffString)
+	}
+	return diffString, err
+}
+
 // Solve generates a diff and walks the graph.
-func (sc *Syncer) Solve(ctx context.Context, parallelism int, dry bool) (Stats, []error) {
+func (sc *Syncer) Solve(ctx context.Context, parallelism int, dry bool, noMaskValues bool) (Stats, []error) {
```
GGabriele, line 352: Instead of changing this public contract, can you add the `noMaskValues` boolean as an attribute of `Syncer`? That way, you can simply do something like this in `cmd/common.go` (as well as in the other places):

```
	s, err := diff.NewSyncer(diff.SyncerOpts{
		CurrentState:  currentState,
		TargetState:   targetState,
		KongClient:    client,
		StageDelaySec: delay,
		NoMaskValues: noMaskValues,
	})
	if err != nil {
		return 0, err
	}

	stats, errs := s.Solve(ctx, parallelism, dry)
```

`Syncer` is also used by [KIC](https://github.com/Kong/kubernetes-ingress-controller/blob/main/internal/dataplane/sendconfig/sendconfig.go#L233), so we cannot break it.

diff/diff_helpers_test.go
carnei-ro: okay! I added tests for this scenario

cmd/common.go
```diff
@@ -81,7 +81,7 @@ func getWorkspaceName(workspaceFlag string, targetContent *file.Content) string
 }
 
 func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
-	delay int, workspace string,
+	delay int, workspace string, noMaskValues bool,
```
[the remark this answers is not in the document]
carnei-ro: Got it, I'll do

diff/diff.go
carnei-ro: on it

GGabriele:
@carnei-ro I just added a couple of more comments which would require a bit of refactoring (sorry for not pushing these earlier).

@rainest can you take a look at this PR too? I want to make sure KIC is good with these changes (and the ones I proposed)

GGabriele (approved):

GGabriele:
@carnei-ro  Awesome job here! Merging this now, thanks a lot again!

carnei-ro:
@GGabriele, thanks for merging it! Does decK project have custom contributor T-shirts? I'm collecting them - I already have Kong, KIC and it would be great to have one of decK.

GGabriele:
> Does decK project have custom contributor T-shirts?

Unfortunately it doesn't :( not yet at least

PR #469, opened 2021-07-27
[wip] prevent user overwrite default configuration 

### LINKED ISSUE #421: restrict some fields from being set in default values

ccfishk:
prevent user overwrite some default configuration.

fixes #421


hbagdi:
feedback from https://github.com/Kong/deck/pull/441 hasn't yet been addressed. Please add tests and resolve the nil reference issues.

ccfishk:
> feedback from #441 hasn't yet been addressed. Please add tests and resolve the nil reference issues.

Tests are added, and nil reference was removed if I understand correctly.

file/validate.go
```diff
@@ -29,5 +31,70 @@ func validate(content []byte) error {
...
+	}
+	if kongdefaults != nil {
+		if err := CheckDefaults(*kongdefaults); err != nil {
+			return fmt.Errorf("default values are not allowed for these fields: %w", err)
+		}
+	}
 	return errs
 }
+
+func Check(item string, t *string) string {
```
hbagdi, line 47: Please lowercase this to not export this function.

file/validate.go
```diff
@@ -29,5 +31,70 @@ func validate(content []byte) error {
...
+	if t != nil && len(*t) > 0 {
+		return item
+	}
+	return ""
+}
+
+func CheckDefaults(kd KongDefaults) error {
```
hbagdi, line 54: Please lowercase this to not export this function.

file/validate.go
```diff
@@ -29,5 +31,70 @@ func validate(content []byte) error {
...
+	var invalid []string
+	if ret := Check("Service.ID", kd.Service.ID); len(ret) > 0 {
```
hbagdi, line 56: kd.Service is a pointer. What happens if it is a `nil`?

file/validate.go
```diff
@@ -29,5 +31,70 @@ func validate(content []byte) error {
...
+		invalid = append(invalid, "Service.ID")
+	}
+
+	if ret := Check("Service.Host", kd.Service.Host); len(ret) > 0 {
```
hbagdi, line 60: kd.Service is a pointer. What happens if it is a `nil`?

file/validate.go
```diff
@@ -29,5 +31,70 @@ func validate(content []byte) error {
...
+		invalid = append(invalid, "Service.Host")
+	}
+
+	if ret := Check("Service.Name", kd.Service.Name); len(ret) > 0 {
```
hbagdi, line 64: kd.Service is a pointer. What happens if it is a `nil`?

file/validate.go
```diff
@@ -29,5 +31,70 @@ func validate(content []byte) error {
...
+		invalid = append(invalid, "Service.Name")
+	}
+
+	if kd.Service.Port != nil {
```
hbagdi, line 68: kd.Service is a pointer. What happens if it is a `nil`?

file/validate.go
```diff
@@ -29,5 +31,70 @@ func validate(content []byte) error {
...
+		invalid = append(invalid, "Service.Port")
+	}
+
+	if ret := Check("Route.Name", kd.Route.Name); len(ret) > 0 {
```
hbagdi, line 72: Same comments for most of these lines..

hbagdi (changes requested):

file/validate.go
ccfishk: hmm, possibly, if deck has another client other than KIC.

file/validate.go
ccfishk: changed 67aa05d6a0a1841031b4910928070b31bfaf9be8

file/validate.go
ccfishk: changed 67aa05d6a0a1841031b4910928070b31bfaf9be8

file/validate.go
ccfishk: checked 67aa05d6a0a1841031b4910928070b31bfaf9be8

file/validate.go
ccfishk: changed 67aa05d6a0a1841031b4910928070b31bfaf9be8

file/validate.go
ccfishk: checked 67aa05d6a0a1841031b4910928070b31bfaf9be8

file/validate.go
ccfishk: checked 67aa05d6a0a1841031b4910928070b31bfaf9be8

file/validate.go
```diff
@@ -29,5 +31,86 @@ func validate(content []byte) error {
 		err := fmt.Errorf(desc.String())
 		errs.Errors = append(errs.Errors, err)
 	}
+
+	var kongdefaults *KongDefaults
+	if err := json.Unmarshal(content, kongdefaults); err != nil {
```
hbagdi, line 36: Have you actually tried running decK with this patch?
This won't work at all. The content in question should represents the following struct: https://github.com/Kong/deck/blob/main/file/types.go#L597-L617

file/validate.go
```diff
@@ -29,5 +31,86 @@ func validate(content []byte) error {
...
+	if t != nil && len(*t) > 0 {
+		return item
+	}
+	return ""
+}
+
+func checkDefaults(kd KongDefaults) error {
+	var invalid []string
+	if kd.Service == nil {
+		invalid = append(invalid, "Invalid Service Address")
```
hbagdi, line 57: This is not an error because the user can choose to not specify a default for service.
If service is nil, you can skip checking any of the properties all together and that's it.

The same comment applies for route, target and upstream.

file/validate.go
ccfishk: Yesterday's offline discussion KIC is a deck client, and the File.Content is the main construct I explored atm (edited) 
white_check_mark
eyes
raised_hands





10:04
Assuming decK applies the same interface to all clients.

file/validate.go
ccfishk: changed 507639081e5cb99ce9e720a28e58c17b7afdafc4

PR #485, opened 2021-09-06
chore(ci): codeql-analysis on go 1.16

mmorel-35:
In https://github.com/Kong/deck/runs/3510966256?check_suite_focus=true we can see
>Extraction failed in file/schema.go with error package embed is not in GOROOT (/opt/hostedtoolcache/go/1.15.15/x64/src/embed)

But embed is a 1.16 capacity, using 1.16 is fixing it

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -44,7 +44,7 @@ jobs:
     - name: Setup go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.16
+        go-version: 1.17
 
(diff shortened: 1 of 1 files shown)


hbagdi:
I'm not sure if this works or not - I'm not familiar with codeql. Can you bump it up to Go 1.17 (since that's the version we use) and then we can try it out?

mmorel-35:
Hi @hbagdi, the scan is not raising the initial error neither with 1.16 or 1.17

hbagdi (approved):

Issue #190, opened 2020-07-14
feature: introduce a more robust validation mode


hbagdi:
Currently, decK provides a `validate` command which does JSON schema validation of the input and also creates an in-memory state. While this detects some errors with user-input, we can do more.
Kong ships with a validate endpoint for each entity:
https://docs.konghq.com/2.0.x/db-less-admin-api/#validate-a-configuration-against-a-schema

A new mode for `validate` command (gated using a boolean CLI flag) should run all the entities present in the state against the Admin API endpoint. This should catch almost all errors with the user input.

PR #502, opened 2021-10-05
feat: add support to validate via Kong API

GGabriele:
Hi! In order to get more familiarity with the tool and the codebase, I decided to give it a try and work on the proposal discussed in https://github.com/Kong/deck/issues/190

Sample usage:

```
$ deck validate --help
The validate command reads the state file and ensures validity.
It reads all the specified state files and reports YAML/JSON
parsing issues. It also checks for foreign relationships
and alerts if there are broken relationships, or missing links present.

When ran with the --use-kong-api flag, it also uses the 'validate' endpoint in Kong
to validate entities configuration.

Usage:
  deck validate [flags]

Flags:
  -h, --help                  help for validate
      --rbac-resources-only   indicate that the state file(s) contains RBAC resources only (Kong Enterprise only).
  -s, --state strings         file(s) containing Kong's configuration.
                              This flag can be specified multiple times for multiple files.
                              Use '-' to read from stdin. (default [kong.yaml])
      --use-kong-api          perform schema validation against Kong API.
```

Sample input:

```
services:
- name: svc1
  host: mockbin.org:80
  tags:
  - team-svc1
  routes:
  - name: r1
    https_redirect_status_code: 301
    paths:
    - /r1
certificates:
- id: 13c562a1-191c-4464-9b18-e5222b46035b
  cert: |
    -----BEGIN CERTIFICATE-----
    -----END CERTIFICATE-----
  key: |
    -----BEGIN PRIVATE KEY-----
    -----END PRIVATE KEY-----
  tags:
  - cloudops-managed
  snis:
  - name: demo1.example.com:123
  - name: demo2.example.com
  - name: demo3.example.com
plugins:
- name: prometheus
  enabled: true
  protocols:
  - httpss
  - https
 ```

Sample output:

```
$ deck validate --use-kong-api
services: HTTP status 400 (message: "schema violation (host: must not have a port)")
certificates: HTTP status 400 (message: "2 schema violations (cert: invalid certificate: x509.new: asn1/tasn_dec.c:309:error:0D07803A:asn1 encoding routines:asn1_item_embed_d2i:nested asn1 error; key: invalid key: pkey.new:load_key: asn1/tasn_dec.c:309:error:0D07803A:asn1 encoding routines:asn1_item_embed_d2i:nested asn1 error)")
plugins: HTTP status 400 (message: "schema violation (protocols.1: expected one of: grpc, grpcs, http, https, tcp, tls, udp)")
snis: HTTP status 400 (message: "schema violation (name: must not have a port)")
```

Let me know what you think! :)

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -58,11 +58,20 @@ func workspaceExists(ctx context.Context, config utils.KongClientConfig, workspa
 		return false, err
 	}
 
-	exists, err := rootClient.Workspaces.ExistsByName(ctx, &workspaceName)
+	exists, err := rootClient.Workspaces.Exists(ctx, &workspaceName)
 	if err != nil {
 		return false, fmt.Errorf("checking if workspace exists: %w", err)
 	}
 	return exists, nil
+}
+
+func getWorkspaceName(workspace string, targetContent *file.Content) string {
+	if workspace != targetContent.Workspace && workspace != "" {
+		cprint.DeletePrintf("Warning: Workspace '%v' specified via --workspace flag is "+
+			"different from workspace '%v' found in state file(s).\n", workspace, targetContent.Workspace)
+		return workspace
+	}
+	return targetContent.Workspace
 }
 
 func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
@@ -79,16 +88,9 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 		return err
 	}
 
-	var wsConfig utils.KongClientConfig
-	var workspaceName string
 	// prepare to read the current state from Kong
-	if workspace != targetContent.Workspace && workspace != "" {
-		cprint.DeletePrintf("Warning: Workspace '%v' specified via --workspace flag is "+
-			"different from workspace '%v' found in state file(s).\n", workspace, targetContent.Workspace)
-		workspaceName = workspace
-	} else {
-		workspaceName = targetContent.Workspace
-	}
+	var wsConfig utils.KongClientConfig
+	workspaceName := getWorkspaceName(workspace, targetContent)
 	wsConfig = rootConfig.ForWorkspace(workspaceName)
 
 	// load Kong version after workspace
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -26,6 +26,19 @@ var (
 	assumeYes  bool
 )
 
+func performValidate(ctx context.Context, kongClient *kong.Client, entity string) (bool, error) {
+	endpoint := fmt.Sprintf("/schemas/%s/validate", entity)
+	req, err := kongClient.NewRequest("POST", endpoint, nil, nil)
+	if err != nil {
+		return false, err
+	}
+	resp, err := kongClient.Do(ctx, req, nil)
+	if err != nil {
+		return false, err
+	}
+	return resp.StatusCode == http.StatusOK, nil
+}
+
 // workspaceExists checks if workspace exists in Kong.
 func workspaceExists(ctx context.Context, config utils.KongClientConfig, workspaceName string) (bool, error) {
 	rootConfig := config.ForWorkspace("")
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -26,20 +26,6 @@ var (
 	assumeYes  bool
 )
 
-func validateEntity(ctx context.Context, kongClient *kong.Client, entityType string, entity interface{}) (bool, error) {
-	errWrap := "validate entity '%s': %s"
-	endpoint := fmt.Sprintf("/schemas/%s/validate", entityType)
-	req, err := kongClient.NewRequest(http.MethodPost, endpoint, nil, entity)
-	if err != nil {
-		return false, fmt.Errorf(errWrap, entityType, err)
-	}
-	resp, err := kongClient.Do(ctx, req, nil)
-	if err != nil {
-		return false, fmt.Errorf(errWrap, entityType, err)
-	}
-	return resp.StatusCode == http.StatusOK, nil
-}
-
 // workspaceExists checks if workspace exists in Kong.
 func workspaceExists(ctx context.Context, config utils.KongClientConfig, workspaceName string) (bool, error) {
 	rootConfig := config.ForWorkspace("")
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -58,7 +58,7 @@ func workspaceExists(ctx context.Context, config utils.KongClientConfig, workspa
 		return false, err
 	}
 
-	exists, err := rootClient.Workspaces.Exists(ctx, &workspaceName)
+	exists, err := rootClient.Workspaces.ExistsByName(ctx, &workspaceName)
 	if err != nil {
 		return false, fmt.Errorf("checking if workspace exists: %w", err)
 	}
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -65,11 +65,11 @@ func workspaceExists(ctx context.Context, config utils.KongClientConfig, workspa
 	return exists, nil
 }
 
-func getWorkspaceName(workspace string, targetContent *file.Content) string {
-	if workspace != targetContent.Workspace && workspace != "" {
+func getWorkspaceName(workspaceFlag string, targetContent *file.Content) string {
+	if workspaceFlag != targetContent.Workspace && workspaceFlag != "" {
 		cprint.DeletePrintf("Warning: Workspace '%v' specified via --workspace flag is "+
-			"different from workspace '%v' found in state file(s).\n", workspace, targetContent.Workspace)
-		return workspace
+			"different from workspace '%v' found in state file(s).\n", workspaceFlag, targetContent.Workspace)
+		return workspaceFlag
 	}
 	return targetContent.Workspace
 }
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -26,9 +26,9 @@ var (
 	assumeYes  bool
 )
 
-func performValidate(ctx context.Context, kongClient *kong.Client, entity string) (bool, error) {
-	endpoint := fmt.Sprintf("/schemas/%s/validate", entity)
-	req, err := kongClient.NewRequest("POST", endpoint, nil, nil)
+func performValidate(ctx context.Context, kongClient *kong.Client, entity interface{}, entityType string) (bool, error) {
+	endpoint := fmt.Sprintf("/schemas/%s/validate", entityType)
+	req, err := kongClient.NewRequest("POST", endpoint, nil, entity)
 	if err != nil {
 		return false, err
 	}
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -26,15 +26,16 @@ var (
 	assumeYes  bool
 )
 
-func performValidate(ctx context.Context, kongClient *kong.Client, entity interface{}, entityType string) (bool, error) {
+func validateEntity(ctx context.Context, kongClient *kong.Client, entityType string, entity interface{}) (bool, error) {
+	errWrap := "validate entity '%s': %s"
 	endpoint := fmt.Sprintf("/schemas/%s/validate", entityType)
-	req, err := kongClient.NewRequest("POST", endpoint, nil, entity)
+	req, err := kongClient.NewRequest(http.MethodPost, endpoint, nil, entity)
 	if err != nil {
-		return false, err
+		return false, fmt.Errorf(errWrap, entityType, err)
 	}
 	resp, err := kongClient.Do(ctx, req, nil)
 	if err != nil {
-		return false, err
+		return false, fmt.Errorf(errWrap, entityType, err)
 	}
 	return resp.StatusCode == http.StatusOK, nil
 }
--- a/cmd/validate.go
+++ b/cmd/validate.go
@@ -118,6 +118,12 @@ func validateEntities(ctx context.Context, obj interface{}, kongClient *kong.Cli
 
 func validateWithKong(cmd *cobra.Command, ks *state.KongState) []error {
 	ctx := cmd.Context()
+	// make sure we are able to connect to Kong
+	_, err := fetchKongVersion(ctx, rootConfig)
+	if err != nil {
+		return []error{fmt.Errorf("couldn't fetch Kong version: %w", err)}
+	}
+
 	kongClient, err := utils.GetKongClient(rootConfig)
 	allErr := []error{}
 	if err != nil {
--- a/cmd/validate.go
+++ b/cmd/validate.go
@@ -1,16 +1,13 @@
 package cmd
 
 import (
-	"context"
 	"fmt"
 	"os"
-	"reflect"
-	"sync"
 
 	"github.com/kong/deck/file"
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by hbagdi** [comment]
So the goal of #190 is to do validation of data within the yaml file.

For example, the following is a valid input as per `deck validate` :

```
services:
- name: svc1
  host: mockbin.org:80
  tags:
  - team-svc1
```

But if you try to do `deck sync`, you get the following:

```
Error: 1 errors occurred:
        while processing event: {Create} service svc1 failed: HTTP status 400 (message: "schema violation (host: must not have a port)")
```

The goal here is to take the entities in the file and then validate them against Kong's Admin API.
See comments for more direction.

File: cmd/common.go [hunk 1]
```diff
@@ -26,6 +26,19 @@ var (
 	assumeYes  bool
 )
 
+func performValidate(ctx context.Context, kongClient *kong.Client, entity string) (bool, error) {
+	endpoint := fmt.Sprintf("/schemas/%s/validate", entity)
+	req, err := kongClient.NewRequest("POST", endpoint, nil, nil)
```
hbagdi, line 31: So you are making the right API call here but instead, you need to pass the entity as body.
I recommend playing with the endpoint directly in Kong to get a better feel of the API in Kong: https://docs.konghq.com/enterprise/2.5.x/admin-api/#validate-a-configuration-against-a-schema

And then write more code here to use it correctly.

cmd/validate.go
```diff
@@ -74,4 +125,9 @@ func init() {
 		"state", "s", []string{"kong.yaml"}, "file(s) containing Kong's configuration.\n"+
 			"This flag can be specified multiple times for multiple files.\n"+
 			"Use '-' to read from stdin.")
+	validateCmd.Flags().BoolVar(&validateCmdKongAPI, "use-kong-api",
+		false, "whether to leverage Kong's API to validate configuration.")
+	validateCmd.Flags().StringSliceVarP(&validateCmdKongAPIEntity,
+		"entity", "e", []string{}, "entities to be validated via Kong API "+
+			"(e.g. -e entity1 -e entity2)")
```
hbagdi, line 132: That's us asking the user to make one more decision. Are there any specific gains that you see here with this option?
A simpler product experience would be:
- deck validate -> validates using only the json schema in place (what decK does today in main branch)
- deck validate --use-kong-api (we will need to explore a better name for the flag) ->  this validates the entities within the file against Kong's Admin API. See the example in the comment I left as to what validation means in this context.

Thoughts?

GGabriele: I totally see your point! I got misleaded when I read the docs in the first place as I hadn't realize the entity should have been passed in the body, which led me to this poor implementation. Thanks for the pointers, I'll rework this!

GGabriele:
@hbagdi thanks for the feedbacks and tips! I just pushed some changes.

Sample input:

```
services:
- name: svc1
  host: mockbin.org:80
  tags:
  - team-svc1
  routes:
  - name: r1
    https_redirect_status_code: 301
    paths:
    - /r1
certificates:
- id: 13c562a1-191c-4464-9b18-e5222b46035b
  cert: |
    -----BEGIN CERTIFICATE-----
    -----END CERTIFICATE-----
  key: |
    -----BEGIN PRIVATE KEY-----
    -----END PRIVATE KEY-----
  tags:
  - cloudops-managed
  snis:
  - name: demo1.example.com:123
  - name: demo2.example.com
  - name: demo3.example.com
plugins:
- name: prometheus
  enabled: true
  protocols:
  - httpss
  - https
 ```

Sample output:

```
$ deck validate --use-kong-api
services: HTTP status 400 (message: "schema violation (host: must not have a port)")
certificates: HTTP status 400 (message: "2 schema violations (cert: invalid certificate: x509.new: asn1/tasn_dec.c:309:error:0D07803A:asn1 encoding routines:asn1_item_embed_d2i:nested asn1 error; key: invalid key: pkey.new:load_key: asn1/tasn_dec.c:309:error:0D07803A:asn1 encoding routines:asn1_item_embed_d2i:nested asn1 error)")
plugins: HTTP status 400 (message: "schema violation (protocols.1: expected one of: grpc, grpcs, http, https, tcp, tls, udp)")
snis: HTTP status 400 (message: "schema violation (name: must not have a port)")
```

File: cmd/common.go [hunk 3]
```diff
@@ -26,6 +26,19 @@ var (
...
 
+func performValidate(ctx context.Context, kongClient *kong.Client, entity interface{}, entityType string) (bool, error) {
+	endpoint := fmt.Sprintf("/schemas/%s/validate", entityType)
+	req, err := kongClient.NewRequest("POST", endpoint, nil, entity)
+	if err != nil {
+		return false, err
+	}
+	resp, err := kongClient.Do(ctx, req, nil)
+	if err != nil {
+		return false, err
```
hbagdi, line 37: Please wrap this error with some contextual info using fmt.Errorf.
fmt.Errorf("validate entity '%s': %s",entittyType,err)

cmd/common.go
```diff
@@ -26,6 +26,19 @@ var (
 	assumeYes  bool
 )
 
+func performValidate(ctx context.Context, kongClient *kong.Client, entity interface{}, entityType string) (bool, error) {
+	endpoint := fmt.Sprintf("/schemas/%s/validate", entityType)
+	req, err := kongClient.NewRequest("POST", endpoint, nil, entity)
```
hbagdi, line 31:
```suggestion
	req, err := kongClient.NewRequest(http.MethodPost, endpoint, nil, entity)
```

main.go
```diff
@@ -18,6 +18,7 @@ func registerSignalHandler() context.Context {
 	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
 
 	go func() {
+		defer signal.Stop(sigs)
```
hbagdi, line 21: I didn't know this was required. Today I learned! Thanks!

cmd/validate.go
```diff
@@ -74,4 +141,6 @@ func init() {
 		"state", "s", []string{"kong.yaml"}, "file(s) containing Kong's configuration.\n"+
 			"This flag can be specified multiple times for multiple files.\n"+
 			"Use '-' to read from stdin.")
+	validateCmd.Flags().BoolVar(&validateCmdKongAPI, "use-kong-api",
+		false, "perform schema validation against Kong API.")
```
hbagdi, line 145: Let's call this flag `--online` instead.

cmd/validate.go
```diff
@@ -66,6 +79,60 @@ this command.
...
+	validateEntities(ctx, ks.Documents, kongClient, "documents")
+	validateEntities(ctx, ks.HMACAuths, kongClient, "hmacauth_credentials")
+	validateEntities(ctx, ks.JWTAuths, kongClient, "jwt_secrets")
+	validateEntities(ctx, ks.KeyAuths, kongClient, "keyauth_credentials")
+	validateEntities(ctx, ks.Oauth2Creds, kongClient, "oauth2_credentials")
+	validateEntities(ctx, ks.Plugins, kongClient, "plugins")
+	validateEntities(ctx, ks.Routes, kongClient, "routes")
+	validateEntities(ctx, ks.SNIs, kongClient, "snis")
+	validateEntities(ctx, ks.Targets, kongClient, "targets")
+	validateEntities(ctx, ks.Upstreams, kongClient, "upstreams")
```
hbagdi, line 132: RBAC resources are missing here. Can you include those?

cmd/validate.go
```diff
@@ -66,6 +79,60 @@ this command.
 	},
 }
 
+func validate(ctx context.Context, entity interface{}, kongClient *kong.Client, entityType string) string {
+	_, err := performValidate(ctx, kongClient, entity, entityType)
+	output := ""
+	if err != nil {
+		output = color.New(color.FgRed, color.Bold).Sprintf("%s: %v", entityType, err)
```
hbagdi, line 86: use the `cprint` package, which wraps the color package with some additional behavior.

cmd/validate.go
```diff
@@ -66,6 +79,60 @@ this command.
...
+
+	values := reflect.ValueOf(entities)
+	var wg sync.WaitGroup
+	wg.Add(values.Len())
+	for i := 0; i < values.Len(); i++ {
+		go func(i int) {
+			defer wg.Done()
+			output := validate(ctx, values.Index(i).Interface(), kongClient, entityType)
+			if output != "" {
+				fmt.Println(output)
```
hbagdi, line 104: Don't print to stdout but instead collect all of these as errors and bubble up all the way and then print them in the RunE fn.

cmd/validate.go
```diff
@@ -66,6 +79,60 @@ this command.
...
+	if err != nil {
+		output = color.New(color.FgRed, color.Bold).Sprintf("%s: %v", entityType, err)
+	}
+	return output
+}
+
+func validateEntities(ctx context.Context, obj interface{}, kongClient *kong.Client, entityType string) {
+	// call GetAll on entity
+	method := reflect.ValueOf(obj).MethodByName("GetAll")
+	entities := method.Call([]reflect.Value{})[0].Interface()
```
hbagdi, line 94: Since there are no tests, we will need to figure out a way to catch an error here.
Can you use `init()` function to invoke code like this to ensure that it doesn't error/panic?

hbagdi (commented):
The approach is correct here. We know need to polish the code to get to merge.

cmd/common.go
GGabriele: Since there are two places where we may return an error here, I was wrapping it with some context on the caller side:

```
_, err := performValidate(ctx, kongClient, entity, entityType)
...
output = color.New(color.FgRed, color.Bold).Sprintf("%s: %v", entityType, err)
```

If you prefer to have it here though, that's fine to me!

cmd/validate.go
GGabriele: Ops, missed this...what is/are the db entities associated to RBAC resources? I couldn't find them :\

cmd/validate.go
GGabriele: Not sure if I got you here, can you elaborate a bit please?

We can also do a bit more of checks like this:

```
method := reflect.ValueOf(obj).MethodByName("GetAll")
	if !method.IsValid() {
		// bla
	}
	entities := method.Call([]reflect.Value{})[0].Interface()
	values := reflect.ValueOf(entities)
	if values.Kind() != reflect.Slice {
		// bla
	}
```

cmd/validate.go
GGabriele: I'm not 100% sure my last change is what you were envisioning, but let me know otherwise!

cmd/validate.go
```diff
@@ -66,6 +84,159 @@ this command.
...
+
+	// check if workspace exists
+	workspaceName := getWorkspaceName(validateWorkspace, targetContent)
+	workspaceExists, err := workspaceExists(ctx, rootConfig, workspaceName)
+	if err != nil {
+		return []error{err}
+	}
+	if !workspaceExists {
+		return []error{fmt.Errorf("workspace doesn't exist: %s", workspaceName)}
+	}
```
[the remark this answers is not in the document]
hbagdi: @GGabriele Please open an issue to track this problem. Can skip for now.

cmd/validate.go
```diff
@@ -66,6 +84,205 @@ this command.
...
+		return []error{fmt.Errorf("couldn't fetch Kong version: %w", err)}
+	}
+
+	// check if workspace exists
+	workspaceName := getWorkspaceName(validateWorkspace, targetContent)
+	workspaceExists, err := workspaceExists(ctx, rootConfig, workspaceName)
+	if err != nil {
+		return []error{err}
+	}
+	if !workspaceExists {
```
hbagdi, line 137: What if decK is used against the open-source version of Kong (which has no concept of workspace)?
All workspace specific code must be executed only when the user supplies a workspace.

hbagdi (changes requested):

cmd/validate.go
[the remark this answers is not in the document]
GGabriele: https://github.com/Kong/deck/issues/563

GGabriele:
@hbagdi I believe I addressed all your comments. Let me know what you think or if I missed something!

utils/utils.go
```diff
@@ -52,3 +53,16 @@ func NameToFilename(name string) string {
...
 	return strings.ReplaceAll(filename, url.PathEscape(string(os.PathSeparator)), string(os.PathSeparator))
 }
+
+func CallGetAll(obj interface{}) (reflect.Value, error) {
+	// call GetAll method on entity
+	var result reflect.Value
+	method := reflect.ValueOf(obj).MethodByName("GetAll")
+	if !method.IsValid() {
+		return result, fmt.Errorf("GetAll() method not found for %v. "+
+			"Please file a bug with Kong Inc", reflect.ValueOf(obj).Type())
```
hbagdi, line 63:
```suggestion
		return result, fmt.Errorf("GetAll() method not found for type '%v'. "+
			"Please file a bug with Kong Inc", reflect.ValueOf(obj).Type())
```

cmd/validate.go
```diff
@@ -74,4 +178,19 @@ func init() {
...
+	validateCmd.Flags().StringVarP(&validateWorkspace, "workspace", "w",
+		"", "validate configuration of a specific workspace "+
+			"(Kong Enterprise only).\n"+
+			"This takes precedence over _workspace fields in state files.")
+	validateCmd.Flags().IntVar(&validateParallelism, "parallelism",
+		10, "Maximum number of concurrent requests to Kong.")
+
+	if err := ensureGetAllMethods(); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
```
hbagdi, line 194:
```suggestion
		panic(err.Error())
```

validate/validate.go
```diff
@@ -0,0 +1,163 @@
...
+	parallelism       int
+	rbacResourcesOnly bool
+}
+
+func NewValidator(
+	ctx context.Context,
+	ks *state.KongState,
+	client *kong.Client,
+	parallelism int,
+	rbacResourcesOnly bool,
```
hbagdi, line 27: Instead create a ValidaterOpt struct and then take that struct as input.

```go
type ValidatorOpts struct {
.
.

}

func NewValidator(opt ValidatorOpt) {
.
.
}
 ```

File: validate/validate.go [hunk 16]
```diff
@@ -0,0 +1,163 @@
...
+}
+
+type ErrorsWrapper struct {
+	Errors []error
+}
+
+func (v ErrorsWrapper) Error() string {
+	var errStr string
+	for _, e := range v.Errors {
+		errStr += fmt.Sprintf("%s\n", e.Error())
```
hbagdi, line 45: Do not add a print after the last error in the array.

utils/utils.go
GGabriele: Done.

cmd/validate.go
GGabriele: Done.

validate/validate.go
GGabriele: Done.

validate/validate.go
GGabriele: Done.

hbagdi (approved):

PR #507, opened 2021-10-19
Automatically release on tag push

rainest:
Adds tag push CI jobs that run goreleaser, upload dist output, and build/push images to Docker Hub.

This will absolve us of the need to generate temporary kongbot tokens to push to the Kong Docker Hub. goreleaser has a good simple action also, so may as well add that too to automate the rest of the release.

The goreleaser job uploads dist output to the action artifacts for handling Homebrew releases.

This **does not work yet**. It requires HARRY_DOCKER_USERNAME, HARRY_DOCKER_TOKEN, DOCKER_USERNAME, and DOCKER_TOKEN secrets to actually push to Docker Hub. These should have access to Harry's registry and the Kong registry, respectively.

Example results: https://github.com/rainest/deck/runs/3943012130

```
11:21:46-0700 yagody $ docker run traines/altdeck:v1.8.11 version
decK traines/deck:v1.8.11 (a3c154682ba932e21e4510e20ebe2b247f90a3e4) 
```

This conditions `latest` tagging on the major version, based on the suggestion from https://github.com/docker/metadata-action/issues/80. I think this is probably reasonable, as I don't think we've historically ever done patch releases for older minor versions here (Kong itself is a different story), and would likely only do backports for major versions in the future. If that's true, this is a simple way to track the last release of the current major version with some infrequent manual work (you need to update CI when bumping the major version).

Our foundations repo has some more complicated tooling for determining latest by seeing if the current tag is higher than anything on Docker Hub, but it's difficult to use, as:
- That repo is currently private. It's probably okay to publish some of it but IDK, we'd need to ask.
- It's designed as an all-in-one bash utility. We'd need to hack out bits and pieces to make it work with the rest of the actions tooling.
- We have no way of tracking upstream updates automatically if we import part of it here.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,97 @@
+name: release
+
+on:
+  push:
+    tags:
+      - 'v*.*'
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.17
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2
+        with:
+          # either 'goreleaser' (default) or 'goreleaser-pro'
+          distribution: goreleaser
+          version: latest
+          args: release --rm-dist
+        env:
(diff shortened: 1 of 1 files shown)


shaneutt (approved):

rainest:
No word back on the kongcloudpush access, but at worst we just still push that one manually; I have access with my account and can handle it more easily than the official repo. Would work whenever we add access later.

Issue #508, opened 2021-10-21
mTLS Kong Admin API authentication support


ludovic-pourrat:
This is a new feature request to support Kong mTLS authentication capabilities.

Kong API administration API can be secured via mTLS authentication by setting the following  NGINX directives,

``` 
    nginx_admin_proxy_ssl_verify: "on"
    nginx_admin_ssl_client_certificate: /etc/secrets/tls.crt
    nginx_admin_ssl_verify_client: "on"
    nginx_admin_ssl_verify_depth: 3
    nginx_admin_ssl_trusted_certificate: /etc/secrets/ca/ca-bundle.crt
``` 

Thus, Deck shroud provide a way to specify client certificates.

Feedback are welcome.

PR #509, opened 2021-10-21
mTLS Kong Admin API authentication support (#508)

ludovic-pourrat:
Feature request proposal related to mTLS Kong Admin API authentication support #508


cmd/root.go
```diff
@@ -161,6 +161,20 @@ func init() {
 		"Set a request timeout for the client to connect with Kong (in seconds).")
 	viper.BindPFlag("timeout",
 		rootCmd.PersistentFlags().Lookup("timeout"))
+
+	rootCmd.PersistentFlags().String("tls-client-cert", "",
```
GGabriele, line 165: Can you please rename this to `tls-client-cert-file`. That's more in line with what we do somewhere else in this repo.

cmd/root.go
```diff
@@ -161,6 +161,20 @@ func init() {
...
+		"Set the Kong's Admin TLS client certificate.\n"+
+			"This value can also be set using DECK_TLS_CLIENT_CERT "+
+			"environment variable. Must be used in conjunction with tls-client-key")
+	viper.BindPFlag("tls-client-cert",
+		rootCmd.PersistentFlags().Lookup("tls-client-cert"))
+
+	rootCmd.PersistentFlags().String("tls-client-key", "",
```
GGabriele, line 172: Same here, can you please append `-file`?

cmd/root.go
```diff
@@ -161,6 +161,20 @@ func init() {
...
+		"Set the Kong's Admin TLS client key.\n"+
+			"This value can also be set using DECK_TLS_CLIENT_KEY "+
+			"environment variable. Must be used in conjunction with tls-client-cert")
```
GGabriele, line 175: Since these 2 flags are expected to be used in conjunction, can you please add a check to verify that either both or none are passed, and error out if only one of them is used?

cmd/root.go
ludovic-pourrat: The expected content is a string representing the certificate, so would you rather think about passing a file instead ?
I will renamed then.

cmd/root.go
ludovic-pourrat: The expected content is a string representing the certificate, so would you rather think about passing a file instead ?
I will renamed then.

cmd/root.go
ludovic-pourrat: Sure, will do.

cmd/root.go
GGabriele: Oh, I misread it then, apologies. Would you mind supporting both ways then?

ludovic-pourrat: Sure will do.

cmd/root.go
ludovic-pourrat: Done.

cmd/root.go
ludovic-pourrat: Done.

ludovic-pourrat:
Do you have advises about testing ?

I will update to PR to add coverage.

cmd/root.go
```diff
@@ -161,6 +161,35 @@ func init() {
 		"Set a request timeout for the client to connect with Kong (in seconds).")
 	viper.BindPFlag("timeout",
 		rootCmd.PersistentFlags().Lookup("timeout"))
+
+	rootCmd.PersistentFlags().String("tls-client-cert", "",
+		"Set the Kong's Admin TLS client certificate.\n"+
```
hbagdi, line 166:
```suggestion
		"PEM-encoded TLS client certificate to use for authentication with Kong's Admin API.\n"+
```

cmd/root.go
```diff
@@ -161,6 +161,35 @@ func init() {
...
+			"This value can also be set using DECK_TLS_CLIENT_CERT "+
+			"environment variable. Must be used in conjunction with tls-client-key")
+	viper.BindPFlag("tls-client-cert",
+		rootCmd.PersistentFlags().Lookup("tls-client-cert"))
+
+	rootCmd.PersistentFlags().String("tls-client-cert-file", "",
+		"Path to the Kong's Admin TLS client certificate file.\n"+
```
hbagdi, line 173:
```suggestion
		"Path to the file containing TLS client certificate to use for authentication with Kong's Admin API.\n"+
```

Please wrap lines to ensure a pretty output.

cmd/root.go
```diff
@@ -161,6 +161,35 @@ func init() {
...
+			"This value can also be set using DECK_TLS_CLIENT_CERT_FILE "+
+			"environment variable. Must be used in conjunction with tls-client-key-file")
+	viper.BindPFlag("tls-client-cert",
+		rootCmd.PersistentFlags().Lookup("tls-client-cert-file"))
+
+	rootCmd.PersistentFlags().String("tls-client-key", "",
+		"Set the Kong's Admin TLS client key.\n"+
```
hbagdi, line 180:
```suggestion
		"PEM-encoded private key for the corresponding client certificate .\n"+
```

cmd/root.go
```diff
@@ -161,6 +161,35 @@ func init() {
...
+			"This value can also be set using DECK_TLS_CLIENT_KEY "+
+			"environment variable. Must be used in conjunction with tls-client-cert")
+	viper.BindPFlag("tls-client-key",
+		rootCmd.PersistentFlags().Lookup("tls-client-key"))
+
+	rootCmd.PersistentFlags().String("tls-client-key-file", "",
+		"Path to the Kong's Admin TLS client key file.\n"+
```
hbagdi, line 187:
```suggestion
		"Path to file containing the private key for the corresponding client certificate.\n"+
```

utils/types.go
```diff
@@ -135,6 +140,15 @@ func GetKongClient(opt KongClientConfig) (*kong.Client, error) {
 		tlsConfig.RootCAs = certPool
 	}
 
+	if opt.TLSClientCert != "" && opt.TLSClientKey != "" {
+		// Read the key pair to create certificate
+		cert, err := tls.LoadX509KeyPair(opt.TLSClientCert, opt.TLSClientKey)
```
hbagdi, line 145: The variables contain the cert and key themselves while the function expects path to files that contain the necessary bits.

cmd/root.go
```diff
@@ -214,6 +243,48 @@ func initConfig() {
...
+		(viper.GetString("tls-client-cert") != "" && viper.GetString("tls-client-key") == "") {
+		fmt.Printf("tls-client-cert and tls-client-key must be used in conjunction")
+		os.Exit(1)
+	}
+
+	if (viper.GetString("tls-client-cert-file") == "" && viper.GetString("tls-client-key-file") != "") ||
+		(viper.GetString("tls-client-cert-file") != "" && viper.GetString("tls-client-key-file") == "") {
+		fmt.Printf("tls-client-cert-file and tls-client-key-file must be used in conjunction")
+		os.Exit(1)
+	}
```
hbagdi, line 258: This code is really hard to read and reason about.
Here is an alternative:
1. Read certificate first (from the flag and then second priority to the file-based flag) and store in a variable.
2. Do the same for the key
3. If cert or key are exclusively empty, return an error.

hbagdi (changes requested):
This patch is a good addition but I don't think it does what it is supposed to do yet.

cmd/root.go
ludovic-pourrat: Ok will update accordingly.

utils/types.go
ludovic-pourrat: Right my bad, will fix asap. Thanks.

utils/types.go
```diff
@@ -97,6 +97,10 @@ type KongClientConfig struct {
 	Timeout int
 
 	CookieJarPath string
+
+	TLSClientCert string
```
hbagdi, line 101: Unnecessary string conversions. Please change both of these types to `[]byte` and update the code accordingly.

cmd/root.go
```diff
@@ -214,6 +243,43 @@ func initConfig() {
...
+	rootConfig.TLSClientCert = clientCertContent
+
+	clientKeyContent := ""
+
+	if viper.GetString("tls-client-key") != "" {
+		clientKeyContent = viper.GetString("tls-client-key")
+	} else if viper.GetString("tls-client-key-file") != "" {
+		fileContent, err := ioutil.ReadFile(viper.GetString("tls-client-key-file"))
+		if err != nil {
+			fmt.Printf("read file %q: %s", viper.GetString("tls-client-key-file"), err)
```
hbagdi, line 268: Get the value from viper once and use it in all places. Best done to avoid typos.

cmd/root.go
```diff
@@ -214,6 +243,43 @@ func initConfig() {
...
 	rootConfig.Timeout = (viper.GetInt("timeout"))
 
+	clientCertContent := ""
+
+	if viper.GetString("tls-client-cert") != "" {
+		clientCertContent = viper.GetString("tls-client-cert")
+	} else if viper.GetString("tls-client-cert-file") != "" {
+		fileContent, err := ioutil.ReadFile(viper.GetString("tls-client-cert-file"))
+		if err != nil {
+			fmt.Printf("read file %q: %s", viper.GetString("tls-client-cert-file"), err)
```
hbagdi, line 253: Same comment as

> Get the value from viper once and use it in all places. Best done to avoid typos.

cmd/root.go
```diff
@@ -214,6 +243,43 @@ func initConfig() {
...
+		}
+		clientKeyContent = string(fileContent)
+		clientKeyContent = strings.TrimRight(clientKeyContent, "\n")
+	}
+	rootConfig.TLSClientKey = clientKeyContent
+
+	if (rootConfig.TLSClientKey == "" && rootConfig.TLSClientCert != "") ||
+		(rootConfig.TLSClientKey != "" && rootConfig.TLSClientCert == "") {
+		fmt.Printf("tls-client-cert and tls-client-key / tls-client-cert-file and tls-client-key-file " +
+			"must be used in conjunction")
```
hbagdi, line 279:
```suggestion
		fmt.Printf("tls-client-cert and tls-client-key / tls-client-cert-file and tls-client-key-file " +
			"must be used in conjunction but only one was provided")
```

hbagdi (changes requested):
LGTM once these changes are taken care of.
@GGabriele can take care of the merge.

cmd/root.go
ludovic-pourrat: Will do.

cmd/root.go
ludovic-pourrat: Will do.

cmd/root.go
ludovic-pourrat: Will do.

utils/types.go
ludovic-pourrat: Will do.

hbagdi: That should have been a []byte as well but oh well, no biggie.

hbagdi (approved):

hbagdi:
@ludovic-pourrat Linter unhappy.

ludovic-pourrat:
Oups too quick sorry for this I will sort this mistake out 😂

hbagdi (approved):

PR #517, opened 2021-11-02
fix(sync) apply --select-tag to new entities

### LINKED ISSUE #362: tags specified with `--select-tag` are not added
when running a `deck sync` the tags used to select the entities from the Kong datastore, should also be added to the entities being added to Kong, but this doesn't happen.

So after a first sync;
```shell
deck sync --select-tag=tieske -s xxx.yaml
```
returns this summary (on a blank datastore):
```
Summary:
  Created: 44
  Updated: 0
  Deleted: 0
```
everything is added, but without a tag `tieske`.

The following command;
```shell
deck diff --select-tag=tieske -s xxx.yaml
```
should return a summary of doing nothing, but it actually returns;
```
Summary:
  Created: 44
  Updated: 0
  Deleted: 0
```
The cause being that the tag was not added, hence the 2nd command selects nothing and adds everything again.

rainest:
When syncing entities, use the tags from either --select-tag or state file metadata. Previously, sync only used file metadata for assigning tags to new resources, and --select-tag would only filter entities from the current state for comparison.

If the tags requested by --select-tag and file metadata differ, fail the sync.

Fix #362

This maybe needs some historical information about the development of `--select-tag` to say whether it's actually by design--this does address the issue raised by #362, but is a fairly significant change, so I wanted some confirmation that it does indeed appear to be an unintended omission rather than intentional for some reason.

Without this change, `--select-tag` only filters entities from the current state. Proposed entities will be compare to these, but they won't themselves get the indicated tag. This suggests that the original expected use of `--select-tag` was with files created by a `--select-tag` dump (which includes tag information in file metadata, which _was_ applied prior to this change), since using it without metadata tags resulted in several odd situations:
- As noted in the original issue, tags are not applied to created resources.
- Repeated syncs result in conflicts, since subsequent current states will not include those created resources; they will be filtered out.
- Attempting to sync against resources that already had the indicated tag will actually strip the tag: 
[stripped_tag.txt](https://github.com/Kong/deck/files/7464150/stripped_tag.txt) (bdeck has this patch, odeck does not).

None of these seem useful, so I'm guessing this was indeed intentional.

Though this changes deck's behavior, I'm unsure of any workflows that would reasonably rely on the original behavior: any I can think of would run into one of the traps above, which will prevent you from syncing after 1-2 runs.

Unit tests of `file.Get()` would require a Kong instance or robust mocking, so we're probably stuck on that without integration tests.

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -114,7 +114,7 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 				sort.Strings(dumpConfig.SelectorTags)
 				sort.Strings(targetContent.Info.SelectorTags)
 				if !reflect.DeepEqual(dumpConfig.SelectorTags, targetContent.Info.SelectorTags) {
-					return fmt.Errorf("different tags specified by state file and --select-tags differ")
+					return fmt.Errorf("tags specified in the state file and via --select-tags flag are different. decK expects tags to be specified in either via flag or via state file. In case both are specified, they must match.")
 				}
 			} else {
 				dumpConfig.SelectorTags = targetContent.Info.SelectorTags
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"reflect"
+	"sort"
 
 	"github.com/blang/semver/v4"
 	"github.com/kong/deck/cprint"
@@ -107,7 +109,17 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 	}
 
 	if targetContent.Info != nil {
-		dumpConfig.SelectorTags = targetContent.Info.SelectorTags
+		if len(targetContent.Info.SelectorTags) > 0 {
+			if len(dumpConfig.SelectorTags) > 0 {
+				sort.Strings(dumpConfig.SelectorTags)
+				sort.Strings(targetContent.Info.SelectorTags)
+				if !reflect.DeepEqual(dumpConfig.SelectorTags, targetContent.Info.SelectorTags) {
+					return fmt.Errorf("different tags specified by state file and --select-tags differ")
+				}
+			} else {
+				dumpConfig.SelectorTags = targetContent.Info.SelectorTags
+			}
+		}
 	}
 
 	// read the current state
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -137,7 +137,7 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 	rawState, err := file.Get(targetContent, file.RenderConfig{
 		CurrentState: currentState,
 		KongVersion:  parsedKongVersion,
-	})
+	}, dumpConfig)
 	if err != nil {
 		return err
 	}
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -108,18 +108,9 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 		return err
 	}
 
-	if targetContent.Info != nil {
-		if len(targetContent.Info.SelectorTags) > 0 {
-			if len(dumpConfig.SelectorTags) > 0 {
-				sort.Strings(dumpConfig.SelectorTags)
-				sort.Strings(targetContent.Info.SelectorTags)
-				if !reflect.DeepEqual(dumpConfig.SelectorTags, targetContent.Info.SelectorTags) {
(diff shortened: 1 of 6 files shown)


file/builder.go
```diff
@@ -40,7 +41,13 @@ func (b *stateBuilder) build() (*utils.KongRawState, *utils.KonnectRawState, err
 	b.konnectRawState = &utils.KonnectRawState{}
 
 	if b.targetContent.Info != nil {
-		b.selectTags = b.targetContent.Info.SelectorTags
+		if len(b.selectTags) > 0 {
+			if !reflect.DeepEqual(b.selectTags, b.targetContent.Info.SelectorTags) {
```
hbagdi, line 45: Compare these only if the tags are specified in the state file (there is a workflow possible where tags are only specified as CLI option).
Also, do a proper element-wise comparison rather than reflect, which will throw an error if the ordering is different.

hbagdi:
If you want to improve the code, here are the things that I'll recommend we should be doing:
- Ensure that tags are specified only in one place: either the cli flag or state file, both should not be present
- figure out the select-tags to honor in cmd package and then proceed forward in diff/file packages. They shouldn't be aware of this decision making at all


PR approved from a product perspective.
Please ensure this doesn't have any side-effect on KIC.

file/builder.go
rainest: 3e1a594

rainest:
> Ensure that tags are specified only in one place: either the cli flag or state file, both should not be present

IMO we shouldn't be this strict. Is there any problem if both are present but they're equal? Different sets are definitely bad, but this fails on that. If you're lazy (like me) and do something like `deck dump --select-tags foo --kong-addr example.com --headers kong-admin-token:foo` and then reverse search/edit to `deck sync --select-tags foo --kong-addr example.com --headers kong-admin-token:foo` to not have to rewrite other flags, I don't think we should require you remove the select-tags unless it'd actually present an issue. Naive users may also not even know about the state file metadata and may reasonably assume that they should always set select-tags.

> figure out the select-tags to honor in cmd package and then proceed forward in diff/file packages. They shouldn't be aware of this decision making at all

b1e68df if I understand this right. Go figure, we had the same logic for using the state tag instead also elsewhere already.

hbagdi:
> IMO we shouldn't be this strict. Is there any problem if both are present but they're equal?

Fair. We must fail hard if both are non-empty and unequal.

cmd/common.go
```diff
@@ -107,7 +109,17 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
...
+			if len(dumpConfig.SelectorTags) > 0 {
+				sort.Strings(dumpConfig.SelectorTags)
+				sort.Strings(targetContent.Info.SelectorTags)
+				if !reflect.DeepEqual(dumpConfig.SelectorTags, targetContent.Info.SelectorTags) {
+					return fmt.Errorf("different tags specified by state file and --select-tags differ")
+				}
+			} else {
+				dumpConfig.SelectorTags = targetContent.Info.SelectorTags
+			}
+		}
```
hbagdi, line 122: move this into a function and unit test that function.

cmd/common.go
```diff
@@ -107,7 +109,17 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 	}
 
 	if targetContent.Info != nil {
-		dumpConfig.SelectorTags = targetContent.Info.SelectorTags
+		if len(targetContent.Info.SelectorTags) > 0 {
+			if len(dumpConfig.SelectorTags) > 0 {
+				sort.Strings(dumpConfig.SelectorTags)
+				sort.Strings(targetContent.Info.SelectorTags)
+				if !reflect.DeepEqual(dumpConfig.SelectorTags, targetContent.Info.SelectorTags) {
+					return fmt.Errorf("different tags specified by state file and --select-tags differ")
```
hbagdi, line 117:
```suggestion
					return fmt.Errorf("tags specified in the state file and via --select-tags flag are different. decK expects tags to be specified in either via flag or via state file. In case both are specified, they must match.")
```

Take care of line-wrapping.

@Tieske This cool?

cmd/common_test.go
```diff
@@ -0,0 +1,71 @@
...
+	}{
+		{
+			name: "both present and equal",
+			args: args{
+				dumpConfig:  dump.Config{SelectorTags: []string{"foo"}},
+				fileContent: file.Content{Info: &file.Info{SelectorTags: []string{"foo"}}},
+			},
+			want:    []string{"foo"},
+			wantErr: false,
+		},
```
hbagdi, line 30: add a test case where both are present and equal but ordering is different.

hbagdi:
Consider PR approved with the added test case.
We desperately need more testing to have confidence in such a large change.
cc @rajkong

PR #520, opened 2021-11-05
feat(reader) handle _transform field

### LINKED ISSUE #510: Error:  (root): Additional property _transform is not allowed on decK sync
I'm using deck 1.8.2 and latest kong OSS version 2.6.0 in DB-less mode.
After pulling down the default YAML declarative definition, that just include the 2 following directives, decK fail synching with the error: 

        (root): Additional property _transform is not allowed

------------------------------ This is my kong.yaml ------------------

```
_format_version: "2.1"

# _transform is optional, defaulting to true.
# It specifies whether schema transformations should be applied when importing this file
# as a rule of thumb, leave this setting to true if you are importing credentials
# with plain passwords, which need to be encrypted/hashed before storing on the database.
# On the other hand, if you are reimporting a database with passwords already encrypted/hashed,
# set it to false.

_transform: true

# Each Kong entity (core entity or custom entity introduced by a plugin)
# can be listed in the top-level as an array of objects:

# ...


...

### PR DESCRIPTION
**Pull request by rainest**
Handle the Kong state _transform field, which indicates whether data should pass through applicable transforms when imported into Kong. As deck passes data through the admin API, it will always pass through transforms. deck rejects the state if _transform=false, as it cannot satisfy this request.

AFAIK, we need take no additional action for `true`, and that is the default per https://docs.konghq.com/gateway-oss/2.6.x/db-less-and-declarative-config/#the-declarative-configuration-format. Rejecting `false` outright instead of warning since that should indicate that the file contains transformed/encrypted content, and running it through transforms again will not result in the intended config.

Fix #510

### CODE REVIEW

**Comment by rainest** [comment]
Testing the positive case for `Get()` fails due to not having a Kong instance, but

```
11:32:38-0700 esenin $ dlv debug -- validate -s foo.yaml 
Type 'help' for list of commands.
(dlv) c
Process 3923781 has exited with status 0
(dlv) exit
11:33:10-0700 esenin $ dlv debug -- diff -s foo.yaml        
Type 'help' for list of commands.
(dlv) c
creating certificate 13c562a1-191c-4464-9b18-e5222b46035b
creating consumer yolo
creating consumer harry
creating upstream upstream1
creating target 198.51.100.11:80 for upstream 42156d4b-b339-457c-9868-60f7636b7e27
creating target 198.51.100.12:80 for upstream 42156d4b-b339-457c-9868-60f7636b7e27
creating service svc1
creating key-auth n90ts for consumer 6e6590a5-d25d-4f1c-b265-9ee6097e5bd4
creating sni demo3.example.com
creating target 198.51.100.13:80 for upstream 42156d4b-b339-457c-9868-60f7636b7e27
creating sni demo2.example.com
creating service svc2
creating sni demo1.example.com
creating service svc3
Warning: import/export of basic-authcredentials using decK doesn't work due to hashing of passwords in Kong.
creating basic-auth user1 for consumer 6e6590a5-d25d-4f1c-b265-9ee6097e5bd4
creating hmac-auth hmac-user for consumer 6e6590a5-d25d-4f1c-b265-9ee6097e5bd4
creating jwt-auth MKWeR0nu9OAUR9HrjpUG82Hbfz7ZXsIw for consumer 6e6590a5-d25d-4f1c-b265-9ee6097e5bd4
creating acl-group foo-group for consumer 6e6590a5-d25d-4f1c-b265-9ee6097e5bd4
creating route r3
creating route r2
creating route r1
creating plugin prometheus (global)
creating plugin rate-limiting for consumer 3488030a-387d-4865-b409-4216131a8a07
Summary:
  Created: 23
  Updated: 0
  Deleted: 0
Process 3924231 has exited with status 0
(dlv) exit
11:33:39-0700 esenin $ diff -u kong.yaml foo.yaml 
--- kong.yaml	2021-11-05 11:17:59.690516063 -0700
+++ foo.yaml	2021-11-05 11:32:38.433901529 -0700
@@ -1,3 +1,4 @@
+_transform: true
 _info:
   select_tags:
   - managed-by-deck
```

File: file/reader_test.go [hunk 1]
```diff
@@ -71,6 +72,46 @@ func TestReadKongStateFromStdinFailsToParseText(t *testing.T) {
...
+	var content bytes.Buffer
+	content.Write([]byte("_transform: false\nservices:\n- host: test.com\n  name: test service\n"))
+
+	tmpfile, err := ioutil.TempFile("", "example")
+	if err != nil {
+		panic(err)
+	}
+	defer os.Remove(tmpfile.Name())
+
+	if _, err := tmpfile.Write(content.Bytes()); err != nil {
```
mflendrich, line 89: `tmpfile.WriteString("_transform: ...")`

file/reader.go
```diff
@@ -54,6 +59,10 @@ func Get(fileContent *Content, opt RenderConfig) (*utils.KongRawState, error) {
 	builder.currentState = opt.CurrentState
 	builder.kongVersion = opt.KongVersion
 
+	if fileContent.Transform != nil && !*fileContent.Transform {
+		return nil, fmt.Errorf("_transform: false is not supported")
```
mflendrich, line 63: (similarly to another PR in KIC where we discussed this)

let's export a named error value:

```go
  var ErrorTransformFalseNotSupported = fmt.Errorf("...")
```

and use it in tests instead of diffing the error text

file/reader_test.go
```diff
@@ -71,6 +72,46 @@ func TestReadKongStateFromStdinFailsToParseText(t *testing.T) {
...
+	if _, err := tmpfile.Seek(0, 0); err != nil {
+		panic(err)
+	}
+
+	oldStdin := os.Stdin
+	defer func() { os.Stdin = oldStdin }() // Restore original Stdin
+
+	os.Stdin = tmpfile
+
+	c, err := GetContentFromFiles(filenames)
```
mflendrich, line 102: I'd argue that now is the moment when `GetContentFromFiles` deserves refactoring so that the concern of file/stdin reading gets separated from logic.

file/reader_test.go
rainest: f108d1b

rainest: 93af3c1d7f79dccd66e9661d855eea08e2791121

file/reader.go
rainest: 93af3c1d7f79dccd66e9661d855eea08e2791121

file/reader_test.go
rainest: I don't have huge issues with it--it makes the tests a bit odd, but this one was mostly just copy/paste from existing tests, hence some of the unnecessary stuff.

I don't think it should be in scope here: this is just a hotfix to make deck not freak out on something that can appear in Kong configs but essentially changes nothing about its behavior. If we want to change it, it should go into a new issue.

hbagdi:
@kikito Can you verify if the understanding of "_transform" here is indeed correct or not?
LGTM. Please get a +1 from someone in the gateway team on the assumption behind the PR.

hbagdi:
@rainest can you squash all your commits into a single commit? We are good to merge after that.

rainest:
@hbagdi auto-merge is now set to squash mode and I've removed irrelevant commit headers from the body. It'll merge once you approve.

file/reader_test.go
mflendrich: Here comes the usual "this is just a hotfix" vs "this is technical debt that makes hotfixes unnecessarily complex" debate :slightly_smiling_face: 

> If we want to change it, it should go into a new issue.

This way technical debt never gets resolved because we keep copy/pasting suboptimal patterns between tests/code sites and punt improvement of these patterns into unspecified future.

That said, understanding our limited capacity at this time, it's not easy to draw the line between code health fixes that are worth fixing when touched and those that aren't.

Let's consider this comment fully optional, but also discuss this issue in a retro; I think it's a quality problem that's gradually driving us into maintenance trouble.

mflendrich (approved):
LGTM but disabled auto-merge so as to prevent the discussion on comments from being missed.

PR #529, opened 2021-11-23
fix: return error when sync context ends

### LINKED ISSUE #528: Solver reports success when terminated early by context expiration
Initially reported internally in FTI-3014.

Syncers will [return an error](https://github.com/Kong/deck/blob/v1.7.0/diff/diff.go#L358-L361) if their [context expires](https://github.com/Kong/deck/blob/v1.7.0/diff/diff.go#L424-L435). This error can occur anywhere during the syncer's lifetime, e.g. partway through the [create/update/delete function](https://github.com/Kong/deck/blob/v1.7.0/diff/diff.go#L86-L90). When the context ends, syncing stops abruptly and some entities will not be synced.

Error handling afterward [excludes this error](https://github.com/Kong/deck/blob/v1.7.0/diff/diff.go#L438-L443). If there are no other errors, the syncer will return no errors, and the solver will return stats on what it did sync, but also report no errors even though some entities were not synced:

https://github.com/Kong/deck/blob/v1.7.0/solver/solver.go#L41
https://github.com/Kong/deck/blob/v1.7.0/solver/solver.go#L83

In the deck CLI, this probably never matters: the context is derived from the command, and if it ends, the process is terminating. In the ingress controller, however, this results in the solve operation apparently completing entirely successfully, even though it only completed partially successfully.

rainest:
If a sync context ends, return its Context.Err() in the list of sync errors. This informs downstream clients that the sync did not process all entities, even if there were no events in flight (which generate their own errors, as seen below) at the time the context ended. On the user end, this will look like:

```
11:52:27-0800 esenin $ /tmp/bdeck sync -s /tmp/foo.yaml
creating consumer foo
...
creating upstream foo.8080.svc
^Creceived interrupt , terminating...
Summary:
  Created: 17
  Updated: 0
  Deleted: 0
Error: 5 errors occurred:
	sync context ended, some entities were not synced: context canceled
	while processing event: {Create} upstream bar.8001.svc failed: making HTTP request: Put "http://localhost:8001/upstreams/3acc5ce0-6a3b-40d2-a63c-6ab73d1db221": context canceled
        ...
```

Fixes #528

### DIFF
--- a/diff/diff.go
+++ b/diff/diff.go
@@ -477,6 +477,7 @@ func (sc *Syncer) Run(ctx context.Context, parallelism int, d Do) []error {
 	var errs []error
 	select {
 	case <-ctx.Done():
+		errs = append(errs, fmt.Errorf("sync context ended, some entities were not synced: %w", ctx.Err()))
 	case err, ok := <-sc.errChan:
 		if ok && err != nil {
 			if err != errEnqueueFailed {
--- a/diff/diff.go
+++ b/diff/diff.go
@@ -477,7 +477,7 @@ func (sc *Syncer) Run(ctx context.Context, parallelism int, d Do) []error {
 	var errs []error
 	select {
 	case <-ctx.Done():
-		errs = append(errs, fmt.Errorf("sync context ended, some entities were not synced: %w", ctx.Err()))
+		errs = append(errs, fmt.Errorf("failed to sync all entities: %w", ctx.Err()))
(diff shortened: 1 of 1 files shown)


hbagdi (dismissed):
I'm not certain if this will indeed solve the issue or not but is minor enough to give it a try.

hbagdi (approved):

PR #535, opened 2021-12-01
docs(*) add 1.9.0 changes

rainest:
Add 1.8.3 changes and release instructions.

Small, but we haven't had a patch release in quite a while, and the error return is something we want in KIC sooner or later.

I also want to confirm the new CI stuff works in the actual deck repo and confirm if there's something I forgot that still requires manual work.

Added release instructions since we'd apparently forgotten to ever take them out of a Google Doc, and they've been stable for a while now.


CHANGELOG.md
```diff
@@ -33,6 +34,19 @@
...
 
+## [v1.8.3]
+
+> Release date: 2021/12/02
+
+### Fixed
+
+- `--select-tag` applies its tags to newly-created entities whether or not the
+  tag is also present in the state file metadata.
+  [#517](https://github.com/Kong/deck/pull/517)
```
mflendrich, line 45: Isn't this a breaking change that would command more meaningful than a patch version bump? :thinking:

mflendrich (commented):
If I understand correctly, each of these changes modifies existing behavior that may potentially be relied upon (per [Hyrum's law](https://www.hyrumslaw.com/)).

Therefore I suggest that we bump the minor version number and release this as `1.9.0`, and replace "Fixed" with "Breaking changes" in the changelog.

CHANGELOG.md
rainest: "Isn't this a breaking change?" absent analysis of the potential breakage isn't of much use. The Hyrum article's advice is "be aware of possible implicit interfaces", not "every change in behavior is guaranteed breaking, you can never make patch releases ever". The latter reductio ad absurdum approach means we should declare any code change breaking. Fix a segfault? It's a breaking change, someone relied on that segfault occurring.

I don't think we have any good hypotheticals for this, so IMO labeling it breaking is more confusing than anything: messaging to users is that it's a breaking change because it's a change, but we aren't aware of any situations where it'd break anything, so we can't offer any advice on what you'd need to adjust beyond "run tests and see if they fail", which they should be doing anyway. Based on what we do know, this change does not break expected workflows, and we should mark it as such, i.e. our messaging should be to users that we do not expect any negative impact or changes required if you're coming from another 1.8.x release.

A workflow that relied on the existing behavior would need both an intimate understanding of the behavior and an intentionally contrived workflow to rely on it. The current behavior de facto makes `--select-tag` without file metadata useless for creating new entities. If you run it twice, the created entities are invisible and deck tries to create them unsuccessfully:

```
$ /tmp/bdeck sync -s /tmp/xxx.yaml --select-tag=foo
creating upstream awesome_learnservice
...

$ /tmp/bdeck sync -s /tmp/xxx.yaml --select-tag=foo
creating upstream awesome_learnservice
Summary:
  Created: 0
  Updated: 0
  Deleted: 0
Error: 1 errors occurred:
	while processing event: {Create} upstream awesome_learnservice failed: HTTP status 409 (message: "UNIQUE violation detected on '{name=\"awesome_learnservice\"}'")
```
Expected behavior (repeated syncs of the same content succeed and do not modify resources after the first sync) only works with `--select-tag` and no file metadata if tags are already present on all resources. This is unaffected by the change, since those resources already had tags listed in their individual YAML objects: 
[existing_tag.txt](https://github.com/Kong/deck/files/7644282/existing_tag.txt) (bdeck is 1.8.2, ndeck is this branch).

As best I can reason, to rely on the existing behavior, you'd need to intentionally either initially dump without `--select-tag` (so you get existing resources but no file tag metadata) or write a state file with no tag metadata from scratch, append new objects, sync with `--select-tag`, know that the newly-created objects would have no tags, and then run subsequent syncs without `--select-tag` to update those untagged objects. At that point, it's unclear why you wouldn't have run without `--select-tag` in the first place, because you'd wind up with the same result. Even if you did that, the breakage is that now your objects do have tags, which makes no difference--the sync you'd have run after, with no `--select-tag`, will manipulate them regardless, because syncs without tag criteria match any tag set.

The only (bizarre, and based on practical testing, also impossible) workflow I can think of where you could utilize the existing behavior and would be broken with this change is if you wanted to create a new object unconditionally if no matching object with the matching tag exists. Effectively:

1. You create a state file without tag metadata and some object `foo`. `foo` must not contain any unique fields (and by extension, its type's schema cannot have any required unique fields).
2. You run `deck sync --select-tag whatever`. If an object matching `foo` exists with tag `whatever`, deck updates that object with any changes needed, or does nothing if no changes are needed. If no object matching foo exists with tag `whatever`, deck creates a new object without the tag `whatever`.
3. Assuming the second case (new untagged object), subsequent runs with the same state and parameters will create additional `foo` objects, one for each run.

There are some types that match this schema constraint, e.g. you can create multiple routes with nothing other than `hostnames=[example.com]`. Routes don't require names or other unique fields (other than ID, which deck will generate automatically). However, you can't actually do this with deck: deck imposes additional constraints beyond the Kong schema:

```
$ /tmp/bdeck sync -s /tmp/wat.yaml --select-tag foo
Error: reading file: validating file content: 2 errors occurred:
	routes.0: Must validate at least one schema (anyOf)
	routes.0: name is required
```
deck does require at least one unique field on such objects intentionally to ensure that the entire Kong config in scope matches state; it's intentionally limited so that syncs are idempotent and cannot create multiple copies of a resource in state.

CHANGELOG.md
```diff
@@ -33,6 +34,19 @@
...
+- Timeouts in `Syncer.Run()` now return an error. Downstream clients using decK
+  as a library can determine when their sync attempt failed due to a timeout.
+  [#529](https://github.com/Kong/deck/pull/529)
```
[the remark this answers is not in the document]
rainest: https://github.com/Kong/deck/commit/cbd958a5cc3fb3f68c354de493a28487de348a3f to further explain what "failing silently" meant.

This is a bit murkier because there's an error where there wasn't before, but the successful return was a lie and indistinguishable from actual success based on `Syncher.Run()` alone (it returns nothing other than errors). You could possibly have checked your context to see if it had expired (still works, just no longer necessary) or independently verified the updated objects (extremely painful), both of which are a bit :thinking: if someone put them in place and didn't report an issue.

Actually relying on partial syncs seems unrealistic, since there's no way to control when in the sync process the timeout actually occurs: the subset of changes applied is essentially random, though it is ordered (deletes always come after updates and so on).

In practice, error handling should be one of:
- You ignore errors form `Syncher.Run()` entirely, and this change has no impact on you.
- You handle errors generically. There's an additional condition where you'll see an error, but arguably one where you do indeed want to see the error, because the requested action didn't complete successfully.
- You handle specific errors differently. There's now a new type of error you need to handle if you want to take some specific action based on it, but you were either ignoring unknown errors or had a catch-all case at the end.

This should furthermore affect library uses of deck only: the CLI can only pass in a command context, and that only exits when you quit. There is a cosmetic change since we [print the previously silenced error](https://github.com/Kong/deck/pull/529#issue-1061656372), but any screen-scraping error handling should still fall into one of the above categories. Library usage is currently [a bunch of KIC forks](https://github.com/Kong/deck/network/dependents?dependent_type=REPOSITORY), so there's not really anything for us to worry about there.

IMO based on that it falls on the non-breaking fix side of things. In the world of infinite possibilities we're potentially wrong, but based on what we know, those are very edgy edge cases if they exist, to the point that we'd need to discuss those problems with those individual users rather than being able to make a blanket "this will break your stuff, you must make changes" statement.

RELEASE.md
```diff
@@ -0,0 +1,54 @@
+eleasing a version
```
hbagdi, line 1: Why this file?

rainest: The file or the line? The file is because "a Google Doc somewhere that I can only find via browser history" is a bad location for maintainer docs.

The line is clipped-off copy paste accident, I'll fix that.

CHANGELOG.md
mflendrich: > I don't think we have any good hypotheticals for this, so IMO labeling it breaking is more confusing than anything

SGTM.

CHANGELOG.md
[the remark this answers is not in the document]
mflendrich: > In the world of infinite possibilities we're potentially wrong, but based on what we know, those are very edgy edge cases if they exist, to the point that we'd need to discuss those problems with those individual users rather than being able to make a blanket "this will break your stuff, you must make changes" statement.

I get your point but I disagree with this sentiment. The central question to me is: "are we making an interface change here?" and the answer seems to be "yes". Whether this interface change is benign (or even desirable) is another question, but (to me) a strictly less important one than whether the interface change is there in the first place.

Your impact analysis (_This should furthermore affect library uses of deck only_) indicates that the blast radius of this decision is minimal, so (while I generally expect that dependable parts of APIs remain fully backwards-compatible between patch versions), I don't think further investment into this specific versioning decision is worthwhile. That said, I do think that (going forward) API versioning deserves a stricter approach to semantic differences between versions.

CHANGELOG.md
mflendrich: > The latter reductio ad absurdum approach means we should declare any code change breaking. Fix a segfault? It's a breaking change, someone relied on that segfault occurring.

I disagree. Changes that do modify the public interface are breaking. A segfault is not a part of the public interface of deck; it's an unrecoverable runtime error.

> bizarre, and based on practical testing, also impossible

This is convincing. I wasn't sure about the impact of this change on decK's semantics, hence my question - I initially understood a scenario where someone would suffer from a tags mismatch handling difference to be more likely.

CHANGELOG.md
```diff
@@ -33,6 +34,20 @@
 - [v0.2.0](#v020)
 - [v0.1.0](#v010)
 
+## [v1.8.3]
+
+> Release date: 2021/12/02
```
mflendrich, line 39:
```suggestion
> Release date: 2021/12/08
```

RELEASE.md
```diff
@@ -0,0 +1,54 @@
+# Releasing a version
+
+## Prepare
+
+Ensure that your `go version` is at least 1.15.x.
```
mflendrich, line 5:
```suggestion
Ensure that your `go version` is not older than indicated in the header of `go.mod`.
```

`go.mod` says 1.16 today. Let's take a step towards de-duplicating the source of truth for version numbers.

RELEASE.md
```diff
@@ -0,0 +1,54 @@
...
+docker push kong/deck:$TAG
+
+
+# if also the latest release (not for a back-ported patch release):
+docker tag hbagdi/deck:$TAG hbagdi/deck:latest
+docker push hbagdi/deck:latest
+
+docker tag hbagdi/deck:latest kong/deck:latest
+docker push kong/deck:latest
+```
```
mflendrich, line 54: Do we still want to push `hbagdi/deck`?

mflendrich (commented):
Since last round I'm now convinced that the `select-tags` change can be deemed a fix; I don't feel very comfortable doing so for the semantic change of error handling, but (as discussed in the inline comment) the impact is minimal (this is becoming an academic discussion) and not worth agonizing over it, so let it be.

Some minor release procedure comments attached.

RELEASE.md
rainest: Open question for @hbagdi. I'm in favor of deprecating it and only pushing to kong/deck going forward. I don't think CI has access as requested in #507 

https://github.com/Kong/deck/pull/507#issuecomment-949963429
https://github.com/Kong/deck/issues/540#issuecomment-988309892

CHANGELOG.md
```diff
@@ -33,6 +34,31 @@
 - [v0.2.0](#v020)
 - [v0.1.0](#v010)
 
+## [v1.9.0]
+
+> Release date: 2021/12/08
```
mflendrich, line 39:
```suggestion
> Release date: 2021/12/09
```

mflendrich (dismissed):

mflendrich (approved):

PR #545, opened 2021-12-10
feat(Session cookie support for Auth to Admin Server)

rajkong:
Adding support for session cookie file in Netscape cookie file format to use in cookie jar. Depending on the User Agent that was used to generate the cookie file, the User-Agent header in deck may need to be over-ridden


utils/cookiejarparser.go
```diff
@@ -0,0 +1,114 @@
+package utils
+
+// Code from https://github.com/ssgelm/cookiejarparser under MIT License
```
hbagdi, line 3: Any reason we are copying this code rather than using it as a dependency?

utils/types.go
```diff
@@ -94,6 +94,11 @@ type KongClientConfig struct {
 	HTTPClient *http.Client
 
 	Timeout int
+
+	SessionFilePath string
```
hbagdi, line 98:
```suggestion
	CookieJarPath string
```
Cookies can do more than sessions.

.gitignore
```diff
@@ -24,6 +24,9 @@ _testmain.go
 *.test
 *.prof
 
+#IDE ignore
+.idea
+
```
hbagdi, line 29: We usually recommend to put these developer-specific files in your global .gitignore: https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/
Please remove this change.

cmd/root.go
```diff
@@ -123,6 +123,13 @@ func init() {
 	viper.BindPFlag("skip-workspace-crud",
 		rootCmd.PersistentFlags().Lookup("skip-workspace-crud"))
 
+	// Support for Session Cookie
+	rootCmd.PersistentFlags().String("kong-session-cookie-path", "",
```
hbagdi, line 127:
```suggestion
	rootCmd.PersistentFlags().String("kong-cookie-jar-path", "",
```

Let's avoid the word 'session' in here since the functionality is more generic than that.

cmd/root.go
```diff
@@ -123,6 +123,13 @@ func init() {
...
+		"Absolute path for the session-cookie file for auth with Admin Server.\n"+
+			"You may also need to pass in as header the User-Agent that was used to create the session.")
```
hbagdi, line 129: Update this text as well.
Please mention the format of the cookie file path (netscape cookie format I think).

Question: what happens if the user-agent is not specified?

utils/types.go
```diff
@@ -94,6 +94,11 @@ type KongClientConfig struct {
...
+
+	// Whether to initialize the Http client with a cookie jar or not
+	ISSessionClient bool
```
hbagdi, line 101: Is this required? Why not check if the SessionFilePath != ""?

utils/cookiejarparser.go
rajkong: Seemed very small and not sure if it would be maintained. So felt it is better to copy instead of import

cmd/root.go
rajkong: It gives a 401. This took me a while to figure out. Looks like Kong associates sessions and user-agent. My guess is that when Kong sees a different user agent, it assumes that the session was hijacked and sends a 401

utils/types.go
rajkong: Yeah, I could do that, felt that the boolean could satisfy multiple flags, so left it. I can remove it

.gitignore
rajkong: Removed

cmd/root.go
rajkong: Done

utils/types.go
rajkong: Done

utils/cookiejarparser.go
hbagdi: If we are not modifying any part of the dependency code, let's introduce the dependency.
Copying code in here is a new pattern and I'd recommend not deviating from current practices (other and future maintainers will wonder why this exception).

cmd/root.go
```diff
@@ -123,6 +123,13 @@ func init() {
 	viper.BindPFlag("skip-workspace-crud",
 		rootCmd.PersistentFlags().Lookup("skip-workspace-crud"))
 
+	// Support for Session Cookie
+	rootCmd.PersistentFlags().String("kong-cookie-jar-path", "",
+		"Absolute path for cookie-jar file in Netscape file format for auth with Admin Server.\n"+
```
hbagdi, line 128:
```suggestion
		"Absolute path to a cookie-jar file in the Netscape cookie format for auth with Admin Server.\n"+
```

utils/types.go
```diff
@@ -153,6 +155,14 @@ func GetKongClient(opt KongClientConfig) (*kong.Client, error) {
 	if opt.Workspace != "" {
 		url.Path = path.Join(url.Path, opt.Workspace)
 	}
+	// Add Session Cookie support if required
+	if opt.CookieJarPath != "" {
+		jar, err := LoadCookieJarFile(opt.CookieJarPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to initialize cookie-jar:%w", err)
```
hbagdi, line 162:
```suggestion
			return nil, fmt.Errorf("failed to initialize cookie-jar: %w", err)
```

cmd/root.go
hbagdi: Yikes. That may introduce additional friction that we can't solve for. Let's ship this and come back here if this PR doesn't achieve the desired outcome for our customers.

utils/cookiejarparser.go
rajkong: Done

utils/types.go
rajkong: Done

cmd/root.go
rajkong: done

utils/types.go
rajkong: done

Issue #468, opened 2021-07-27
deck konnect commands do not honor --headers flag


rainest:
Originally reported internally in FTI-2727.

Some Konnect test/development environments have additional authentication requirements (specifically basic-auth) beyond the Konnect login system.

The `--headers` flag should be able to add the additional auth headers, but cannot, as konnect commands just ignore it. We need to review how we [add headers to the client configuration](https://github.com/Kong/deck/blob/v1.7.0/cmd/root.go#L178) for other commands and make it apply to konnect commands also.

hbagdi:
https://github.com/Kong/deck/commit/6cf2b011d98bdc11928df41db6827ce904805622 fixes this one

PR #557, opened 2022-01-03
fix:deck konnect commands do not honor --headers flag #468

rajkong:
Enabling header flags for konnect commands


utils/types.go
```diff
@@ -193,9 +195,13 @@ func parseHeaders(headers []string) (http.Header, error) {
...
 	error) {
 	address := CleanAddress(config.Address)
+	headers, err := parseHeaders(config.Headers)
+	if err != nil {
+		return nil, fmt.Errorf("parsing headers: %w", err)
+	}
 	client, err := konnect.NewClient(httpClient, konnect.ClientOpts{
 		BaseURL: address,
-	})
+	}, headers)
```
hbagdi, line 204: Let's make changes only in this file and bring changes that are in `konnect/client.go` into this file. 
This way we can keep the "client with header" logic within this file and re-use it across the Kong and Konnect clients.

konnect/client.go
```diff
@@ -41,7 +41,7 @@ type ClientOpts struct {
 // NewClient returns a Client which talks to Konnect's API.
 func NewClient(httpClient *http.Client, opts ClientOpts) (*Client, error) {
 	if httpClient == nil {
-		httpClient = http.DefaultClient
+		return nil, fmt.Errorf("nil httpClient passed")
```
hbagdi, line 44: Please revert this.

hbagdi (approved):

Issue #25, opened 2019-06-21
Fetch and set defaults for plugins


hbagdi:
Currently, default values for plugins need to be hard-coded into the YAML file.
It would be nice if users can set only the values they need in the file and then the rest of the defaults is something deck figures out on it's own.

Kong Ingress Controller already does something similar to this.

Prashansa-K:
This is already completed.

PR #562, opened 2022-01-14
feat: populate plugin defaults from their schemas

GGabriele:
Currently `deck` doesn't automatically inject plugins' defaults, which means it keeps showing misleading diffs when running the `sync` subcommand, trying to remove the default values for both `config` and `protocols` properties:

```
$ cat kong.yaml
_format_version: "1.1"
plugins:
- config:
    day: null
    header_name: null
    hide_client_headers: false
    hour: 10
  enabled: true
  name: rate-limiting
- config:
    add:
      body: []
      headers: []
      querystring: []
    http_method: null
  enabled: true
  name: request-transformer
  protocols:
  - grpc
  - grpcs
  - http
  - https
```

```
$ deck sync
creating plugin rate-limiting (global)
creating plugin request-transformer (global)
Summary:
  Created: 2
  Updated: 0
  Deleted: 0

$ deck sync
updating plugin rate-limiting (global)  {
   "config": {
     "day": null,
-    "fault_tolerant": true,
     "header_name": null,
     "hide_client_headers": false,
     "hour": 10,
-    "limit_by": "consumer",
-    "minute": null,
-    "month": null,
-    "path": null,
-    "policy": "cluster",
-    "redis_database": 0,
-    "redis_host": null,
-    "redis_password": null,
-    "redis_port": 6379,
-    "redis_server_name": null,
-    "redis_ssl": false,
-    "redis_ssl_verify": false,
-    "redis_timeout": 2000,
-    "second": null,
-    "year": null
   },
   "enabled": true,
   "id": "bcaa11dd-73a5-44ad-bd7b-8de79c3e1ea7",
   "name": "rate-limiting",
-  "protocols": [
-    "grpc",
-    "grpcs",
-    "http",
-    "https"
-  ]
 }

updating plugin request-transformer (global)  {
   "config": {
     "add": {
       "body": [
       ],
       "headers": [
       ],
       "querystring": [
       ]
     },
-    "append": {
-      "body": [
-      ],
-      "headers": [
-      ],
-      "querystring": [
-      ]
-    },
     "http_method": null,
-    "remove": {
-      "body": [
-      ],
-      "headers": [
-      ],
-      "querystring": [
-      ]
-    },
-    "rename": {
-      "body": [
-      ],
-      "headers": [
-      ],
-      "querystring": [
-      ]
-    },
-    "replace": {
-      "body": [
-      ],
-      "headers": [
-      ],
-      "querystring": [
-      ],
-      "uri": null
-    }
   },
   "enabled": true,
   "id": "26d9af55-57d2-42e8-b121-1d80ccdafe19",
   "name": "request-transformer",
   "protocols": [
     "grpc",
     "grpcs",
     "http",
     "https"
   ]
 }

Summary:
  Created: 0
  Updated: 2
  Deleted: 0
```

What this PR does is adding the capability to inject the defaults values into plugins, pulling them from the Admin API `/schemas/plugin/$pluginID` endpoint. This works with "plain" `config` attributes like the case of the `rate-limiting` plugin, as well as with nested objects, like with `request-transformer`.

This currently **doesn't** work with Konnect, whose support for this will be added in a separate PR.

```
$ cat kong.yaml
_format_version: "1.1"
plugins:
- config:
    day: null
    header_name: null
    hide_client_headers: false
    hour: 10
  enabled: true
  name: rate-limiting
- config:
    add:
      body: []
      headers: []
      querystring: []
    http_method: null
  enabled: true
  name: request-transformer
  protocols:
  - grpc
  - grpcs
  - http
  - https
```

```
$ deck sync
creating plugin rate-limiting (global)
creating plugin request-transformer (global)
Summary:
  Created: 2
  Updated: 0
  Deleted: 0

$ deck sync
Summary:
  Created: 0
  Updated: 0
  Deleted: 0

$ deck diff
Summary:
  Created: 0
  Updated: 0
  Deleted: 0
```

Related to https://github.com/Kong/deck/pull/562 and https://github.com/Kong/deck/issues/25


file/builder.go
```diff
@@ -768,9 +786,101 @@ func (b *stateBuilder) ingestRoute(r FRoute) error {
...
+		for _, protocol := range defaultValue.([]interface{}) {
+			if protocolStr, ok := protocol.(string); ok {
+				p.Protocols = append(p.Protocols, &protocolStr)
+			}
+		}
+	}
+}
+
+func (b *stateBuilder) getDefaults(entity, entityID string) (map[string]interface{}, error) {
+	ctx := context.Background()
```
hbagdi, line 834: Plumb context from the command through and create a context with a timeout as defined by --timeout. No background or TODO context should exist in any code path.

file/builder.go
```diff
@@ -768,9 +786,101 @@ func (b *stateBuilder) ingestRoute(r FRoute) error {
...
+	if defaultValue, ok := protocols.(map[string]interface{})["default"]; ok {
+		for _, protocol := range defaultValue.([]interface{}) {
+			if protocolStr, ok := protocol.(string); ok {
+				p.Protocols = append(p.Protocols, &protocolStr)
+			}
+		}
+	}
+}
+
+func (b *stateBuilder) getDefaults(entity, entityID string) (map[string]interface{}, error) {
```
hbagdi, line 833:
```suggestion
func (b *stateBuilder) getDefaults(pluginName string) (map[string]interface{}, error) {
```

Simplify this function for plugin defaults.

file/builder.go
```diff
@@ -768,9 +786,101 @@ func (b *stateBuilder) ingestRoute(r FRoute) error {
...
+	if b.konnectRawState == nil ||
+		len(b.konnectRawState.Documents) > 0 ||
+		len(b.konnectRawState.ServicePackages) > 0 ||
+		b.client == nil {
+		return plugins, nil
+	}
+	pluginsWithDefault := []FPlugin{}
+	for _, p := range plugins {
+		p := p
+		defaults, err := b.getDefaults("plugins", *p.Name)
```
hbagdi, line 859: Please introduce caching of plugin defaults.
If there are 100 key-auth plugins defined by the user, this code will invoke a GET request for the same endpoint (and same response) 100 times. This will have a severe impact on execution time of decK.

file/builder.go
```diff
@@ -768,9 +786,101 @@ func (b *stateBuilder) ingestRoute(r FRoute) error {
...
+				if p.Protocols != nil {
+					continue
+				}
+				ingestDefaultProtocols(&p, protocols)
+			} else if config, ok := field.(map[string]interface{})["config"]; ok {
+				ingestDefaultConfigFields(&p, config.(map[string]interface{})["fields"])
+			}
+		}
+		pluginsWithDefault = append(pluginsWithDefault, p)
+	}
```
hbagdi, line 880: Since we are at it, please add a default of true for the plugin.Enabled field.

file/builder.go
```diff
@@ -768,9 +786,101 @@ func (b *stateBuilder) ingestRoute(r FRoute) error {
...
+	req, err := b.client.NewRequest(http.MethodGet, endpoint, nil, nil)
+	if err != nil {
+		return schema, err
+	}
+	_, err = b.client.Do(ctx, req, &schema)
+	return schema, err
+}
+
+func (b *stateBuilder) ingestPluginDefaults(plugins []FPlugin) ([]FPlugin, error) {
```
hbagdi, line 848: Separate out functions that get the defaults from Kong and functions that take these defaults and merge with the plugins.
The latter must be unit tested (blocker for merge).

file/builder.go
```diff
@@ -768,9 +786,101 @@ func (b *stateBuilder) ingestRoute(r FRoute) error {
...
+				innerField := map[string]interface{}{}
+				for _, field := range fields.([]interface{}) {
+					for fieldKey, value := range field.(map[string]interface{}) {
+						if defaultValue, ok := value.(map[string]interface{})["default"]; ok {
+							innerField[fieldKey] = defaultValue
+						} else {
+							innerField[fieldKey] = nil
+						}
+					}
+				}
```
hbagdi, line 814: Can we not recurse through? What if there are innerFields inside innerField?

hbagdi:
@GGabriele Midway reviewing this PR I realized that I had written this code in KIC: https://github.com/Kong/kubernetes-ingress-controller/blob/main/internal/deckgen/deckgen.go#L121-L122.

Please consult that piece of code. We must reuse the same code across KIC and decK. The common code can go in `kong/go-kong` and then both projects can use it.

@GGabriele @mflendrich @rainest Please work together on this since it lies across the boundary of KIC and decK.

GGabriele:
I tried implementing the logic from KIC here (https://github.com/Kong/deck/pull/562/commits/f1373b02fef4687d552e7ecdcee28a78ee6f743e), expanding the existing one to add support for default protocols as well. KIC currently uses the `GetSchema` method from `go-kong`, which uses the `/plugins/schema` endpoint, which only returns the `config` schema, omitting the `protocols` entry. That is included from the `/schema/plugins` endpoint instead, which I'm using here.

I guess at this point what we can do is the following:
- add a method in `go-kong` to support `/schema/plugins` as well
- add a method to fill both config and protocols defaults given a plugin
- use these new methods here in `deck` and `KIC`

Thoughts?

GGabriele:
I went ahead and opened [this PR](https://github.com/Kong/go-kong/pull/114) in `go-kong`. Once we agree on that and merge it, I'll update this PR as well.

cmd/konnect_sync.go
```diff
@@ -19,7 +21,9 @@ to get Konnect's state in sync with the input state.` + konnectAlphaState,
 			return fmt.Errorf("writing to stdout is not supported in Konnect mode")
 		}
 		_ = sendAnalytics("konnect-sync", "")
-		return syncKonnect(cmd.Context(), konnectDiffCmdKongStateFile, false,
+		ctx, cancel := context.WithTimeout(cmd.Context(), time.Second*time.Duration(rootConfig.Timeout))
+		defer cancel()
+		return syncKonnect(ctx, konnectDiffCmdKongStateFile, false,
```
hbagdi, line 26: This seems incorrect. This would imply that the whole sync process should finish in rootconfig.Timeout.

rootConfig.Timetout is a request-level timeout. The cmd.context() seems right, which will be canceled on a signal.

cmd/sync.go
```diff
@@ -22,7 +24,9 @@ var syncCmd = &cobra.Command{
 to get Kong's state in sync with the input state.`,
 	Args: validateNoArgs,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return syncMain(cmd.Context(), syncCmdKongStateFile, false,
+		ctx, cancel := context.WithTimeout(cmd.Context(), time.Second*time.Duration(rootConfig.Timeout))
+		defer cancel()
+		return syncMain(ctx, syncCmdKongStateFile, false,
```
hbagdi, line 29: Same comment as above.

file/builder.go
```diff
@@ -766,11 +784,47 @@ func (b *stateBuilder) ingestRoute(r FRoute) error {
 	var plugins []FPlugin
 	for _, p := range r.Plugins {
 		p.Route = &kong.Route{ID: kong.String(*r.ID)}
+		if err := b.ingestPluginDefaults(p); err != nil {
+			return err
+		}
```
hbagdi, line 789: It seems like all of the ingestPluginDefaults are followed by ingestPlugins.

Suggest an alternate concise code path:
- get rid off all existing ingentPluginDefaults() invocations
- add a single one inside ingestPlugins()

file/reader.go
```diff
@@ -60,12 +65,16 @@ func GetForKonnect(fileContent *Content, opt RenderConfig) (*utils.KongRawState,
 
 // Get process the fileContent and renders a RawState.
 // IDs of entities are matches based on currentState.
-func Get(fileContent *Content, opt RenderConfig, dumpConfig dump.Config) (*utils.KongRawState, error) {
+func Get(ctx context.Context, fileContent *Content, opt RenderConfig, dumpConfig dump.Config, wsClient *kong.Client) (
+	*utils.KongRawState, error,
```
hbagdi, line 69: This is an ever growing list. Could you please add a Github issue to take care of this in future?
We need to replace all the parameters with a "GetOpts" struct and then pass that around.

utils/utils.go
```diff
@@ -10,6 +10,10 @@ import (
 	"strings"
 )
 
+type Schema map[string]interface{}
+
+type Schemas map[string]Schema
```
hbagdi, line 15: These are unnecessary. Simply use the types as is where they are needed.
If you were adding methods onto these types, then it would make sense to pull them into a util package.

file/builder.go
```diff
@@ -766,11 +784,47 @@ func (b *stateBuilder) ingestRoute(r FRoute) error {
...
+		return err
+	}
+	p, err := kong.FillPluginsDefaults(&plugin.Plugin, schema)
+	if err != nil {
+		return err
+	}
+	plugin.Config = p.Config
+	plugin.Protocols = p.Protocols
+	plugin.Enabled = p.Enabled
+	return nil
```
hbagdi, line 825: Code smell. See https://github.com/Kong/go-kong/pull/114/files#r790030256

file/builder.go
```diff
@@ -766,11 +784,47 @@ func (b *stateBuilder) ingestRoute(r FRoute) error {
...
+	return schema, nil
+}
+
+func (b *stateBuilder) ingestPluginDefaults(plugin *FPlugin) error {
+	if b.client == nil {
+		return nil
+	}
+	schema, err := b.getPluginSchema(plugin.Name)
+	if err != nil {
+		return err
```
hbagdi, line 816:
```suggestion
		return fmt.Errorf("retrieve schema for "%v" from Kong: %v", *plugin.Name, err)
```

Annotate error where possible. You can skip this in lower-level library code but as the code gets "closer" to the user, do more and more of it. It contextualizes the error and makes it a bit more actionable for the user.

file/builder_test.go
```diff
@@ -2439,3 +2440,427 @@ func Test_stateBuilder_fillPluginConfig(t *testing.T) {
...
+	}
+)
+
+func TestFillPluginDefaults(T *testing.T) {
+	tests := []struct {
+		name     string
+		plugin   *kong.Plugin
+		schema   utils.Schema
+		expected *kong.Plugin
+	}{
```
hbagdi, line 2712: Remove these tests. They are now part of go-kong.

hbagdi (changes requested):

file/reader.go
GGabriele: https://github.com/Kong/deck/issues/567

file/builder.go
GGabriele: https://github.com/Kong/go-kong/pull/116

file/builder_test.go
GGabriele: done

file/builder.go
GGabriele: done

utils/utils.go
GGabriele: done

file/builder.go
GGabriele: done

file/builder.go
```diff
@@ -771,9 +778,38 @@ func (b *stateBuilder) ingestRoute(r FRoute) error {
...
+		return fmt.Errorf("retrieve schema for %v from Kong: %v", *plugin.Name, err)
+	}
+	return kong.FillPluginsDefaults(&plugin.Plugin, schema)
+}
+
 func (b *stateBuilder) ingestPlugins(plugins []FPlugin) error {
 	for _, p := range plugins {
 		p := p
+		if err := b.ingestPluginDefaults(&p); err != nil {
+			return err
```
hbagdi, line 811: wrap error
fmt.Errorf("add defaults to plugin '%v': %v", *p.Name, err)

cmd/validate.go
```diff
@@ -45,10 +45,9 @@ this command unless --online flag is used.
 		if err != nil {
 			return err
 		}
-
-		rawState, err := file.Get(targetContent, file.RenderConfig{
+		rawState, err := file.Get(cmd.Context(), targetContent, file.RenderConfig{
 			CurrentState: dummyEmptyState,
-		}, dump.Config{})
+		}, dump.Config{}, nil)
```
hbagdi, line 50: Question: Why not populate defaults for validate when online flag is in use?

GGabriele: Good catch, I started this PR before the online `validate` feature was merged and I forgot to add it back. I'll do it!

GGabriele: Added

file/builder.go
```diff
@@ -21,6 +22,11 @@ type stateBuilder struct {
 	selectTags   []string
 	intermediate *state.KongState
 
+	client *kong.Client
+	ctx    context.Context
+
+	schemas map[string]map[string]interface{}
```
hbagdi, line 28:
```suggestion
	schemasCache map[string]map[string]interface{}
```

file/builder.go
```diff
@@ -771,9 +778,38 @@ func (b *stateBuilder) ingestRoute(r FRoute) error {
 	return b.ingestPlugins(plugins)
 }
 
+func (b *stateBuilder) getPluginSchema(entityID *string) (map[string]interface{}, error) {
```
hbagdi, line 781:
```suggestion
func (b *stateBuilder) getPluginSchema(pluginName string) (map[string]interface{}, error) {
```
Take a string.
This code base deals with a lot of pointers but that's not what one should do in Go usually.

file/builder.go
```diff
@@ -771,9 +778,38 @@ func (b *stateBuilder) ingestRoute(r FRoute) error {
...
+	var schema map[string]interface{}
+
+	// lookup in cache
+	if schema, ok := b.schemas[*entityID]; ok {
+		return schema, nil
+	}
+	schema, err := b.client.Plugins.GetFullSchema(b.ctx, entityID)
```
hbagdi, line 788: What if client is nil?

hbagdi: This is possible with `validate` cmd, no?

GGabriele: This method is called by `ingestPluginDefaults`, which does a [check](https://github.com/Kong/deck/blob/DECK-8/file/builder.go#L797-L799) on whether `b.client` is `nil` or not

file/builder.go
GGabriele: done

file/builder.go
GGabriele: done

file/builder.go
```diff
@@ -771,9 +778,38 @@ func (b *stateBuilder) ingestRoute(r FRoute) error {
...
+	}
+	schema, err := b.client.Plugins.GetFullSchema(b.ctx, &pluginName)
+	if err != nil {
+		return schema, err
+	}
+	b.schemasCache[pluginName] = schema
+	return schema, nil
+}
+
+func (b *stateBuilder) ingestPluginDefaults(plugin *FPlugin) error {
```
hbagdi, line 796:
```suggestion
func (b *stateBuilder) addPluginDefaults(plugin *FPlugin) error {
```
"ingest" has a slightly different meaning in this file it seems, more about "ingesting the objects into state".

hbagdi:
Another way to implement this feature without exposing the kong "client" to the builder would have been to make all the defaulting logic part of utils.Defaulter package and then inject that in here. Makes testing and code management easier.
Don't do this in this PR now. We will refactor when we have problems.

file/builder.go
GGabriele: done

hbagdi (approved):

PR #564, opened 2022-01-18
feat: don't manage Konnect internal-only plugins

GGabriele:
Plugins created by Konnect for internal-use only
are tagged with 'konnect-managed-plugin'. decK should not allow users to
interact with any plugins with this tag.

This skips plugins with internal-only konnect tags.

### DIFF
--- a/dump/dump_konnect.go
+++ b/dump/dump_konnect.go
@@ -170,20 +170,23 @@ func kongServiceIDs(cpID string,
 	return res
 }
 
-// excludeKonnectManagedPlugins filter out konnect-managed plugins
+// excludeKonnectManagedPlugins filters out konnect-managed plugins
 func excludeKonnectManagedPlugins(plugins []*kong.Plugin) []*kong.Plugin {
-	var filtered []*kong.Plugin
+	filtered := []*kong.Plugin{}
 	for _, p := range plugins {
-		var konnectManaged bool
-		for _, t := range p.Tags {
-			if *t == konnect.KonnectManagedPluginTag {
-				konnectManaged = true
-			}
-		}
-		if konnectManaged {
+		if isManagedByKonnect(p) {
 			continue
 		}
 		filtered = append(filtered, p)
 	}
 	return filtered
+}
+
+func isManagedByKonnect(plugin *kong.Plugin) bool {
+	for _, t := range plugin.Tags {
+		if *t == konnect.KonnectManagedPluginTag {
+			return true
+		}
+	}
+	return false
 }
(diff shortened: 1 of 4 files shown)


dump/dump_konnect.go
```diff
@@ -168,3 +169,21 @@ func kongServiceIDs(cpID string,
...
 }
+
+// excludeKonnectManagedPlugins filter out konnect-managed plugins
+func excludeKonnectManagedPlugins(plugins []*kong.Plugin) []*kong.Plugin {
+	var filtered []*kong.Plugin
+	for _, p := range plugins {
+		var konnectManaged bool
+		for _, t := range p.Tags {
+			if *t == konnect.KonnectManagedPluginTag {
+				konnectManaged = true
```
hbagdi, line 180:
```suggestion
				konnectManaged = true
				break
```

dump/dump_konnect.go
```diff
@@ -168,3 +169,21 @@ func kongServiceIDs(cpID string,
 	}
 	return res
 }
+
+// excludeKonnectManagedPlugins filter out konnect-managed plugins
```
hbagdi, line 173:
```suggestion
// excludeKonnectManagedPlugins filters out konnect-managed plugins.
```

dump/dump_konnect_test.go
```diff
@@ -280,3 +280,63 @@ func Test_filterNonKongPackages(t *testing.T) {
...
 }
+
+func Test_excludeKonnectManagedPlugins(t *testing.T) {
+	tests := []struct {
+		name    string
+		plugins []*kong.Plugin
+		want    []*kong.Plugin
+	}{
+		{
+			name: "eclude konnect tags",
```
hbagdi, line 291:
```suggestion
			name: "exclude konnect tags",
```

We should enable a spell check linter to automate this.

dump/dump_konnect_test.go
```diff
@@ -280,3 +280,63 @@ func Test_filterNonKongPackages(t *testing.T) {
...
+				{
+					Name: kong.String("rate-limiting"),
+					Tags: []*string{kong.String("tag1")},
+				},
+				{
+					Name: kong.String("basic-auth"),
+					Tags: []*string{},
+				},
+			},
+		},
```
hbagdi, line 332: Add test cases:
1. Empty array input
2. Non-empty input but empty result (all plugins managed)

hbagdi (commented):
Needs a bit of clean up but LGTM otherwise.

The nested for loop would look better if the logic to detect the tag was in a separate function.

dump/dump_konnect_test.go
GGabriele: Done.

hbagdi (approved):

Issue #399, opened 2021-05-21
Analytics issues on decK v1.7.0


joanbm:
* From the usage printed by running `deck` it's not clear how one may disable analytics from the command line.

```
$ deck
...
Flags:
      --analytics                      share anonymized data to help improve decK (default true)
      --ca-cert string                 Custom CA certificate to use to verify Kong's Admin TLS certificate.
                                       This value can also be set using DECK_CA_CERT environment variable.
...
```

The default is `true` and just adding `--analytics` will keep it to true. It looks like the syntax is `--analytics=false` but that's not evident from the command line usage. The related environment variable is `DECK_ANALYTICS=off` but `--analytics=off` will fail. Demo:

```
$ strace -f -s 10000 deck validate 2>&1 | grep "<14>" # Sent
[pid 61373] write(3, "<14>cmd=validate;osv=arch ;signal=decK;os=linux;arch=amd64;v=1.7.0;", 67) = 67
$ strace -f -s 10000 deck validate --analytics 2>&1 | grep "<14>" # Still sent
[pid 61500] write(3, "<14>arch=amd64;v=1.7.0;cmd=validate;osv=arch ;signal=decK;os=linux;", 67 <unfinished ...>
$ deck validate --analytics=off
Error: invalid argument "off" for "--analytics" flag: strconv.ParseBool: parsing "off": invalid syntax
$ strace -f -s 10000 deck validate --analytics=false 2>&1 | grep "<14>" # Not sent now
$ 
```

---

* The command `deck konnect ping` ignores the `--analytics` flag.

```
$ strace -f -s 10000 deck konnect ping 2>&1 | grep "<14>" # Sent
[pid 128631] write(3, "<14>signal=decK;os=linux;arch=amd64;v=1.7.0;cmd=konnect-ping;osv=arch ;", 71 <unfinished ...>
$ strace -f -s 10000 deck konnect ping --analytics=false 2>&1 | grep "<14>" # Still sent
[pid 128670] write(3, "<14>osv=arch ;signal=decK;os=linux;arch=amd64;v=1.7.0;cmd=konnect-ping;", 71) = 71
$ DECK_ANALYTICS=off strace -f -s 10000 deck konnect ping 2>&1 | grep "<14>" # Not sent now
$
```

GGabriele:
Improved docs in https://github.com/Kong/deck/pull/566

PR #566, opened 2022-01-21
style: clarify how to disable 'analytics'

GGabriele:
Currently decK collects and shares anonymized data with Kong.
This feature is controlled via the '--analytics' flag,
which is True by default. The flag description doesn't
describe how users can prevent decK to share this data though.

This PR makes it clear to use --analytics=false for that.

Related to https://github.com/Kong/deck/issues/399


hbagdi (approved):
True by default APIs are bad. I should have named this flag "--disable-analytics". 😞

Issue #561, opened 2022-01-12
Error: tags specified in the state file and via --select-tags flag are different. when using more than one config file


michaelhsk:
Kong Version 2.7
Deck Version 1.10.0

When using deck sync with --select-tag with multiple yaml files containing select_tags (always the samte tag "deck-managed" used in command) i get the error

Error: tags specified in the state file and via --select-tags flag are different.
                                        decK expects tags to be specified in either via flag or via state file.
                                        In case both are specified, they must match

If i use the command  only with one of the files everything works fine.

With deck 1.8.2 no problems with the same config/commands.

f.g. 
config for file1.yaml - file10.yaml contains tag deck-managed
```
...
_info:
  select_tags:
  - deck-managed
...
```

command: 
  deck sync --select-tag deck-managed -s /config
OR
  deck sync --select-tag deck-managed -s /config/file1.yaml -s /config/file2.yaml

produces the error

if i remove parameter from command 
"deck sync -s /config"
-> everything works as expected

same if i use only 1 file with parameter
deck sync --select-tag deck-managed -s /config/file1.yaml
deck sync --select-tag deck-managed -s /config/file2.yaml
...

or if i use parameter in command line and remove 
following lines  from yaml files
_info:
  select_tags:
  - deck-managed
...

no error.

Maybe that this bug come in with this change in v.1.9.0 ?:
1beef01 fix(sync) apply --select-tag to new entities

GGabriele:
As stated in the [docs](https://docs.konghq.com/deck/1.8.x/guides/distributed-configuration/#sync), you don't need to use `--select-tag` with `sync` when the tag information is already in the state files, as `sync` will already use that/those.

Also, always from the docs:

> It is strongly advised that you do not supply select-tags to sync and diff commands via flags. This is because the tag information should be part of the declarative configuration file itself in order to provide a practical declarative file. The tagging information and entity definitions should be present in one place, else an error in supplying the wrong tag via the CLI can break the configuration.

That said, deck shouldn't error out like it's doing in this spot.

GGabriele:
Fixed via https://github.com/Kong/deck/pull/571

PR #571, opened 2022-01-26
fix: unexpected error with --select-tag and multiple files

GGabriele:
When running sync with multiple files and `--select-tag`,
deck simply appends the existing `_info.select_tags` of the
various files. If the same tag(s) are present in the files,
these are appended together making `deck` fail because the
tags passed via the command flag and the ones coming from
the files are expected to match, but the way files are
merged may cause duplicates, hence the unexpected failure.

This removes duplicates from the _info.select_tags resulted
from the merged files.

Related to https://github.com/Kong/deck/issues/561

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -172,12 +172,13 @@ func determineSelectorTag(targetContent file.Content, config dump.Config) ([]str
 	if targetContent.Info != nil {
 		if len(targetContent.Info.SelectorTags) > 0 {
 			if len(config.SelectorTags) > 0 {
+				utils.RemoveDuplicates(&targetContent.Info.SelectorTags)
 				sort.Strings(config.SelectorTags)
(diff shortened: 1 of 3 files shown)


rainest (changes requested):
Can you add a case to https://github.com/Kong/deck/blob/d0573b46b7f549c6422f8f639839205c94f80d3b/cmd/common_test.go#L11-L67 that would trigger the issue without this fix?

GGabriele:
> Can you add a case to
> 
> https://github.com/Kong/deck/blob/d0573b46b7f549c6422f8f639839205c94f80d3b/cmd/common_test.go#L11-L67
> 
> that would trigger the issue without this fix?

Done: https://github.com/Kong/deck/pull/571/commits/32071ed2c7405b2f55c1277ca8d94335f4f6bd4f

rainest (approved):

PR #572, opened 2022-01-26
tests: initial skeleton for integration tests

GGabriele:
This PR starts introducing some basic integration tests for the `sync` command. The goal is to agree on an overall structure for such tests, so that we can easily iterate over it.

At the moment we are only testing `deck` outputs here, but then we will expand the scope to test Kong state as well.

What this PR introduces:
- [.ci/setup_kong.sh](https://github.com/Kong/deck/blob/init_integration_tests/.ci/setup_kong.sh): borrowed from [go-kong](https://github.com/Kong/go-kong/blob/main/.ci/setup_kong.sh), this script setups a Kong instance to be used during tests
- CI setup for OSS [versions](https://github.com/Kong/deck/blob/init_integration_tests/.github/workflows/test.yaml#L10) (more OSS versions and EE to be added)
- a basic structure for test cases. Each test case is "represented" by a folder in `cmd/testdata/sync/$KONG_VERSION`: it's hard to test entities creation and modification across Kong releases since their schemas is not guaranteed to remain the same. For this reason, I'm proposing to version tests too with the mentioned format. This can easily become verbose, but I think it's a fair price to pay in exchange of a more complete testing scope. Each [testcase](https://github.com/Kong/deck/tree/init_integration_tests/cmd/testdata/sync/2.7.0/create_service) is composed of a `kong.yaml`, that will be used to deploy Kong, and a `output` file with the expected `deck` results.
- testing logic in [sync_test.go](https://github.com/Kong/deck/blob/init_integration_tests/cmd/sync_test.go): clears current Kong environment (the equivalent of running `reset`), collects test case files, deploys the specific `kong.yaml` and finally checks whether the output matches the expectations or not


cmd/sync_test.go
```diff
@@ -0,0 +1,134 @@
+package cmd
+
```
hbagdi, line 2: Guard this file with an integration tag.
make tests shouldn't run tests that have dependencies.

hbagdi: This test also must move under "tests/e2e" or something like that.

rainest:
* This approach, where you actually run the command, is what we've called an E2E test in KIC. We probably do want this for deck also, but it can just be a single basic input file with 2-3 entities to confirm that the finished product is viable. For the integration tests, you'll want to instead call `cmd.syncMain()` directly. This allows for debugging--we can't step into the deck internals when invoking the command.
* For the tests themselves we'll want to run a go-kong request to confirm that we actually see the expected entities after syncing. deck's own output is pretty information sparse by default, and even the verbose version isn't great since it'd all be string parsing instead of structs.
* For version-agnostic tests, do you see any reason we wouldn't be able to use `t.Skip()` if outside a version range similar to go-kong? Note that the utility functions we use for that [do need some changes currently in a staged branch](https://github.com/Kong/go-kong/commit/877bf41bf0976be93c0ff2df4105902624162fb0) to handle Enterprise versions properly (as much as we can) for semver comparisons, but that strategy should generally work: we don't usually make breaking API changes to existing entities (not since 2.0.0), only add new features to them. As such, you can, for example, write a generic route test hitting the main fields for any version and then optionally write a test that includes `request_buffering` for versions higher than whenever we added that.
* We may want to consider reusing the [KIC+KTF scaffolding for standing up Kong instances and supporting services in a Kubernetes cluster](https://github.com/Kong/kubernetes-ingress-controller/blob/main/test/integration/suite_test.go), and sub out KIC for deck. That's been useful there since the test runner is entirely Golang and we can use a debugger throughout, which we can't do for the bash portion here. IIRC using setup-kong.sh also imposed some limitation on what we were able to do with go-kong CI, though I can't remember the specifics. Not hugely important since we don't care about debugging environment provisioning, usually, but it's an option.

hbagdi:
> This approach, where you actually run the command, is what we've called an E2E test in KIC. We probably do want this for deck also, but it can just be a single basic input file with 2-3 entities to confirm that the finished product is viable. For the integration tests, you'll want to instead call cmd.syncMain() directly. This allows for debugging--we can't step into the deck internals when invoking the command.


We certainly need to test the internals as well, this is about testing only part one of the grand testing plan.

> For the tests themselves we'll want to run a go-kong request to confirm that we actually see the expected entities after syncing.

+1

> For version-agnostic tests, do you see any reason we wouldn't be able to use t.Skip() if outside a version range similar to go-kong?

We can certainly do that for a test that is specific to a version of Kong. I think most of our tests won't be specific to versions of Kong so starting first with generic tests that we can throw against all supported versions of Kong is the first step.

> We may want to consider reusing the KIC+KTF scaffolding for standing up Kong instances and supporting services in a Kubernetes cluster, and sub out KIC for deck.

-2
I'd like to keep the CI and setup simpler for decK - nothing more than a couple Docker containers on localhost or CI.
There is no reason to increase the complexity - contributors to decK shouldn't have to deal with k8s at all.

.ci/setup_kong.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
+
+set -e
+# download Kong deb
+
+sudo apt-get update
+sudo apt-get install openssl libpcre3 procps perl wget zlibc
+
```
hbagdi, line 8: Can we switch to Docker instead?
That would help with keeping CI and localhost using the same setup.

GGabriele: Sure, makes sense

tests/e2e/sync_test.go
```diff
@@ -0,0 +1,114 @@
+//go:build e2e_tests
+// +build e2e_tests
```
hbagdi, line 2: You don't need both, the first line should suffice (with Go 1.17).
nit: let's call this "integration"

Makefile
```diff
@@ -36,3 +36,13 @@ coverage:
...
+
+.PHONY: setup-kong
+setup-kong:
+	bash .ci/setup_kong.sh
+
+.PHONY: test.e2e
+test.e2e:
+	GOFLAGS="-tags=e2e_tests" go test -v \
+		-race \
+		./tests/e2e/...
```
hbagdi, line 48:
```suggestion
	go test -v -tags=e2e_tests\
		-race \
		./tests/e2e/...
```

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,26 @@
+name: e2e Test
+
+on: [push, pull_request]
+
+jobs:
+  e2e-tests:
+    strategy:
+      matrix:
+        kong_image:
+        - 'kong:2.7.0'
```
hbagdi, line 10: Let's do all minor releases until 1.x and see what breaks.

hbagdi: You will need to pass this as an "env var" and then read that in the setup-kong.sh script. See https://github.com/Kong/go-kong/blob/main/.github/workflows/integration-test.yaml#L16

tests/e2e/sync_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+var DECK = filepath.Join("../../deck")
+
+func getKongAddress() string {
+	if os.Getenv("KONG_ADDRESS") != "" {
+		return os.Getenv("KONG_ADDRESS")
```
hbagdi, line 25: Read once and return

addr := os.Getenv("foo")
if addr != "" {
  return addr
}

tests/e2e/sync_test.go
```diff
@@ -0,0 +1,114 @@
...
+	}{
+		{
+			name: "create_service",
+		},
+	}
+
+	if err := cleanUpEnv(); err != nil {
+		t.Errorf(err.Error())
+	}
+
```
hbagdi, line 89: You are mixing table-driven tests with t.Run() based tests.
The former has most of the data in a table and then the same logic across all t.Run().
Later has no table of data, but each function does the setup and teardown.

I think it makes sense to do table-driven tests here.
The table struct should take in an "input file name", "expected output file name", and then we simply do the same thing across multiple tests.
t.Run() then has the common logic.
You only need a bit of refactoring to get this done, should be quick but the result with be much better (especially as we add more tests).

hbagdi (changes requested):

tests/e2e/sync_test.go
GGabriele: The reason why I was doing this is because I was thinking to have the different test cases to be "incremental" (e.g. create a service, re-create it and expect to see no change, add a route and test the diff etc), so having an initial "setup" to run just once for every test suite seemed the way to go.

If we don't want to mix things, we can have something like "initial state" as part of the table struct (along with "input file name" and "expected output file name") being a kong file to deploy at the beginning of each testcase, and a teardown step at the end of each testcase.

hbagdi: It would be best to have small atomic tests. Where we need to compare against an existing state, the "setup" phase for that test should do that setup. This ensures two things:
- A developer can run each and every test, debug it individually.
- Ordering of tests doesn't matter.
- Maintaining state across tests in relation to Kong's database can lead to situations that are annoyingly hard to debug.
- One test failing won't lead to failure of other (dependent) tests.
- Changing one test due to a bug won't require changing any other test.
- Small individual tests decrease the likelihood of tests being buggy (and resulting in code being buggy).

tests/integration/sync_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"path/filepath"
+	"reflect"
+	"sort"
+	"strings"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+var DECK = filepath.Join("../../deck")
```
hbagdi, line 20: This will result in a bad developer experience - a compile step is required whenever a test has to run.
Can you please instead invoke main() from here? You can use https://pkg.go.dev/testing#B.Setenv to set the right environment variables for configuration.

tests/e2e/sync_test.go
GGabriele: Okay I see what I was doing/thinking wrong, makes sense. Thanks for the thoughtful explanation!

tests/integration/sync_test.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// cleanUpEnv removes all existing entities from Kong.
+func cleanUpEnv() error {
+	cmd := exec.Command(
+		"go",
+		"run",
+		DECK,
+		"reset",
+		"--force",
```
hbagdi, line 37: Why not simply call main() as a Go command?

rainest: Similar to the first point from https://github.com/Kong/deck/pull/572#issuecomment-1023621268 earlier, there are two approaches: invoke the actual command as a user would or invoke functions directly from the tests. I'd recommend starting with the latter, since most of your complex testing should go that route to allow for debugger usage and properly-typed checks.

Command invocation tests are useful also, but more to test a minimal "does the final product work/are there any issues with the wrappers we use around the functions we call directly in integration tests?" question.

hbagdi: +2 ^

GGabriele: +3 ^

I think we should agree on a plan for this effort, and this PR was open with this intent. Does the following sounds reasonable?

Limited to the scope of this PR:
- general CI setup and utilities (changes in `.ci/` and `.github/` folders and `Makefile`), already present in this PR 
- Simple test invoking `syncMain` directly and test results via API (this would require a bit of changes, as `syncMain` is unexported and also today there is no way to inject config parameters into it, which are populated as part of the root CLI)

On another PR:
- add more test cases to the above, including more entities etc

On an even further PR:
- add "e2e" tests testing the CLI usage (what I was doing here at this time)

GGabriele: Rework on this direction: https://github.com/Kong/deck/pull/572/commits/1ca2c4966498a2a39a07c87117bfa84ccc910f9d

tests/integration/sync_test.go
```diff
@@ -0,0 +1,106 @@
...
+		Address: getKongAddress(),
+	})
+}
+
+func testKongState(
+	t *testing.T,
+	client *kong.Client,
+	expectedServices []*kong.Service,
+) {
+	ctx := context.Background()
```
hbagdi, line 48: Use a timeout. In this codebase, it will be rare (if there is one) to do an operation without a timeout.

tests/integration/sync_test.go
```diff
@@ -0,0 +1,106 @@
...
+		t.Errorf(err.Error())
+	}
+
+	tests := []struct {
+		name             string
+		kongFile         string
+		expectedServices []*kong.Service
+	}{
+		{
+			name:     "create_service",
```
hbagdi, line 75:
```suggestion
			name:     "creates a service",
```

hbagdi (approved):
Let's start adding tests now!

PR #573, opened 2022-01-27
feat: add entities defaults from their schemas

GGabriele:
Today deck populates entities defaults defining them as part
of the Defaulter utility. This PR introduces the ability
to pull the schema of an entity from the Admin API and to
add its defaults from there.

What this PR doesn't do is change a thing on Defaulter.
A refactoring PR will follow.

This needs https://github.com/Kong/go-kong/pull/119

**before this change: deck keep showing diffs for unfilled defaults**
```
$ cat kong.yaml
services:
- name: svc1
  host: mockbin.org
  tags:
  - team-svc1
  routes:
  - name: r1
    https_redirect_status_code: 301
    paths:
    - /r1
upstreams:
- name: upstream1
  targets:
  - target: 198.51.100.11:80
```

```
$ deck sync
updating upstream upstream1  {
-  "algorithm": "round-robin",
   "hash_fallback": "none",
   "hash_on": "none",
   "hash_on_cookie_path": "/",
   "healthchecks": {
     "active": {
       "concurrency": 10,
       "healthy": {
         "http_statuses": [
           200,
           302
         ],
         "interval": 0,
         "successes": 0
       },
       "http_path": "/",
-      "https_verify_certificate": true,
       "timeout": 1,
       "type": "http",
       "unhealthy": {
         "http_failures": 0,
         "http_statuses": [
           429,
           404,
           500,
           501,
           502,
           503,
           504,
           505
         ],
         "interval": 0,
         "tcp_failures": 0,
         "timeouts": 0
       }
     },
     "passive": {
       "healthy": {
         "http_statuses": [
           200,
           201,
           202,
           203,
           204,
           205,
           206,
           207,
           208,
           226,
           300,
           301,
           302,
           303,
           304,
           305,
           306,
           307,
           308
         ],
         "successes": 0
       },
-      "type": "http",
       "unhealthy": {
         "http_failures": 0,
         "http_statuses": [
           429,
           500,
           503
         ],
         "tcp_failures": 0,
         "timeouts": 0
       }
     },
-    "threshold": 0
   },
   "id": "0898618e-df83-4289-b112-0490d74051a6",
   "name": "upstream1",
   "slots": 10000
 }

updating service svc1  {
   "connect_timeout": 60000,
   "host": "mockbin.org",
   "id": "5b2f5b96-dc80-4b0f-9efa-10f3789b2a72",
   "name": "svc1",
   "port": 80,
   "protocol": "http",
   "read_timeout": 60000,
-  "retries": 5,
   "tags": [
     "team-svc1"
   ],
   "write_timeout": 60000
 }

updating route r1  {
   "https_redirect_status_code": 301,
   "id": "acbb5161-be3d-4f78-9cf1-855c71fd01c2",
   "name": "r1",
-  "path_handling": "v0",
   "paths": [
     "/r1"
   ],
   "preserve_host": false,
   "protocols": [
     "http",
     "https"
   ],
   "regex_priority": 0,
-  "request_buffering": true,
-  "response_buffering": true,
   "service": {
     "id": "5b2f5b96-dc80-4b0f-9efa-10f3789b2a72"
   },
   "strip_path": false
 }

Summary:
  Created: 0
  Updated: 3
  Deleted: 0
```

**after this change: all defaults are added**

```
$ deck sync
Summary:
  Created: 0
  Updated: 0
  Deleted: 0
```


go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	github.com/hashicorp/go-memdb v1.3.2
 	github.com/hexops/gotextdiff v1.0.3
 	github.com/imdario/mergo v0.3.12
-	github.com/kong/go-kong v0.25.2-0.20220123181121-e965de22d701
+	github.com/kong/go-kong v0.25.2-0.20220126173454-9aab6cf4ade8
```
hbagdi, line 16: Blocker for this PR.
Let's get the upstream dependent PR merged. And then do a release of go-kong and then bump it up here.
These commit hash-based dependencies are a bit scary since they can point to a stale commit.

file/builder.go
```diff
@@ -551,6 +551,9 @@ func (b *stateBuilder) services() {
 }
 
 func (b *stateBuilder) ingestService(s *FService) error {
+	if err := b.addEntityDefaults("services", &s.Service); err != nil {
+		return fmt.Errorf("add defaults to service '%v': %v", *s.Name, err)
+	}
```
hbagdi, line 556: See line below:
> b.defaulter.MustSet(&s.Service)

Can we please keep the defaulting logic in the defaulter?
Take "Defaulter" as an input to StateBuilder if that is required.

go.mod
GGabriele: indeed

file/builder.go
```diff
@@ -56,9 +56,11 @@ func (b *stateBuilder) build() (*utils.KongRawState, *utils.KonnectRawState, err
...
 		kongDefaults = b.targetContent.Info.Defaults
 	}
-	b.defaulter, err = defaulter(kongDefaults)
-	if err != nil {
-		return nil, nil, fmt.Errorf("creating defaulter: %w", err)
+	if !isKongDefaultEmpty(kongDefaults) || b.defaulter == nil {
+		b.defaulter, err = defaulter(kongDefaults)
+		if err != nil {
+			return nil, nil, fmt.Errorf("creating defaulter: %w", err)
+		}
```
hbagdi, line 63: The KongDefaults thing hasn't been around for long enough and most users won't fill it in.
We should have a precedence order here:
- Take an entity, populate fields take first priority
- Fill defaults as present in KongDefaults
- Fill defaults via the API

utils/defaulter.go
```diff
@@ -126,3 +131,69 @@ func (d *Defaulter) MustSet(arg interface{}) {
...
+	if err := d.addEntityDefaults("targets", &target); err != nil {
+		return nil, fmt.Errorf("get defaults for targets: %v", err)
+	}
+	if err := d.Register(&target); err != nil {
+		return nil, fmt.Errorf("registering target with defaulter: %w", err)
+	}
+	return d, nil
+}
+
+func GetDefaulter(ctx context.Context, client *kong.Client) (*Defaulter, error) {
```
hbagdi, line 194: Same function signature for two exported functions (this and WithClient), remove one.

hbagdi (commented):
This is looking much better.
With the change, I'm requesting here (prioritized defaults), this can get a bit more complex but the complexity should be hidden in the "Deafulter".

file/builder.go
GGabriele: A couple of comments after my last commit:
- this change makes the hardcoded defaults in `deck` obsolete, so I'm removing them here
- which means I had to rework a bit the tests too. I moved the "old" default objects inside the test files in order to minimize the changes on tests, since we are already changing a lot of code. This should keep confidence of changes higher

let me know what you think!

utils/defaulter.go
GGabriele: ops, fixed

hbagdi:
@rainest This is a high-risk change. Could you please review and test this change?

rainest:
Loading the branch into the ingress controller and running integration tests surfaced one issue--possibly more, but they're not obvious (the same tests fail repeatedly, but there's no schema violation shown). Some entities are effectively several different entities under a trench coat, and those sub-entities can swap out different parts of the parent schema as they choose. Routes [use and require a different default for `strip_path`](https://github.com/Kong/kong/blob/d9d20b917b225869b146daaa9e68a64c57e82fbe/kong/db/schema/entities/routes_subschemas.lua#L50), so this breaks gRPC route creation unless you explicitly set it. By adding

```
      protocols:  
      - grpc
      - grpcs
```
to the example route in the OP, you get

```
16:29:10-0800 esenin $ /tmp/bdeck sync -s /tmp/kong.yaml
updating route r1  {
   "https_redirect_status_code": 301,
   "id": "790e50d2-e19f-438b-ac16-470107183658",
   "name": "r1",
   "path_handling": "v0",
   "paths": [
     "/r1"
   ],
   "preserve_host": false,
   "protocols": [
-    "http",
+    "grpc",
-    "https"
+    "grpcs"
   ],
   "regex_priority": 0,
   "request_buffering": true,
   "response_buffering": true,
   "service": {
     "id": "dd530429-739a-471f-80fc-6a90f1074893"
   },
   "strip_path": true
 }

Summary:
  Created: 0
  Updated: 0
  Deleted: 0
Error: 1 errors occurred:
	while processing event: {Update} route r1 failed: HTTP status 400 (message: "schema violation (strip_path: cannot set 'strip_path' when 'protocols' is 'grpc' or 'grpcs')")
```
It doesn't look like anything other than routes exhibits that problem (at least as far as file naming--nothing else has a `_subschemas.lua`.

Checking further on the other things that failed (mostly UDP routes).

go.mod
GGabriele: I created a new go-kong release and updated the dependence here

rainest:
Other KIC failures are poorly-configured timeouts on our end and can be ignored.

For actual deck usage the only thing that comes to mind is the [existing defaults system](https://docs.konghq.com/deck/1.10.x/guides/defaults/), and this is apparently overriding that. After syncing [test1.yaml.txt](https://github.com/Kong/deck/files/7982540/test1.yaml.txt) with current main, syncing again with this branch overwrites the value from the manifest default:

```
15:13:21-0800 esenin $ /tmp/adeck sync -s /tmp/poing.yaml
creating service example_service
creating route mockpath
Summary:
  Created: 2
  Updated: 0
  Deleted: 0
15:13:24-0800 esenin $ /tmp/bdeck sync -s /tmp/poing.yaml 
updating route mockpath  {
   "https_redirect_status_code": 426,
   "id": "aad4b34e-bd8b-451c-be44-2e78d8559e66",
   "name": "mockpath",
   "path_handling": "v0",
   "paths": [
     "/mock"
   ],
   "preserve_host": false,
   "protocols": [
     "http",
     "https"
   ],
   "regex_priority": 0,
   "request_buffering": true,
   "response_buffering": true,
   "service": {
     "id": "bdaca057-5da5-4f36-b7cb-9e8f809a904e"
   },
-  "strip_path": false
+  "strip_path": true
 }

Summary:
  Created: 0
  Updated: 1
  Deleted: 0
```

That seems unwanted, since the defaults in the manifest are user-specified, and should take precedence as such.

Other than that and base functionality (which seems fine), are there any other feature interactions we'd want to review?

Tangentially, why the heck does syncing [test2.yaml.txt](https://github.com/Kong/deck/files/7982544/test2.yaml.txt) just delete everything with the defaults in place? If you remove that block, it syncs the resources normally (excluding the invalid route). This happens on current main also, it's not something this branch broke.

GGabriele:
> Tangentially, why the heck does syncing test2.yaml.txt just delete everything with the defaults in place? If you remove that block, it syncs the resources normally (excluding the invalid route). This happens on current main also, it's not something this branch broke.

Apparently `deck` is not happy about some leading spaces in the first couple of lines. Removing them will make `sync` work as expected

rainest:
Works for me with the change, and importing the modified into KIC didn't trigger any test failures, so https://github.com/Kong/go-kong/pull/126

GGabriele:
@hbagdi @rainest I made few changes to make sure to achieve the following order of precedence:

1. values set in the state file
2. values set in the `{_info: defaults:}` object in the state file
3. schema defaults coming from Admin API (excluded Konnect)
4. hardcoded defaults under utils/constants.go (Konnect-only)

Let me know what you think!

rainest:
We'd very much like to avoid having all users set `strip_path` manually for gRPC routes. In the KIC context that's a "you just have to know to set it" gotcha, else your sync breaks. We could add a webhook case for it, but that's still a bit annoying since you'll still get a rejection that you need to fix.

It looks like we can work around the non-standard schema behavior by having a conditional in the middle of [ingestRoute()](https://github.com/Kong/deck/blob/v1.10.0/file/builder.go#L757-L759): after running `MustSet()`, if `r.Route.Protocols` contains `grpc` or `grpcs`, set `r.Route.StripPath = false`. Any reason not to do that to handle the special case?

Prior to setting the defaults, we should have the same condition check if there's a user-set `strip_path: true` that we can't honor and error out. Seems a bit silly as there's only one possible value, but it seems better to alert users that their requested state is impossible rather than silently correct their mistake.

GGabriele:
> handle case of strip_path and grpc protocols

@rainest can you please check the last commit? I added a patch to handle the case you mentioned.

rainest:
With the latest changes, automated KIC tests pass without modification, so we're good from my perspective: https://github.com/Kong/kubernetes-ingress-controller/runs/5102116046?check_suite_focus=true

file/builder.go
```diff
@@ -51,15 +53,11 @@ func (b *stateBuilder) build() (*utils.KongRawState, *utils.KonnectRawState, err
 		return nil, nil, err
 	}
 
-	// defaulter
-	var kongDefaults KongDefaults
-	if b.targetContent.Info != nil {
-		kongDefaults = b.targetContent.Info.Defaults
-	}
-	b.defaulter, err = defaulter(kongDefaults)
+	defaulter, err := defaulter(b.ctx, b.client, b.targetContent, b.isKonnect)
```
hbagdi, line 56: Not in scope but we really need to pas the defaulter as an input to rather than constructing it here.

file/builder.go
```diff
@@ -27,6 +27,8 @@ type stateBuilder struct {
 
 	schemasCache map[string]map[string]interface{}
 
+	isKonnect bool
```
hbagdi, line 30: rename to disableDynamicDefaults

utils/defaulter.go
```diff
@@ -1,39 +1,68 @@
...
+		target:   &kong.Target{},
+	}
 }
 
 // GetKongDefaulter returns a defaulter which can set default values
 // for Kong entities.
-func GetKongDefaulter() (*Defaulter, error) {
-	var d Defaulter
-	err := d.Register(&serviceDefaults)
+func GetKongDefaulter(kongDefaults interface{}, isKonnect bool) (*Defaulter, error) {
```
hbagdi, line 38: Take input DefaulterOpts with these two.

Rename isKonnect to disableDynamicDefaults

utils/defaulter.go
```diff
@@ -126,3 +155,143 @@ func (d *Defaulter) MustSet(arg interface{}) {
...
+
+// GetDefaulter returns a Defaulter object to be used to set defaults
+// on Kong entities. The order of precedence is as follow, from higher to lower:
+//
+// 1. values set in the state file
+// 2. values set in the {_info: defaults:} object in the state file
+// 3. schema defaults coming from Admin API (excluded Konnect)
+// 4. hardcoded defaults under utils/constants.go (Konnect-only)
+func GetDefaulter(
+	ctx context.Context, client *kong.Client, kongDefaults interface{}, isKonnect bool,
```
hbagdi, line 225: Take in all args as options, except ctx.

utils/defaulter.go
```diff
@@ -126,3 +155,143 @@ func (d *Defaulter) MustSet(arg interface{}) {
...
+) (*Defaulter, error) {
+	if client != nil && !isKonnect {
+		return getKongDefaulterWithClient(ctx, client, kongDefaults)
+	}
+	return GetKongDefaulter(kongDefaults, isKonnect)
```
hbagdi, line 230: Can we only export a single method - either this or the parent one? KISS - keep it simple silly.

tests/integration/sync_test.go
```diff
@@ -52,7 +52,7 @@ func testKongState(
 		t.Errorf(err.Error())
 	}
 	opt := []cmp.Option{
-		cmpopts.IgnoreFields(kong.Service{}, "ID", "CreatedAt", "UpdatedAt"),
+		cmpopts.IgnoreFields(kong.Service{}, "ID", "CreatedAt", "UpdatedAt", "Enabled"),
```
hbagdi, line 55: code smell: we will have new properties being defined all the time. Our tests should handle these.

file/builder.go
GGabriele: The only reason I did it so is to avoid duplicating the same exact lines for Kong and Konnect

tests/integration/sync_test.go
GGabriele: yeah, I will take care of this in the other "test" PR along with other test-related changes.

utils/defaulter.go
GGabriele: done

utils/defaulter.go
GGabriele: done

utils/defaulter.go
GGabriele: done

file/builder.go
```diff
@@ -748,6 +746,32 @@ func (b *stateBuilder) plugins() {
...
+	if grpcProtocols {
+		if initialStripPath != nil {
+			if *initialStripPath {
+				return nil, fmt.Errorf("schema violation (strip_path: cannot set " +
+					"'strip_path' when 'protocols' is 'grpc' or 'grpcs')")
+			}
+			return initialStripPath, nil
+		}
+		return kong.Bool(false), nil
+	}
```
hbagdi, line 771: Question: If strip path can't be set when protocol is grpc, why return false on the two returns above?

GGabriele: `strip_path` is a required field, so when protocol is grpc it needs to be set to `false` (`true` being the schema default)

hbagdi:
```go
	for _, p := range route.Protocols {
		if *p == "grpc" || *p == "grpcs" {
			if stripPath != nil { return nil, fmt.Errorf("strip path not allowed")
      return kong.Bool(false), nil
		}
	}
```
Is that not sufficient?

GGabriele: Not directly...I just reworked/cleaned a bit the code to have something like you are suggesting (https://github.com/Kong/deck/pull/573/commits/67d1dcc9485e7754195449dff24f1a663ae4f4eb)

rainest (approved):

Issue #506, opened 2021-10-14
Deck creates only one workspace and put all services and routes inside it


mkopeyka:
I use `decK v1.8.2 (b538f70)` and `kong-ee:v1.5.0.1-alpine`

I created 3 test workspaces via kong ui and added 1 service and 1 route to each of them. After that I created a dump with key `--all-workspaces` so deck created 3 files (one per workspace).
[auth.yaml.txt](https://github.com/Kong/deck/files/7349480/auth.yaml.txt)
[nondeck.yaml.txt](https://github.com/Kong/deck/files/7349481/nondeck.yaml.txt)
[unifiedlogin.yaml.txt](https://github.com/Kong/deck/files/7349482/unifiedlogin.yaml.txt)
    
    ls -la deck/config/
    total 20
    drwxr-xr-x 2 maxim maxim 4096 жов 15 00:20 .
    drwxr-xr-x 3 maxim maxim 4096 жов 11 16:27 ..
    -rwxr-xr-x 1 maxim maxim  496 жов 15 00:07 auth.yaml
    -rwxr-xr-x 1 maxim maxim  496 жов 15 00:18 nondeck.yaml
    -rwxr-xr-x 1 maxim maxim  498 жов 15 00:18 unifiedlogin.yaml
    

I deleted old database to test decK on fresh empty DB. I expected this command will create 3 new workspaces
    
    deck sync --state deck/config --kong-addr http://localhost:8001 --select-tag deck
    
But it creates only first workspace and adds all services and routes to this workspace
    
    creating workspace auth
    creating service ul-service
    creating service non-deck-service
    creating service auth-service
    creating route auth-route
    creating route non-deck-route
    creating route ul-route
    Summary:
      Created: 6
      Updated: 0
      Deleted: 0

GGabriele:
Now deck explicitly fail when attempting to sync in such conditions: https://github.com/Kong/deck/pull/576

PR #575, opened 2022-01-30
feat: support sync with multiple workspaces

GGabriele:
As highlighted in https://github.com/Kong/deck/issues/506, currently `deck` is not able to deploy multiple state files with multiple workspaces, and it actually mistakenly deploys all of them into a single workspace.

This PR is a proposal to start supporting such cases.

The main change is about the `file.getContent` function, which used to return a single `file.Content` pointer, but here it's returning a map of workspaces (`map[string]*Content`). All the other changes are consequences of this.

As a result, `deck` would show different reports per-workspace too.

Sample usage:

```
$ deck sync -s ws1.yaml -s ws2.yaml
creating service svc1
Summary (ws1):
  Created: 1
  Updated: 0
  Deleted: 0

creating service svc2
Summary (ws3):
  Created: 1
  Updated: 0
  Deleted: 0
```

```
$ deck reset --all-workspaces
This will delete all configuration from Kong's database.
> Are you sure? y
Summary (ws2):
  Created: 0
  Updated: 0
  Deleted: 0

deleting service svc1
Summary (ws1):
  Created: 0
  Updated: 0
  Deleted: 1

Summary (default):
  Created: 0
  Updated: 0
  Deleted: 0
```
 
@rainest  @hbagdi thoughts?

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -66,10 +66,21 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 	delay int, workspace string) error {
 
 	// read target file
-	targetContent, err := file.GetContentFromFiles(filenames)
+	targetContents, err := file.GetContentFromFiles(filenames)
 	if err != nil {
 		return err
 	}
+
+	for _, targetContent := range targetContents {
+		if err := syncWs(ctx, targetContent, dry, parallelism, delay, workspace); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func syncWs(ctx context.Context, targetContent *file.Content, dry bool, parallelism,
+	delay int, workspace string) error {
 	if dumpConfig.SkipConsumers {
 		targetContent.Consumers = []file.FConsumer{}
 	}
@@ -157,7 +168,7 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 		return err
 	}
 
-	totalOps, err := performDiff(ctx, currentState, targetState, dry, parallelism, delay, wsClient)
+	totalOps, err := performDiff(ctx, currentState, targetState, dry, parallelism, delay, wsClient, workspaceName)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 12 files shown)


hbagdi (commented):
We are deliberate about not supporting the configuration of multiple workspaces as part of a single run.
Conceptually, workspaces are a configuration isolation mechanism, with strict access control between them.
Configuration management tools must respect such a boundary.

Our users are free to sync multiple workspaces with some additional shell scripting. I'm very hesitant about adding such a mode in decK itself.
Having said that, I'm open to reasons for adding and hearing more about use-cases that can't be solved with the existing feature set.

GGabriele:
> We are deliberate about not supporting the configuration of multiple workspaces as part of a single run.
> Conceptually, workspaces are a configuration isolation mechanism, with strict access control between them.
> Configuration management tools must respect such a boundary.

I'm fine with that if it is intentional, maybe we should clarify this in our docs as sometimes it's hard to distinguish between intended behaviours, missing features or bugs (pinging @lena-larionova for awareness, but I can take care of this).

If that's the case, we should at least error out when users attempt to sync multiple files containing multiple workspace references.

https://docs.konghq.com/deck/1.10.x/guides/kong-enterprise/#manage-multiple-workspaces

hbagdi:
> If that's the case, we should at least error out when users attempt to sync multiple files containing multiple workspace references.

Let's do that.
cc @rainest who may be aware of any complications around this.

GGabriele:
Closing this in favor of https://github.com/Kong/deck/pull/576

rainest:
> We are deliberate about not supporting the configuration of multiple workspaces as part of a single run.
Conceptually, workspaces are a configuration isolation mechanism, with strict access control between them.
Configuration management tools must respect such a boundary.

We do handle this already inherently, correct? deck will only dump workspaces that its provided RBAC user has access to, and there are reasonable cases where the access granted to deck is different than that granted to other users (e.g. you grant your admins access only to their workspace, but use deck to pull a complete set of all configuration for a daily snapshot).

The availability of `--all-workspaces` on dump means you can act across workspace boundaries in one direction, so IMO it stands to reason that you can do it when syncing configuration to Kong also.

hbagdi:
I meant we are deliberate about "sync/diff" to be scoped to a workspace.

PR #576, opened 2022-02-01
fix: unexpected behaviour with sync and multiple ws

GGabriele:
When passing to `sync` multiple files (or a folder) with different
_workspace definition, sync merges the files and applies them
only to a single workspace.

This change makes `sync` returning an error when attempting
a deployment with multiple workspaces.

### DIFF
--- a/file/validate.go
+++ b/file/validate.go
@@ -35,7 +35,9 @@ func validate(content []byte) error {
 func validateWorkspaces(workspaces []string) error {
 	utils.RemoveDuplicates(&workspaces)
 	if len(workspaces) > 1 {
-		return fmt.Errorf("cannot sync multiple workspaces at the same time: %v", workspaces)
+		return fmt.Errorf("it seems like you are trying to sync multiple workspaces "+
+			"at the same time (%v).\ndecK doesn't support syncing multiple workspaces at the same time, "+
+			"please sync one workspace at a time", workspaces)
 	}
 	return nil
 }
--- a/file/validate.go
+++ b/file/validate.go
@@ -31,3 +31,11 @@ func validate(content []byte) error {
 	}
 	return errs
 }
+
+func validateWorkspaces(workspaces []string) error {
+	utils.RemoveDuplicates(&workspaces)
+	if len(workspaces) > 1 {
+		return fmt.Errorf("cannot sync multiple workspaces at the same time: %v", workspaces)
+	}
+	return nil
+}
--- a/file/readfile.go
+++ b/file/readfile.go
@@ -20,6 +20,7 @@ import (
 // these files and renders a Content.
 func getContent(filenames []string) (*Content, error) {
 	var allReaders []io.Reader
+	var workspaces []string
 	for _, fileOrDir := range filenames {
 		readers, err := getReaders(fileOrDir)
 		if err != nil {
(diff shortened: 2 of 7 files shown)


hbagdi:
@rainest Do we allow for "workspace" provided via CLI flag and that provided in the state file to differ?
That's one thing we need to take into account otherwise this is good.

rainest:
Current behavior is to print a warning and then proceed, preferring the 

```
/tmp/adeck sync -s decs --workspace foo
Warning: Workspace 'foo' specified via --workspace flag is different from workspace 'bar' found in state file(s).
...
Error: 1 errors occurred:
	while processing event: {Create} route r2 failed: HTTP status 409 (message: "API route collides with an existing API")
```
This is after I'd synced to the workspace in the manifest, so it failed to create a new one due to path collision.

With the new patch, it's rejected regardless:

```
13:13:42-0800 esenin $ /tmp/bdeck sync -s decs                
Error: cannot sync multiple workspaces at the same time: [bar ]
13:15:12-0800 esenin $ /tmp/bdeck sync -s decs --workspace foo
Error: cannot sync multiple workspaces at the same time: [bar ]
```
I think that's what we want, but that said, it looks like the output of the error is a bit off: I'd expect this to print `[bar, baz]`, but one of the workspaces is being omitted.

GGabriele:
> I think that's what we want, but that said, it looks like the output of the error is a bit off: I'd expect this to print [bar, baz], but one of the workspaces is being omitted.

What's the content of those files? I can't reproduce it:

```
$ deck sync -s ws1.yaml -s ws2.yaml
Error: cannot sync multiple workspaces at the same time: [ws1 ws2]
```

rainest:
Nevermind, weird malformed YAML from botched paste behavior strikes again. One had:

```
 _format_version: "0.1"
_workspace: baz
```
It properly detected the duplicate workspaces but then couldn't actually read it due to the leading space on the format. Edge case we don't care about for this, it should be solved with a general failure on any malformed YAML for https://github.com/Kong/deck/issues/578

GGabriele:
Let me know if any of you have more comments or if this is good to go

file/validate.go
```diff
@@ -31,3 +31,11 @@ func validate(content []byte) error {
 	}
 	return errs
 }
+
+func validateWorkspaces(workspaces []string) error {
+	utils.RemoveDuplicates(&workspaces)
+	if len(workspaces) > 1 {
+		return fmt.Errorf("cannot sync multiple workspaces at the same time: %v", workspaces)
```
hbagdi, line 38:
```suggestion
		return fmt.Errorf("It seems like you are trying to sync multiple workspaces at the same time (%v). decK doesn't support syncing multiple workspaces at the same time, please sync one workspace at a time.", workspaces)
```

Use appropriate \ns and maybe add an HTTP link to docs if we have one.
cc @lena-larionova

hbagdi (commented):
Merge with approval from Travis.

hbagdi:
> Current behavior is to print a warning and then proceed, preferring the

Preferring what? Whatever it be, we must ensure that behavior is not broken - a user ignoring the warning shouldn't receive an error after this patch.

GGabriele:
> > Current behavior is to print a warning and then proceed, preferring the
> 
> Preferring what? Whatever it be, we must ensure that behavior is not broken - a user ignoring the warning shouldn't receive an error after this patch.

Confirming this patch doesn't change that behaviour:

```
$ deck sync -s kong.yaml --workspace test
Warning: Workspace 'test' specified via --workspace flag is different from workspace 'ws1' found in state file(s).
...
...
Error: 1 errors occurred:
	while processing event: {Create} route r1 failed: HTTP status 409 (message: "API route collides with an existing API")
```

file/validate.go
lena-larionova: The doc is light on this, needs some improvement, but we do have something: https://docs.konghq.com/deck/1.10.x/guides/kong-enterprise/#workspaces

rainest (approved):
Preferring the workspace in the flag, whoops.

PR #579, opened 2022-02-03
tests: add more integration tests for sync

GGabriele:


tests/integration/sync_test.go
```diff
@@ -303,6 +310,10 @@ func Test_Sync(t *testing.T) {
 			teardown := setup(t, tc.initialKongFile)
 			defer teardown(t)
 
+			if kongVersion < "2.0" && len(tc.expectedState.Plugins) > 0 {
```
hbagdi, line 313: Please consult the go-kong codebase for how to detect kong versions and have a function to skip such tests, we may need to do such things more frequently.

tests/integration/testdata/sync/create_new_service/kong.yaml
```diff
@@ -0,0 +1,18 @@
+_format_version: "1.1"
```
hbagdi, line 1: I didn't catch this in the last review, it would be best to prefix these directories with a number: 001-create-new-service, that is much easier on the eyes and is easy to grep.

cmd/sync.go
```diff
@@ -7,7 +7,10 @@ import (
 )
 
 var (
-	syncCmdKongStateFile []string
+	// SyncCmdKongStateFile holds kong files passed as an input
+	// NOTE: exported for testing purposes.
+	SyncCmdKongStateFile []string
```
hbagdi, line 12: If you need to export this, the tests are doing something wrong.
Tests must be able to pass in the appropriate input - filenames in this case - via the main execute command.

tests/integration/sync_test.go
```diff
@@ -25,82 +26,296 @@ func getKongAddress() string {
...
+
+	if diff := cmp.Diff(kongState, &expectedState, opt...); diff != "" {
 		t.Errorf(diff)
 	}
 }
 
+func reset(t *testing.T) {
+	deckCmd.SetArgs([]string{"reset", "--force"})
+	if err := deckCmd.Execute(); err != nil {
+		t.Errorf(err.Error())
```
hbagdi, line 79:
```suggestion
		t.Failf(err.Error(), "failed to reset Kong's state")
```

tests/integration/sync_test.go
```diff
@@ -25,82 +26,296 @@ func getKongAddress() string {
...
 
+	kongVersion := getKongVersionFromEnv()
+
 	tests := []struct {
-		name             string
-		kongFile         string
-		expectedServices []*kong.Service
+		name            string
+		kongFile        string
+		initialKongFile string
```
hbagdi, line 118: Question: why two files? Where these variables are used, both are exactly the same, what is it the purpose here?

tests/integration/sync_test.go
hbagdi: I also recommend isolating tests that are version-specific rather than sibling assertion of a specific "state" for a specific version.
This additional complexity can get complicated as we add more tests.

An alternate method here will be to create a dedicated TestFoo that is disabled for version x or below, or to have "DisableFilter string" (that takes in a semver filter like >2.3.0) in the test table.

hbagdi (changes requested):

tests/integration/sync_test.go
GGabriele: The intent was to use this in order to test `sync` with an existing state, rather than only test results from a clean one.

hbagdi: I'd recommend having separate TestFuncs for these. Fresh vs Diffs.

cmd/sync.go
GGabriele: reworked

tests/integration/sync_test.go
rainest: We should export the go-kong functions for this and reuse them as much as possible; no reason to maintain separate versions of them since any Go project is likely going to include go-kong for other reasons anyway.

I've opened https://github.com/Kong/go-kong/pull/129 as well since the current Enterprise version code doesn't work properly for initial minor releases (they're turned into pre-releases so semver comparisons say they're older).

go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/hashicorp/go-memdb v1.3.2
 	github.com/hexops/gotextdiff v1.0.3
 	github.com/imdario/mergo v0.3.12
-	github.com/kong/go-kong v0.25.2-0.20220123181121-e965de22d701
+	github.com/kong/go-kong v0.27.0
```
hbagdi, line 19: This and the other PR both have bumps.
Both are violating the principle of atomic change - which will also reduce the time for PR reviews.
Not requesting a change here but please avoid such changes.
I know that doing such changes increases the time it takes to do git management and plan, but if other smaller PRs are any proof, they overall benefit the code quality and everyone's personal productivity.

tests/integration/sync_test.go
```diff
@@ -25,82 +137,443 @@ func getKongAddress() string {
...
+			teardown := setup(t)
+			defer teardown(t)
+
+			sync(tc.kongFile)
+			testKongState(t, client, tc.expectedState, nil)
+		})
+	}
+}
+
+// func Test_Sync_Plugins(t *testing.T) {
```
hbagdi, line 449: Either remove this code or get it working. Commented out code with VCS is a no no.

hbagdi (approved):

tests/integration/sync_test.go
GGabriele: sure, just wanted to have a +1 on the overall structure...I'll add them back now

tests/integration/sync_test.go
GGabriele: @rainest  Added an issue on go-kong for that https://github.com/Kong/go-kong/issues/130

GGabriele:
@rainest rebased after `go-kong` release.

rainest (approved):

PR #580, opened 2022-02-03
chore(deps): update mergo

GGabriele:
This makes us able to set zero-value fields and not have them overwritten from non-zero value defaults.

Before:

```
$ cat kong.yaml
upstreams:
- name: upstream1
  algorithm: round-robin
  targets:
  - target: 198.51.100.11:80
    weight: 0  # default being 100

$ deck sync
<--omitted-->

$ http :8001/upstreams/upstream1/targets | jq ".data[0].weight"
100
```

After:

```
$ deck sync
<--omitted-->

$ http :8001/upstreams/upstream1/targets | jq ".data[0].weight"
0
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,8 @@ module github.com/kong/deck
 
 go 1.16
 
+replace github.com/imdario/mergo v0.3.12 => github.com/Kong/mergo v0.3.13
+
 require (
 	github.com/alecthomas/jsonschema v0.0.0-20191017121752-4bb6e3fae4f2
 	github.com/blang/semver/v4 v4.0.0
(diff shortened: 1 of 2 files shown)


hbagdi:
Umm, based on the PR description, it seems that we won't be able to set target's weight to 0.
Is that intended?

From what I understand, we want to give the user the ability to specify a zero value.

GGabriele:
> Umm, based on the PR description, it seems that we won't be able to set target's weight to 0. Is that intended?

I included this snipped in the **After** section, which shows how we can now set zero-values:

```
$ deck sync
<--omitted-->

$ http :8001/upstreams/upstream1/targets | jq ".data[0].weight"
0
```

Sorry if that wasn't clear

hbagdi:
Sorry, my bad.
Can we include an end to end test using our shiny new testing framework?

GGabriele:
> Can we include an end to end test using our shiny new testing framework?

Just added [this](https://github.com/Kong/deck/pull/579/commits/533df79c5559c3cffc14a1567de25e902264412f#diff-b09c72ddce91eda1770990ad2744a3867eb59e006e6842fc4d01b8d3ce5eec77R322) in the other PR about integrations tests.

That one is currently failing, but it will pass once we merge this PR.

Issue #341, opened 2021-04-23
feature: implement --dedup-plugin-configs flag for dump command 


hbagdi:
`_plugin_configs` feature of decK allows users to abstract out configuration of plugins that are used repeatedly and then reference them in services/routes/consumers. This helps with de-duplication of configuration code.
While users would like to take advantage of this, most of them cannot because de-duplicating the configuration in the first place is a cumbersome and an error-prone process. The scope here is to automate the process for our users:

How can we do this?
After decK exports Kong's configuration but before it writes the configuration to file, deck analyzes if there are multiple plugins in the configuration that share identical configuration. If so, it will de-duplicate those configuration by putting them under `_plugin_configs` section and then add a reference to them wherever that configuration is in use.

PR #581, opened 2022-02-04
feat: implement --dedup-plugin-configs flag for dump

GGabriele:
> `_plugin_configs` feature of decK allows users to abstract out configuration
> of plugins that are used repeatedly and then reference them in
> services/routes/consumers. This helps with de-duplication of configuration
> code. While users would like to take advantage of this, most of them cannot
> because de-duplicating the configuration in the first place is a cumbersome
> and an error-prone process.

https://github.com/Kong/deck/issues/341 ([doc](https://docs.konghq.com/deck/1.10.x/guides/deduplicate-plugin-configuration/))

This implements such flag and feature for the dump command.

Test:
`sync` this file first ([before.yaml.txt](https://github.com/Kong/deck/files/8003745/before.yaml.txt)), having some shared plugins across different consumers, routes and services.

```
$ deck sync
<--omitted->
```

```
$ deck dump --yes --dedup-plugin-configs

$ deck sync
Summary:
  Created: 0
  Updated: 0
  Deleted: 0
```

Result: [after.yaml.txt](https://github.com/Kong/deck/files/8003780/after.yaml.txt)

### DIFF
--- a/dump/dump.go
+++ b/dump/dump.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"reflect"
 
 	"github.com/kong/deck/utils"
 	"github.com/kong/go-kong/kong"
@@ -26,6 +27,10 @@ type Config struct {
 	// SelectorTags can be used to export entities tagged with only specific
 	// tags.
 	SelectorTags []string
+
+	// If true, deduplicate plugins with the same configuration into a shared
+	// config object.
+	DedupPluginsConfig bool
 }
 
 func deduplicate(stringSlice []string) []string {
@@ -178,6 +183,9 @@ func getProxyConfiguration(ctx context.Context, group *errgroup.Group,
 			plugins = excludeConsumersPlugins(plugins)
 		}
 		state.Plugins = plugins
+		if config.DedupPluginsConfig {
+			dedupPluginsConfig(state, plugins)
+		}
 		return nil
 	})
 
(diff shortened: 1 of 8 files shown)


file/writer.go
```diff
@@ -605,6 +633,36 @@ func populateConsumers(kongState *state.KongState, file *Content,
...
+		}
+		for _, route := range content.Routes {
+			if route != routeID {
+				continue
+			}
+			fp.Plugin.Config = nil
+			fp.ConfigSource = kong.String(name)
+			break
+		}
+	}
```
hbagdi, line 663: Suggest two functions: one takes in the plugin map and the relation IDs, and returns the name of the plugin config source (if found), and this one then sets the config to nil and source to non-nil. This way you can unit-test the function that looks through sharedPlugins map.

hbagdi (commented):
Good start.

A few corner cases to think about:
- What if a plugin configuration is unique and is only present once in the entire state? The code shouldn't introduce an indirection (via plugin source name) in such a case.
- What if a plugin has multiple parents at the same time? Does this logic hold correctly in such a case? I'm not sure either way - please give that some thought and we will certainly need tests around it.

file/writer.go
GGabriele: refactored a bit

dump/dump.go
```diff
@@ -218,6 +226,68 @@ func getProxyConfiguration(ctx context.Context, group *errgroup.Group,
...
+// 		},
+// 		"rate-limiting-1": {
+// 			"config": _specific_plugin_config_1,
+// 			"consumers": ["consumer-3", "consumer-4"],
+// 			"services": []
+// 		}
+// 	  }
+// }
+func dedupPluginsConfig(state *utils.KongRawState, plugins []*kong.Plugin) {
+	state.SharedPluginMap = make(map[string]map[string]utils.SharedPlugin)
```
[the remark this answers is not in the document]
GGabriele: refactored a bit

dump/dump.go
```diff
@@ -218,6 +227,89 @@ func getProxyConfiguration(ctx context.Context, group *errgroup.Group,
...
+	if p.Consumer != nil && p.Consumer.ID != nil {
+		sharedPlugin.Consumers = append(sharedPlugin.Consumers, *p.Consumer.ID)
+	}
+	if p.Service != nil && p.Service.ID != nil {
+		sharedPlugin.Services = append(sharedPlugin.Services, *p.Service.ID)
+	}
+	if p.Route != nil && p.Route.ID != nil {
+		sharedPlugin.Routes = append(sharedPlugin.Routes, *p.Route.ID)
+	}
+	return sharedPlugin
```
hbagdi, line 240: The tracking here loses the relationship of a plugin with multiple entities.
Consider the following plugin:
```yaml
  plugins:
    name: rate-limiting
    service:
       id: svc-id
    consumer:
       id: consumer-id
    config:
      day: null
      fault_tolerant: true
      hide_client_headers: false
      hour: null
      limit_by: consumer
      minute: 10
      month: null
      policy: redis
      redis_database: 0
      redis_host: redis.common.svc
      redis_password: null
      redis_port: 6379
      redis_timeout: 2000
      second: null
      year: null
    enabled: true
    protocols:
    - http
    - https
```
In this case, this data structure would track the plugin as a plugin that is applied to the consumer, and the same plugin is applied to the service, in reality, that is not actually true, the plugin is present only once, on a combination of service and consumer.

The code that uses this data structure, later on, is structured in a way that one never runs into this bug currently. But that is a bit too risky to rely on. Please consider a data structure that captures the entire relationship explicitly.

hbagdi:

PR #591, opened 2022-02-16
feat: refactor cmd package

GGabriele:
The current version of the `cmd` package is structured
around some global variables holding the `cobra` commands.
This makes the package hard to test.

This reworks the `cobra` commands definition with 'constructor' helpers.
This also rearranges the existing flags, moving the ones specific
to `konnect` under the `konnect` subcommand.

### DIFF
--- a/cmd/completion.go
+++ b/cmd/completion.go
@@ -7,10 +7,11 @@ import (
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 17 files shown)


hbagdi (approved):
These large changes without any tests are very uncomfortable.

PR #592, opened 2022-02-17
tests: add integration tests to check default filling

GGabriele:


tests/integration/sync_test.go
```diff
@@ -127,6 +127,43 @@ var (
...
+				kong.String("grpc"),
+				kong.String("grpcs"),
+				kong.String("http"),
+				kong.String("https"),
+			},
+			Enabled: kong.Bool(true),
+			Config: kong.Configuration{
+				"anonymous":        "58076db2-28b6-423b-ba39-a797193017f7",
+				"hide_credentials": false,
+			},
```
hbagdi, line 163: Why not pick a plugin like rate-limiting which has a lot more default fields rather than this one which has only one it seems.

GGabriele: I was mostly testing the automatic-defaults functionality itself across all kong versions, so I picked one of the simplest plugins available. I'm actually testing [rate-limiting](https://github.com/Kong/deck/blob/test_defaults/tests/integration/sync_test.go#L775-L813) too but just for `>=2.70`, as a lot more fields may also mean a lot more differences across versions.

If you think it's worth to use something like `rate-limiting` insted of `basic-auth` for these tests, I'll do the switch.

tests/integration/sync_test.go
```diff
@@ -871,3 +877,172 @@ func Test_Sync_RateLimitingPlugin(t *testing.T) {
...
+				Routes:    route1_151,
+				Plugins:   plugin,
+				Targets:   target,
+				Upstreams: upstream,
+			},
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			kong.RunWhenKong(t, "<1.5.1 !1.4.3")
```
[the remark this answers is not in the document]
GGabriele: Apparently nothing, I added it splitting this test ¯\_(ツ)_/¯

hbagdi (approved):

Issue #601, opened 2022-03-02
Deck no longer creates workspace with deck sync --workspace flag


dmortim:
Prior to deck 1.11. We could run the following command to create a new workspace as we applied our statefiles. 
```
deck sync --workspace some-new-workspace -s state-files ... 
```

Now we receive the error
```
retrieve schema for services from Kong: HTTP status 404(message: "some-new-workspace" not found).
```

Is there a way we can create the workspace without using the admin API? Either through deck or through Kong declarative files? 

edit: 
Environment 
Kong Hybrid EE 2.7.1.1
Openapi -> inso CLI -> deck 

Thanks, 

btw, love the 1.11 update for injecting plugin defaults. Great work

GGabriele:
Hi @dmortim , happy to hear you like the new features! :)

I'm not currently able to reproduce your issue:

```
$ deck version
decK v1.11.0 (4235c73)

$ deck ping
Successfully connected to Kong!
Kong version:  2.7.1.1-enterprise-edition

$ deck sync --workspace test
Warning: Workspace 'test' specified via --workspace flag is different from workspace '' found in state file(s).creating workspace test
creating service svc1
Summary:
  Created: 1
  Updated: 0
  Deleted: 0
```

What happens if you add a `_workspace: some-new-workspace` entry in your state file? Do you have any RBAC in place? Does the issue go away if you switch to an older version of decK but with the same exact environment?

dmortim:
Hi @GGabriele, your are correct. Deck Sync Creates the workspace. 

We identified the root cause of the issue. We are now getting an error code 1 when we run a deck diff against a workspace that does not exist - and hence that error I put above. Previously, we would get an error code 0 if the workspace did not exist and it would list the changes to be made, including creating the new workspace. Is this the expected behavior? 

``` bash
$ deck diff --workspace test
Warning: Workspace 'test' specified via --workspace flag is different from workspace '' found in state file(s).
creating workspace test
Error: building state: creating defaulter: get defaults for services: retrieve schema for services from Kong: HTTP status 404 (message: "Workspace 'test' not found)

$ echo $?
1
```


Thanks

GGabriele:
PR has been merged. It will be included in the next release!

PR #608, opened 2022-03-09
fix: support defaulter with not-existing workspaces

GGabriele:
The new defaulter interacts with Kong API to retrieve
entities schemas. Doing such query against a not existing
workspace returns an error, blocking the diff operation in decK.

This makes sure decK can run 'diff' regardless of the workspace
by simply not retrieving entities schemas from Kong when the
workspace doesn't exist yet.

Solves https://github.com/Kong/deck/issues/601#issuecomment-1057647420

### DIFF
--- a/file/builder.go
+++ b/file/builder.go
@@ -813,7 +813,7 @@ func (b *stateBuilder) getPluginSchema(pluginName string) (map[string]interface{
 
 	exists, err := utils.WorkspaceExists(b.ctx, b.client)
 	if err != nil {
-		return nil, fmt.Errorf("getPluginSchema: checking if workspace exists: %w", err)
+		return nil, fmt.Errorf("ensure workspace exists: %w", err)
 	}
 	if !exists {
 		return schema, nil
--- a/file/builder.go
+++ b/file/builder.go
@@ -810,7 +810,16 @@ func (b *stateBuilder) getPluginSchema(pluginName string) (map[string]interface{
 	if schema, ok := b.schemasCache[pluginName]; ok {
 		return schema, nil
 	}
-	schema, err := b.client.Plugins.GetFullSchema(b.ctx, &pluginName)
+
+	exists, err := utils.WorkspaceExists(b.ctx, b.client)
+	if err != nil {
+		return nil, fmt.Errorf("getPluginSchema: checking if workspace exists: %w", err)
+	}
+	if !exists {
+		return schema, nil
+	}
+
+	schema, err = b.client.Plugins.GetFullSchema(b.ctx, &pluginName)
 	if err != nil {
 		return schema, err
 	}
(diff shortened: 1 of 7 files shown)


file/builder.go
```diff
@@ -807,7 +807,16 @@ func (b *stateBuilder) getPluginSchema(pluginName string) (map[string]interface{
 	if schema, ok := b.schemasCache[pluginName]; ok {
 		return schema, nil
 	}
-	schema, err := b.client.Plugins.GetFullSchema(b.ctx, &pluginName)
+
+	exists, err := utils.WorkspaceExists(b.ctx, b.client)
+	if err != nil {
+		return nil, fmt.Errorf("getPluginSchema: checking if workspace exists: %w", err)
```
hbagdi, line 813:
```suggestion
		return nil, fmt.Errorf("ensure workspace exists: %w", err)
```

utils/defaulter.go
```diff
@@ -225,7 +225,11 @@ func getKongDefaulterWithClient(ctx context.Context, opts DefaulterOpts) (*Defau
 // 3. schema defaults coming from Admin API (excluded Konnect)
 // 4. hardcoded defaults under utils/constants.go (Konnect-only)
 func GetDefaulter(ctx context.Context, opts DefaulterOpts) (*Defaulter, error) {
-	if opts.Client != nil && !opts.DisableDynamicDefaults {
+	exists, err := WorkspaceExists(ctx, opts.Client)
+	if err != nil {
+		return nil, fmt.Errorf("defaulter: checking if workspace exists: %w", err)
```
hbagdi, line 230: same

hbagdi (commented):
Is it possible to add an e2e test for this patch?

file/builder.go
GGabriele: done

utils/defaulter.go
GGabriele: done

GGabriele:
> addressing comments

done: https://github.com/Kong/deck/pull/608/commits/452c490830dce24ddd058450779a2c8924838cab

tests/integration/diff_test.go
```diff
@@ -0,0 +1,32 @@
...
+	"testing"
+
+	"github.com/kong/deck/utils"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_Diff_Workspace(t *testing.T) {
+	tests := []struct {
+		name          string
+		kongFile      string
```
hbagdi, line 15:
```suggestion
		stateFile      string
```

hbagdi:
While this doesn't throw an error and gets the user past the creation, this is not enough.
Kong must be able to serve default outside the workspace context.
This change has the potential to surprise some users: depending on the existence of the workspace, decK diff may error out or may not error out.

Having said that, please merge this to get moving.

GGabriele:
@hbagdi can you please formally approve this?

hbagdi (approved):

PR #613, opened 2022-03-15
Add basic validation for restricted fields

aboudreault:
I spend some time yesterday to implement a validation for #421. It could have been simpler by just checking the object fields directly with if statements, but I was interested to learn a bit about the **reflect** module.

To check; this implementation do the validation *after* the user input injection, which mean that the defaults in `constants.go` can still be applied without validation. 

```shell
$ ./deck validate
Error: building state: creating defaulter: 3 errors occurred:
        service defaults cannot have these restricted fields set: id, name
        route defaults cannot have these restricted fields set: id, name
        upstream defaults cannot have these restricted fields set: id
        ...
```

TODO:

- [ ] I think the CI failure is because my branch is in my personal fork

### DIFF
--- a/utils/constants.go
+++ b/utils/constants.go
@@ -5,7 +5,6 @@ import (
 )
 
 const (
-//	defaultPort        = 80
 	defaultTimeout     = 60000
 	defaultSlots       = 10000
 	defaultWeight      = 100
@@ -71,10 +70,10 @@ var (
 		HashOnCookiePath: kong.String("/"),
 	}
 	defaultsRestrictedFields = map[string][]string{
-		"service": {"ID", "Host", "Name", "Port"},
-		"route": {"ID", "Name"},
-		"target": {"ID"}, // TODO check what is target4..
-		"upstream": {"ID", "Name"},
+		"Service":  {"ID", "Host", "Name", "Port"},
+		"Route":    {"ID", "Name"},
+		"Target":   {"ID", "Target"},
+		"Upstream": {"ID", "Name"},
 	}
 )
 
--- a/utils/constants.go
+++ b/utils/constants.go
@@ -5,7 +5,7 @@ import (
 )
 
 const (
-	defaultPort        = 80
+//	defaultPort        = 80
 	defaultTimeout     = 60000
 	defaultSlots       = 10000
 	defaultWeight      = 100
@@ -14,7 +14,6 @@ const (
 
 var (
 	serviceDefaults = kong.Service{
-		Port:           kong.Int(defaultPort),
 		Protocol:       kong.String("http"),
 		ConnectTimeout: kong.Int(defaultTimeout),
 		WriteTimeout:   kong.Int(defaultTimeout),
@@ -70,6 +69,12 @@ var (
 		HashOn:           kong.String("none"),
 		HashFallback:     kong.String("none"),
 		HashOnCookiePath: kong.String("/"),
+	}
+	defaultsRestrictedFields = map[string][]string{
+		"service": {"ID", "Host", "Name", "Port"},
+		"route": {"ID", "Name"},
+		"target": {"ID"}, // TODO check what is target4..
+		"upstream": {"ID", "Name"},
 	}
 )
 
--- a/utils/defaulter.go
+++ b/utils/defaulter.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 
 	"github.com/imdario/mergo"
 	"github.com/kong/go-kong/kong"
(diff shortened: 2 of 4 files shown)


utils/constants.go
```diff
@@ -71,6 +70,12 @@ var (
 		HashFallback:     kong.String("none"),
 		HashOnCookiePath: kong.String("/"),
 	}
+	defaultsRestrictedFields = map[string][]string{
+		"service": {"ID", "Host", "Name", "Port"},
+		"route": {"ID", "Name"},
+		"target": {"ID"}, // TODO check what is target4..
```
GGabriele, line 76: `target4` may easily be a typo, I think harry was referring to the `target` field of the `target` entity.

utils/defaulter.go
```diff
@@ -44,13 +45,20 @@ func getKongDefaulter(opts DefaulterOpts) (*Defaulter, error) {
 	if err := d.populateDefaultsFromInput(opts.KongDefaults); err != nil {
 		return nil, err
 	}
+
+	// TODO check if it's OK to do the validation only for the user input
+	err := d.validateKongDefaults()
```
GGabriele, line 50: Since this is about checking whether the user is setting restricted fields defaults, what do you think of moving this call within the `populateDefaultsFromInput` method?

utils/defaulter.go
```diff
@@ -69,6 +77,56 @@ func getKongDefaulter(opts DefaulterOpts) (*Defaulter, error) {
 	return d, nil
 }
 
+// Check if `entity` has restricted fields set
+func checkEntityDefaults(entity interface{}, restrictedFields []string) error {
```
GGabriele, line 81: nice!

utils/constants.go
```diff
@@ -14,7 +14,6 @@ const (
 
 var (
 	serviceDefaults = kong.Service{
-		Port:           kong.Int(defaultPort),
```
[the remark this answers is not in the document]
GGabriele: Yeah I think it's fine removing it, as we are restricting that field

utils/defaulter.go
```diff
@@ -44,13 +45,20 @@ func getKongDefaulter(opts DefaulterOpts) (*Defaulter, error) {
 	if err := d.populateDefaultsFromInput(opts.KongDefaults); err != nil {
 		return nil, err
 	}
+
+	// TODO check if it's OK to do the validation only for the user input
```
GGabriele, line 49: What do you mean with this?

aboudreault: not applicable anymore with the refactor

utils/defaulter.go
aboudreault: +1, I refactored **validateKongDefaults** and moved the check in **populateDefaultsFromInput**

utils/defaulter.go
```diff
@@ -69,6 +77,56 @@ func getKongDefaulter(opts DefaulterOpts) (*Defaulter, error) {
...
+	}
+	if len(invalidFields) > 0 {
+		return fmt.Errorf("cannot have these restricted fields set: %s", strings.Join(invalidFields, ", "))
+	}
+	return nil
+}
+
+func (d *Defaulter) validateKongDefaults() error {
+	var errs ErrArray
+	// TODO ideally, I would like to avoid this code duplication, but got some
```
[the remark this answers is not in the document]
aboudreault: fixed with the refactor of validateKongDefaults

utils/defaulter.go
```diff
@@ -277,6 +301,28 @@ func (d *Defaulter) populateDefaultsFromInput(defaults interface{}) error {
...
+			entityErr := fmt.Errorf(
+				"%s defaults %s", objectName, err)
+			errs.Errors = append(errs.Errors, entityErr)
+		}
+	}
+	if errs.Errors != nil {
+		return errs
+	}
+	return nil
+}
```
aboudreault, line 324: I feel there is a little bit too much of reflection, but that's mainly caused by https://github.com/Kong/deck/blob/main/utils/defaulter.go#L27 (which I'm not sure why it is an interface{} and not a normal struct like in the file builder).

utils/defaulter_test.go
```diff
@@ -572,3 +573,160 @@ func TestGetDefaulter_Konnect(t *testing.T) {
...
+			},
+			restrictedFields: []string{"ID", "Name", "Port"},
+			invalidFields:    []string{"ID", "Name", "Port"},
+		},
+	}
+
+	for _, tC := range testCases {
+		t.Run(tC.desc, func(t *testing.T) {
+			err := checkEntityDefaults(tC.entity, tC.restrictedFields)
+			if len(tC.invalidFields) > 0 {
```
GGabriele, line 618: What about having two more fields in the testcase struct like these:

```
		wantErr          bool
		expectedErr      string
```

and refactor this like:

```
if (err != nil) != tC.wantErr {
	t.Errorf("got error = %v, expected error = %v", err, tC.wantErr)
}
if tC.expectedErr != "" {
	assert.Equal(err.Error(), tC.expectedErr)
}
```

This makes you test the whole error string

utils/defaulter.go
GGabriele: Changing [this]((https://github.com/Kong/deck/blob/main/utils/defaulter.go#L27) would be a bit more involving, so if you feel this can benefit from a bit of refactoring turning it into a proper struct (I agree), please open an GH issue for tracking and let's tackle it separately!

utils/defaulter_test.go
aboudreault: I agree! fixing this.

PR #617, opened 2022-03-16
feat: add --skip-ca-certificates flag

GGabriele:
Since CA certificates are 'global' entities in Kong,
they cannot be managed on a per-workspace basis,
making it hard to be handled declaratively with decK.

This introduces a new `--skip-ca-certificates` to
sync/dump/diff/reset to make sure CA certs are
ignored when needed.


tests/integration/sync_test.go
```diff
@@ -852,3 +770,37 @@ func Test_Sync_RateLimitingPlugin(t *testing.T) {
...
+			expectedState: utils.KongRawState{
+				Services:       svc1_207,
+				CACertificates: []*kong.CACertificate{},
+			},
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			kong.RunWhenKong(t, ">=2.7.0")
```
rainest, line 798: Is there a reason that this is gated at 2.7? CA certificates were added [much earlier](https://github.com/Kong/kong/blob/master/CHANGELOG.md#core-42) and I'm not aware of any API difference that'd affect the way this flag operates since.

tests/integration/reset_test.go
```diff
@@ -0,0 +1,65 @@
...
+			kongFile: "testdata/reset/001-skip-ca-cert/kong.yaml",
+			expectedState: utils.KongRawState{
+				CACertificates: []*kong.CACertificate{caCert},
+			},
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			kong.RunWhenKong(t, ">=2.7.0")
```
rainest, line 56: Ditto other version comment.

rainest (changes requested):
`diff` should have parity with sync for the most part, but lacks support for this flag:

```
14:24:24-0700 esenin $ /tmp/bdeck diff -s /tmp/empty.yaml                       
deleting ca-certificate b368ce04-cacd-4d0c-8811-9c75a5b7fab5
deleting ca-certificate a37fcb6d-a452-42c9-989f-cd5c49c0828f
Summary:
  Created: 0
  Updated: 0
  Deleted: 2
14:24:27-0700 esenin $ /tmp/bdeck sync -s /tmp/empty.yaml --skip-ca-certificates
Summary:
  Created: 0
  Updated: 0
  Deleted: 0
14:24:29-0700 esenin $ /tmp/bdeck diff -s /tmp/empty.yaml --skip-ca-certificates    
Error: unknown flag: --skip-ca-certificates
```

Other than that and the test version question, looks good to go.

GGabriele:
> `diff` should have parity with sync for the most part, but lacks support for this flag:

Ops, I forgot to git add it :\ [did it now](https://github.com/Kong/deck/pull/617/files#diff-e87bbc6e9aa30388a1a17e258a5a3beef59dae6caf33f6d8e963d89e17348ac1R64)

tests/integration/sync_test.go
GGabriele: I did it mostly to avoid having N tests for this, due to schema differences on various version. If you think that's needed/good to have, I'll split the tests per-version

rainest: What are the differences? If the flag functions the same either way and it's just a matter of the input/output objects not being identical across all versions, fine to leave the gate in place as-is, just add a comment to the effect of "ca_certificates first appeared in 1.3, but we limit to 2.7+ here because the schema changed and the entities aren't the same across all versions, even though the skip functionality works the same."

GGabriele: Exactly. Done.

rainest (approved):

PR #629, opened 2022-03-24
doc: add code overview

rainest:
This is a copy of the onboarding overview document I presented earlier this year. I originally stuffed it in Confluence, but it's probably more discoverable (and accessible to non-Kong contributors) here.


GGabriele (approved):
thanks!

PR #642, opened 2022-04-13
docs: customize docs generation

GGabriele:


lena-larionova:
Thanks for working on this, this already looks way better!

For flags that don't have default values, looks like a blank default entry gets generated. Eg:

```
`--ca-cert`
:  Custom CA certificate (raw contents) to use to verify Kong's Admin TLS certificate.
This value can also be set using DECK_CA_CERT environment variable.
This takes precedence over --ca-cert-file flag. (default "")
```

lena-larionova:
Can we format the "default" entries with "default" capitalized, and backticks for the value? Eg:

>  (Default: `true`)

lena-larionova:
For any flag, the `--` characters get converted into just `-` in the output.

So this entry:

```
--analytics=false
```

Becomes:

Screen Shot 2022-04-14 at 1 17 58 PM

lena-larionova:
Can you convert any uses of single or double quotes in the flag descriptions into backticks?

For example, from `deck sync`, currently we get this:

````
--db-update-propagation-delay`
:  artificial delay (in seconds) that is injected between insert operations 
for related entities (usually for Cassandra deployments).
See 'db_update_propagation' in kong.conf. (default 0)
```
Where `'db_update_propagation'` generates as:
![Screen Shot 2022-04-14 at 1 17 58 PM](https://user-images.githubusercontent.com/54370747/163470491-3842687a-5b76-4a94-9672-d6d9606a26cb.png)

But what we want is this:
````
--db-update-propagation-delay`
:  artificial delay (in seconds) that is injected between insert operations 
for related entities (usually for Cassandra deployments).
See `db_update_propagation` in kong.conf. (default 0)
```
![Screen Shot 2022-04-14 at 1 32 51 PM](https://user-images.githubusercontent.com/54370747/163470844-0da8f6ae-de4e-47ce-8b9b-510b9f3d24ab.png)


That might cause issues for contractions though - so maybe the real answer is that the source files need to use double quotes (or backticks, if that's not an issue) consistently.

GGabriele:
@lena-larionova I think I addressed all your comments, some of them were related more to the doc strings and not with the docs generation itself.

Let me know what you think!

lena-larionova (approved):
It's looking great! Just one small issue: for `deck konnect diff` and `deck konnect dump`, the second paragraph of the description is indented, so it automatically generates a codeblock: 

Screen Shot 2022-04-15 at 11 15 57 AM

here's the markdown:

```
The konnect diff command is similar to a dry run of the 'deck konnect sync' command.

	It loads entities from Konnect and performs a diff with
	the entities in local files. This allows you to see the entities
	that will be created, updated, or deleted.

WARNING: This command is currently in alpha state. This command
might have breaking changes in future releases.
```

Honestly though - if we're deprecating these commands anyway, people shouldn't be looking at them going forward, and eventually we won't have a doc for them. So this is a negligible fix, you can leave it as-is if you want 😄 

Otherwise, LGTM. Thanks for putting this work in, it's a huge improvement.

PR #649, opened 2022-04-19
feat: add deprecation notice to 'convert' for konnect

GGabriele:
```
$ ./deck convert --from kong-gateway --to konnect --input-file kong.yaml --output-file kong.konnect.yaml
Notice: The `konnect` format type has been deprecatedas of v1.12.
Please use the `kong-gateway` format with `deck <cmd>` if you would like
to declaratively manage your Kong gateway config in Konnect.
```

### DIFF
--- a/cmd/convert.go
+++ b/cmd/convert.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/kong/deck/convert"
+	"github.com/kong/deck/cprint"
 	"github.com/kong/deck/utils"
 	"github.com/spf13/cobra"
 )
@@ -43,6 +44,11 @@ can be converted into a 'konnect' configuration file.`,
 			err = convert.Convert(convertCmdInputFile, convertCmdOutputFile, sourceFormat, destinationFormat)
 			if err != nil {
 				return fmt.Errorf("converting file: %v", err)
+			}
+			if convertCmdDestinationFormat == "konnect" {
+				cprint.UpdatePrintf("Warning: konnect format type was deprecated in v1.12 and it will be removed\n" +
+					"in a future version. Please use your Kong configuration files with deck <cmd>.\n" +
+					"Please see https://docs.konghq.com/konnect/deployment/import.\n")
 			}
 			return nil
 		},


cmd/convert.go
```diff
@@ -44,6 +45,11 @@ can be converted into a 'konnect' configuration file.`,
 			if err != nil {
 				return fmt.Errorf("converting file: %v", err)
 			}
+			if convertCmdDestinationFormat == "konnect" {
+				cprint.UpdatePrintf("Notice: The `konnect` format type has been deprecated" +
+					"as of v1.12.\nPlease use the `kong-gateway` format with `deck <cmd>` " +
+					"if you would like \nto declaratively manage your Kong gateway config in Konnect.\n")
```
hbagdi, line 51: @lena-larionova Can we provide a link to a documentation page here as well?

lena-larionova: Do you want it linking to the CLI reference or an actual guide? 

The CLI ref link would be: https://docs.konghq.com/deck/latest/reference/deck/

The guide link would be: https://docs.konghq.com/konnect/deployment/import/
Note that the guide link won't be live until Helium goes GA. The temp preview site is here: https://konnect--kongdocs.netlify.app/konnect/deployment/import/

hbagdi: Let's put in the live link. The window between GA and this going out is not huge and this is only useful once we GA the new feature.

GGabriele: Added! Let me know if that's fine now.

cmd/convert.go
```diff
@@ -44,6 +45,12 @@ can be converted into a 'konnect' configuration file.`,
 			if err != nil {
 				return fmt.Errorf("converting file: %v", err)
 			}
+			if convertCmdDestinationFormat == "konnect" {
+				cprint.UpdatePrintf("Notice: The `konnect` format type has been deprecated" +
+					"as of v1.12.\nPlease use the `kong-gateway` format with `deck <cmd>` " +
+					"if you would like \nto declaratively manage your Kong gateway config in Konnect.\n" +
+					"See https://docs.konghq.com/konnect/deployment/import/\n")
```
hbagdi, line 52: Warning: `konnect` format type was deprecated in v1.12 and it will be removed in a future version. Please use your Kong configuration files with `deck <cmd>`. Please see https://docs.konghq.com/konnect/deployment/import.

With appropriate `\n`s.

hbagdi:
The correct commit type is `feat` and not `docs` since this is a change to the code.

Please merge after resolving the above comment.

PR #654, opened 2022-04-22
feat: inject deck version into User-Agent header for konnect

GGabriele:


hbagdi:
Will this inject the user-agent header into the auth requests?

GGabriele:
> Will this inject the user-agent header into the auth requests?

Yes, the headers are part of the [konnectConfig](https://github.com/Kong/deck/blob/konnect_headers/cmd/common_konnect.go#L59) which are injected [here](https://github.com/Kong/deck/blob/konnect_headers/utils/types.go#L221)

hbagdi (approved):

PR #661, opened 2022-04-26
tests: add integration tests for Konnect

GGabriele:


aboudreault (approved):
LGTM. Great to see these konnect integration tests. To run locally, what is the konnect addr should I use?

GGabriele:
> To run locally, what is the konnect addr should I use?

The default address is fine, but you need to pass some env variables to "trigger" the test against Konnect: https://github.com/Kong/deck/blob/a3752dc0e2ae32a27d10dd076743d407babad04f/tests/integration/test_utils.go#L28-L30

Something like this:

```
$ DECK_KONNECT_EMAIL=$KONNECT_EMAIL DECK_KONNECT_PASSWORD=$KONNECT_PASSWORD go test -run "^(Test_Sync_Upstreams_Target_ZeroWeight_Konnect)$" -count 1  ./tests/integration/...
```

PR #652, opened 2022-04-21
feat: inject deck version into User-Agent header

GGabriele:
Before this change:

```
$ deck ping

<-- kong logs -->

172.19.0.1 - "GET / HTTP/1.1" 200 17479 "-" "Go-http-client/1.1"
```

After this change:

```
$ deck ping

<-- kong logs -->

172.19.0.1 - "GET / HTTP/1.1" 200 17479 "-" "decK/v1.11.0"
```

PR #668, opened 2022-05-03
doc: add unreleased 1.13 changelog

rainest:
Adds an unreleased 1.13 changelog. Per https://github.com/Kong/deck/pull/662#discussion_r862110930 we should start adding these as we make PRs to reduce release tasks and ensure we don't miss important changes when reviewing older PRs prior to release.

The changelog items include user-facing changes since 1.12 based on my review of closed PRs as of today.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Table of Contents
 
+- [v1.13.0](#v1130)
 - [v1.12.0](#v1120)
 - [v1.11.0](#v1110)
 - [v1.10.0](#v1100)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -37,6 +38,33 @@
...
+- The `konnect` format for the `convert` command is deprecated and will be
+  removed in a future release. Switch to using the `kong-gateway` format when
+  interacting with Konnect.
+  [#649](https://github.com/Kong/deck/pull/649)
+
+### Added
+
+- decK now uses its own user-agent header value instead of the generic
+  Go-http-client agent string.
+  [#652](https://github.com/Kong/deck/pull/652)
```
GGabriele, line 56: #652 was already part of 1.12, but we added #654 to have this with konnect too

CHANGELOG.md
```diff
@@ -37,6 +38,33 @@
 - [v0.2.0](#v020)
 - [v0.1.0](#v010)
 
+## [v1.13.0]
+
+> Release date: unreleased/TBD
+
+### Breaking changes
+
+- The `konnect` format for the `convert` command is deprecated and will be
```
GGabriele, line 47: This was already part of 1.12. Not sure if I'd call it "breaking change" just yet, as prior 1.11 functionality was not broken, but a bunch of warnings are raised in preparation for the actual breaking change that will happen with a release in the future

GGabriele (approved):
thanks!

CHANGELOG.md
rainest: We've raised deprecations in breaking changes previously, I think mostly to make it prominent and to get users to switch over early. Removed it though since it was actually in 1.12.

PR #678, opened 2022-05-10
fix: leave previous single arch binaries

1newsr:
I install deck using [asdf](https://github.com/nutellinoit/asdf-deck) and need the previous format of release such as `deck_${version}_darwin_amd64.tar.gz` or `deck_${version}_darwin_arm64.tar.gz`.

Please let me fix it to leave the previous release format.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,5 +1,5 @@
 universal_binaries:
-- replace: true
+- replace: false
 builds:
 - env:
   - CGO_ENABLED=0


.goreleaser.yml
```diff
@@ -1,5 +1,5 @@
 universal_binaries:
-- replace: true
+- replace: false
```
1newsr, line 2: document: https://goreleaser.com/customization/universalbinaries/

1newsr:
@GGabriele Could you please review this PR?

GGabriele:
Hi @1newsr , thanks for your contribution!

I have a question: why can't `asdf` be patched to work with universal binaries instead?

1newsr:
Thank you for your comment!
It is possible to modify `asdf` to work with universal binaries instead.
I created this PR  because it was the simplest and fastest way to use new `deck` version  with `asdf`.
If this PR is rejected, I will create a patch for `asdf`.

hbagdi:
@1newsr Could you please check and see if asdf can be patched in this case?
Universal binaries are easier to maintain and work with for us as well as users (no need to care about the underlying CPU arch) and we would like to keep it that way.

PR #681, opened 2022-05-13
chore: bump changelog for v1.12.1

GGabriele:


rainest (approved):
LGTM--we have added stuff but nothing that changes functional behavior, so patch bump is fine.

PR #705, opened 2022-06-24
feat: rate-limit konnect requests when receiving 429s

GGabriele:


go.mod
```diff
@@ -35,6 +35,8 @@ require (
 	k8s.io/code-generator v0.24.2
 )
 
+require github.com/hashicorp/go-retryablehttp v0.7.1
```
hbagdi, line 38: Please move this into the first require block that starts at line 9.

utils/types.go
```diff
@@ -119,6 +125,75 @@ func (kc *KongClientConfig) ForWorkspace(name string) KongClientConfig {
...
+		return true, nil
+	}
+	return false, nil
+}
+
+func getRetryableClient(client *http.Client) *http.Client {
+	const (
+		minRetryWait = 1 * time.Second
+		maxRetryWait = 10 * time.Second
+		retryMax     = 10
```
hbagdi, line 183:
```suggestion
		minRetryWait = 10 * time.Second
		maxRetryWait = 60 * time.Second
		retryMax     = 10
```

Let's start with a bit more conservative values that we can tune later on.

hbagdi (commented):
Please get a +1 from @tjasko.

utils/types.go
```diff
@@ -119,6 +125,75 @@ func (kc *KongClientConfig) ForWorkspace(name string) KongClientConfig {
...
 
+// backoffStrategy provides a callback for Client.Backoff which
+// will perform exponential backoff based on the attempt number and limited
+// by the provided minimum and maximum durations.
+//
+// It also tries to parse Retry-After response header when a http.StatusTooManyRequests
+// (HTTP Code 429) is found in the resp parameter. Hence it will return the number of
+// seconds the server states it may be ready to process more requests from this client.
+//
+// This is the same as DefaultBackoff (https://github.com/hashicorp/go-retryablehttp/blob/master/client.go#L510)
```
tjasko, line 136: Let's update this link so that it doesn't go out of date:
https://github.com/hashicorp/go-retryablehttp/blob/v0.7.1/client.go#L503

utils/types.go
```diff
@@ -119,6 +125,75 @@ func (kc *KongClientConfig) ForWorkspace(name string) KongClientConfig {
...
+// except that here we are only retrying on 429s.
+func backoffStrategy(min, max time.Duration, attemptNum int, resp *http.Response) time.Duration {
+	const (
+		base            = 10
+		bitSize         = 64
+		baseExponential = 2
+	)
+	if resp != nil {
+		if resp.StatusCode == http.StatusTooManyRequests {
```
tjasko, line 145: nit: Can be optimized to:
```suggestion
	if resp != nil && resp.StatusCode == http.StatusTooManyRequests {
```

tjasko (commented):
Changes seem sane, however, aren't we waiting on hashicorp/go-retryablehttp#168 to get merged?

GGabriele:
> Changes seem sane, however, aren't we waiting on [hashicorp/go-retryablehttp#168](https://github.com/hashicorp/go-retryablehttp/pull/168) to get merged?

That's not a blocker, because the way decK uses the underlying library via `go-kong` cannot hit that corner case.

tjasko (dismissed):

PR #721, opened 2022-07-19
tests: restrict plugins test due to schema changes

GGabriele:


rainest (commented):
Do we know what that change was? Can we add a comment indicating it and/or a new 3.0 test? We may opt to defer that for now since 3.0 isn't actually out yet, but should have a follow-up issue to investigate and document the new behavior if so.

rainest (approved):

Issue #655, opened 2022-04-23
Remove port as a restricted default field for service


dancallaghan:
I was just wondering the reasoning behind making port a restricted default field for services? We currently have `protocol: https` set as the default for our service, however we can't set `port: 443` as a default as it is listed as a restricted field.

Related PRs/Issues:
* #613
* #421

GGabriele:
Hi @dancallaghan ,
This has been fixed and released in `v.1.14.0`.

Sorry for the delay!

PR #724, opened 2022-08-01
feat: allow setting defaults for services' host and port

GGabriele:
solves https://github.com/Kong/deck/issues/655

### DIFF
--- a/file/builder_test.go
+++ b/file/builder_test.go
@@ -17,6 +17,7 @@ import (
 )
 
 const (
+	defaultPort        = 80
 	defaultTimeout     = 60000
 	defaultSlots       = 10000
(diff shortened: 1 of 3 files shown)


rainest (commented):
For future reference, this default does align with upstream; it does exist and is 80: https://github.com/Kong/kong/blob/9d06df08161a978296496d0e09e2ed9ec5c093a1/kong/db/schema/entities/services.lua#L18

That said, do we need the explicit constant with https://github.com/Kong/deck/pull/573? It seems like we'd just need to remove it from `defaultsRestrictedFields`, since the original inclusion there does appear arbitrary.

rainest (approved):

PR #725, opened 2022-08-01
feat: Personal Access Tokens auth with Konnect

GGabriele:


konnect/login_service.go
```diff
@@ -77,6 +77,24 @@ func (s *AuthService) LoginV2(ctx context.Context, email,
...
+		authResponse AuthResponse
+	)
+	if email != "" && password != "" {
+		authResponse, err = s.basicAuth(ctx, email, password)
+		if err != nil {
+			return AuthResponse{}, fmt.Errorf("basic authentication: %v", err)
+		}
+	} else if token != "" {
+		s.client.token = token
+	}
```
hbagdi, line 97: else error out?

utils/types.go
```diff
@@ -234,6 +237,9 @@ func GetKongClient(opt KongClientConfig) (*kong.Client, error) {
 	if err != nil {
 		return nil, fmt.Errorf("parsing headers: %w", err)
 	}
+	if opt.BearerToken != "" {
+		headers.Add("Authorization", "Bearer "+opt.BearerToken)
+	}
```
hbagdi, line 242: Avoid konnect-specific additions in Kong client (unless a technical addition is required).
You can re-use opt.Headers for this purpose.

cmd/root.go
```diff
@@ -175,6 +175,17 @@ It can be used to export, import, or sync entities to Kong.`,
...
 		rootCmd.PersistentFlags().Lookup("konnect-password-file"))
 
+	rootCmd.PersistentFlags().String("konnect-token", "",
+		"Personal Access Token associated with your Konnect account, "+
+			"this takes precedence over `--konnect-token-file` flag.")
+	viper.BindPFlag("konnect-token",
+		rootCmd.PersistentFlags().Lookup("konnect-token"))
+
+	rootCmd.PersistentFlags().String("konnect-token-file", "",
+		"File containing the PAT to your Konnect account.")
```
hbagdi, line 185: Remove acronym here.

cmd/root.go
```diff
@@ -329,9 +340,26 @@ func initKonnectConfig() error {
...
+		fileContent, err := ioutil.ReadFile(tokenFile)
+		if err != nil {
+			return fmt.Errorf("read file %q: %w", tokenFile, err)
+		}
+		if len(fileContent) == 0 {
+			return fmt.Errorf("file %q: empty", tokenFile)
+		}
+		token = string(fileContent)
+		token = strings.TrimRight(token, "\n")
+	}
```
hbagdi, line 357: Add a check to make sure the user supplies either a token or a (email and password).
Try using [this](https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagsMutuallyExclusive) to minimize if/else statements.

cmd/root.go
```diff
@@ -329,9 +340,26 @@ func initKonnectConfig() error {
 		password = strings.TrimRight(password, "\n")
 	}
 
+	token := viper.GetString("konnect-token")
+	tokenFile := viper.GetString("konnect-token-file")
+	// read from token file only if token is not supplied using an
+	// environment variable or flag
+	if token == "" && tokenFile != "" {
```
hbagdi, line 347: Would it be better to mark the flags as mutually exclusive instead?

hbagdi (changes requested):

utils/types.go
GGabriele: Good point, thanks for the tip

cmd/root.go
GGabriele: I added a mutually exclusive check, but it seems having these as env vars or coming from a deck config file would by-pass the check, so I think we still need something like this...so we either add more checks to enforce mutually exclusivity, or we keep this code which is also consistent with what we were already doing in regards to `konnect-password` and `konnect-password-file`

hbagdi: I see. Your call to mark these flags is exclusive or not since it is double work.

cmd/common_konnect.go
```diff
@@ -29,13 +29,18 @@ var addresses = []string{
 	defaultLegacyKonnectURL,
 }
 
-func authenticate(ctx context.Context, client *konnect.Client, host string) (konnect.AuthResponse, error) {
+var konnectAuthResponse konnect.AuthResponse
```
hbagdi, line 32: Yikes. This global variable seems very wrong.

Can you please instead change the return value for LoginV2 to be a custom struct that is populated no matter what authentication method is used?
And that can be populated using me API of Konnect.

Am I missing something?

hbagdi (changes requested):

cmd/root.go
GGabriele: For now I removed the cobra checks and kept the existing logic which is consistent with other parts of the code.

konnect/login_service.go
```diff
@@ -53,7 +54,7 @@ func (s *AuthService) Login(ctx context.Context, email,
 	return authResponse, nil
 }
 
-func (s *AuthService) LoginV2(ctx context.Context, email,
+func (s *AuthService) basicAuth(ctx context.Context, email,
```
hbagdi, line 57:
```suggestion
func (s *AuthService) sessionAuth(ctx context.Context, email,
```

IT is technically not basic-auth.

hbagdi (approved):

Issue #722, opened 2022-07-19
Configuration may trigger panic in upstream library


rainest:
Per https://github.com/Kong/kubernetes-ingress-controller/issues/2621#issuecomment-1182018847 there appears to be some condition where deck state comparisons will hit a bug in go-diff and panic. We do not currently have reproduction steps, but there is a very similar-looking failure in an upstream go-diff issue.

While there is a PR with a proposed fix upstream, the last upstream commit was in March 2021 and the PR has not received comment from a maintainer since it was filed in December 2021. As an upstream update appears unlikely, our options are likely:

- Make a fork of https://github.com/yudai/gojsondiff/ (which uses go-diff) and migrate it to the fixed version in https://github.com/iambus/go-diff (or our own fork of that fork). We may be able to just import the fixed dependency directly.
- Refactor around [go-cmp](https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff). This looks easy enough since we only use go-jsondiff [in one location](https://github.com/Kong/deck/blob/4c5fa764f7f9b5d4963c833aef69b686735f7950/diff/diff_helpers.go#L60-L85), and that function already receives Go structs before marshalling them to JSON. This does, however, change deck's diff output, and changes it to a less user-friendly format (since it will look like actual Go structs, not the JSON versions users will be familiar with from the admin API).

rainest:
We already had our own fork of gojsondiff, so I've replaced the original upstream with our fork incorporating iambus' fix in https://github.com/Kong/gojsondiff/pull/3. That will need a version bump after merge, after which we can bump it here and release a patch version of decK.

PR #726, opened 2022-08-01
chore(deps) update gojsondiff to v1.3.1

rainest:
Update gojsondiff to incorporate the fix from https://github.com/Kong/gojsondiff/pull/3

Fix https://github.com/Kong/deck/issues/722

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,6 +37,8 @@ require (
 )
 
 require (
+	github.com/Kong/go-diff v1.2.2 // indirect
+	github.com/Kong/gojsondiff v1.3.1 // indirect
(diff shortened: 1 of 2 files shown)


pmalek:
This https://github.com/Kong/gojsondiff/pull/3/files#r934732191 got me thinking. Why do we still need to reference `github.com/sergi/go-diff`?

rainest:
Looks like we left a bunch of references to upstream in the fork originally. Will need https://github.com/Kong/gojsondiff/pull/4 and another version bump in the fork, and that should purge this (it clears sergi/go-diff from Kong/gojsondiff so it _should_ cascade remove it here).

PR #731, opened 2022-08-10
chore: make nightly image tests conditional

rainest:
Require access to the nightly image credentials to test against it.


GGabriele (approved):
Thanks!

PR #755, opened 2022-09-20
fix: use escaped strings for services' path

GGabriele:
Right now, when a `url` containing an encoded whitespace (`%20`)
is passed to a service, decK unwraps it and replace it with an
actual whitespace. This is because decK uses `url.Parse` under
the cover, which is the one making this translation.

This commit makes decK URL unwrapping method using the `url.EscapedPath`
method to parse the correct path format.


javierguerragiraldez (commented):
on a quick reading of `EscapedPath()` source, it doesn't feel like the safest approach, as it checks if the raw url is "okish" to use, and if so, returns it directly.

my first approach would be to first distinguish two very different cases:
1.  "we're having too many issues with "special" characters" -> better to straight up encode.
2. "everything has been working well with decoded path, except this very specific thing" -> just do `path.replace(' ', '%20')`

GGabriele:
> on a quick reading of `EscapedPath()` source, it doesn't feel like the safest approach, as it checks if the raw url is "okish" to use, and if so, returns it directly.
> 
> my first approach would be to first distinguish two very different cases:
> 
> 1. "we're having too many issues with "special" characters" -> better to straight up encode.
> 2. "everything has been working well with decoded path, except this very specific thing" -> just do `path.replace(' ', '%20')`

Yeah I think we're falling mostly on the second case, which was also the initial way I was working it out.
Addressing this here: https://github.com/Kong/deck/pull/755/commits/7df6eaece6555a9e0138e264cbdca4fbba5d01fa

GGabriele:
@javierguerragiraldez would you mind taking another look whenever you have time?

javierguerragiraldez (approved):
LGTM.   don't forget to update the commit message, since this no longer uses EscapedPath.

PR #773, opened 2022-10-17
refactor(diff): allow custom printer funcs

vbehar:
when using deck as a lib, if we want to capture the operations logs (creations, updates, deletions), we need to reconfigure the global vars in the `cprint` package. this is an issue if we want to do different deck operations in parallel.

this change introduce new parameters to the Syncer, to configure custom printer funcs, without depending on the `cprint` package. Note that by default if no custom printer funcs are configured, the ones from the `cprint` package will be used - so no behaviour change.

### DIFF
--- a/diff/diff.go
+++ b/diff/diff.go
@@ -51,6 +51,10 @@ type Syncer struct {
 	silenceWarnings bool
 	stageDelaySec   int
 
+	createPrintln func(a ...interface{})
+	updatePrintln func(a ...interface{})
+	deletePrintln func(a ...interface{})
+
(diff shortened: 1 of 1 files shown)


vbehar:
Hi @GGabriele is there anything missing to get this one merged?
thanks a lot!

GGabriele:
Hi @vbehar ,

thanks a lot for your contribution and apologies for the delay in coming back to you!

This looks good to me. Merging it now!

Thanks again!

vbehar:
thanks!

PR #776, opened 2022-10-19
ci: use kong:master-alpine instead of kong:latest

GGabriele:


.github/workflows/integration-enterprise.yaml
```diff
@@ -16,6 +16,7 @@ jobs:
         - 'kong/kong-gateway:2.6.0.2'
         - 'kong/kong-gateway:2.7.0.0'
         - 'kong/kong-gateway:2.8.0.0'
+        - 'kong/kong-gateway:3.0.0.0'
```
mikefero, line 19: 3.0.0.0 switched the default distro from Alpine to Ubuntu; not sure if you wanted to specify the distro here and moving forward for 3.0+.

mikefero: Note: Official OSS 3.0.0 images are still defaulting to Alpine distro so all good there.

mikefero (approved):

PR #788, opened 2022-11-08
Feat/konnect defaults

GGabriele:


utils/defaulter.go
```diff
@@ -181,7 +191,16 @@ func (d *Defaulter) MustSet(arg interface{}) {
...
+		endpoint = fmt.Sprintf("/v1/schemas/json/%s", entityType)
+	}
+	req, err := d.client.NewRequest(http.MethodGet, endpoint, nil, nil)
+	if err != nil {
+		return schema, err
+	}
+	_, err = d.client.Do(d.ctx, req, &schema)
 	if err != nil {
 		return schema, err
 	}
```
aboudreault, line 206: we could remove this if and its following line and just do `return schema, err`

aboudreault (approved):
LGTM, minor nit, feel free to ignore.

utils/defaulter.go
GGabriele: Totally, good catch, thanks!

aboudreault (approved):

Issue #783, opened 2022-10-26
bug: failures when creating a workspace via deck sync


mklaber:
In a PostgreSQL-backed hybrid deployment of Kong Gateway 2.8.1.4, the `deck sync` command fails when the target workspace (either via `_workspace` in the deck file or the `--workspace` flag) does not yet exist.

When a deck file has a workspace that does not exist, it successfully creates the workspace and then fails any subsequent actions.

```bash
$ deck version
decK v1.15.1 (bb70f8f) 
$ deck sync -s kong/small_workspace.yaml
creating workspace small-workspace
Error: building state: add defaults to plugin 'kafka-log': retrieve schema for kafka-log from Kong: HTTP status 404 (message: "Workspace 'small-workspace' not found")
```

Re-running the exact same command a moment later successfully syncs the state file.

<details>
<summary>small_workspace.yaml</summary>

```yaml
_workspace: small-workspace
_format_version: '1.1'
plugins:
  - name: kafka-log
    config:
      producer_async: false
      topic: kongRequestLogRaw
      bootstrap_servers:
        - host: kafka-broker.service.company.com
          port: 9092
  - name: prometheus
    config:
      per_consumer: false
```

</details>

It seems a reasonable solution would be for `deck sync` to consider the `--db-update-propagation-delay` flag after creating a workspace [here](https://github.com/Kong/deck/blob/d51951a7dcf36428c0fa67efe1b6747dfcb0b36a/cmd/common.go#L197-L209). Based on the description of that flag ("artificial delay (in seconds) that is injected between insert operations for related entities"), it seems like an oversight that it's ignored at all (if not the root of this bug).

ludovic-pourrat:
Hello,

We are getting hits by this bug too any news / progress in this ? 

Best Regards.

PR #789, opened 2022-11-08
fix: add --db-workspace-update-propagation-delay CLI flag

GGabriele:
When some configuration is applied to a non-existent workspace, decK creates the workspace before proceeding with the other queries. In such cases, it may happen that Kong is not "fast" enough to propagate workspace changes, leading to some misleading 404s due to not-existing (not-propagated) workspaces.

This commit is adding a new --db-workspace-update-propagation-delay CLI flag to introduce an artificial delay when a new workspace is created.
This is different from the --db-update-propagation-delay flag, as the latter gets applied to every insert operation and it was initially introduced to address some Cassandra limitation.

Related to https://github.com/Kong/deck/issues/783

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"reflect"
 	"sort"
+	"time"
 
 	"github.com/blang/semver/v4"
 	"github.com/kong/deck/cprint"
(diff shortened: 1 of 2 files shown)


hbagdi:

PR #791, opened 2022-11-10
feat: enhance error messages during dump for plugins

mikefero:
This update adds additional information if a failure occurs when gathering referenced (or parent) resources; consumer, service, and route. The error message will contain the ID of the referenced resource that failed to be retrieved and include the plugin ID and name for easily identifying where the missing resource is referenced.


mikefero:
@GGabriele I think a lot of the other errors in this file could be wrapped in a similar fashion to give more feedback on the error that occurred. There are some instances where it wouldn't make sense to fully wrap the error as this would change how a lot of the code base is currently instrumented.

aboudreault (approved):

GGabriele (approved):

PR #806, opened 2022-12-13
fix: don't build diff doc for creates and deletes

GGabriele:
This commit fixes a 'visual' regression introduced in https://github.com/Kong/deck/pull/463, which is making decK print full diff documents for 'create' and 'delete' actions. This can result in very
verbose output in case of configuration files
containing hundreds/thousands of resources.


aboudreault (approved):
tested locally, works as expected. :+1:

PR #821, opened 2023-01-13
fix: do not error out on EE endpoints with no license

GGabriele:
Right now decK error out when the Consumer Groups and Vaults list endpoints are hit and no license is present in the Gateway.
In such case, decK should continue its flow.

### DIFF
--- a/dump/dump.go
+++ b/dump/dump.go
@@ -248,7 +248,7 @@ func getProxyConfiguration(ctx context.Context, group *errgroup.Group,
 			consumerGroups, err = GetAllConsumerGroups(ctx, client, config.SelectorTags)
 		}
 		if err != nil {
-			if kong.IsNotFoundErr(err) {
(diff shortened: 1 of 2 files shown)


dump/dump.go
```diff
@@ -248,7 +248,7 @@ func getProxyConfiguration(ctx context.Context, group *errgroup.Group,
 			consumerGroups, err = GetAllConsumerGroups(ctx, client, config.SelectorTags)
 		}
 		if err != nil {
-			if kong.IsNotFoundErr(err) {
+			if kong.IsNotFoundErr(err) || kong.IsForbiddenErr(err) {
```
aboudreault, line 251: since we are just returning `nil` will they see a warning about this too?

GGabriele: They will only see the status code in verbose mode. Do you think we should log a warning too?

aboudreault: could be a good idea, but we can keep this as is for this PR.

aboudreault (dismissed):

PR #839, opened 2023-02-06
fix: workspaces validation with multiple files

GGabriele:
When running `deck sync` with multiple state files, either by passing multiple `-s` flags
(e.g. `-s foo.yaml -s bar.yaml`) or by passing a folder (e.g. `-s ./statefiles`), decK merges the whole configuration from all files and it fails the content validation if different `_workspace` entries are found. While this is good and expected by design, decK also fails in the following case:

```
$ cat foo.yaml

services:
- name: svc1
  host: 1.example.com
  tags:
  - team-svc1
  routes:
  - name: r1
    paths:
    - /r1

$ cat meta.yaml

_format_version: "1.1"
_workspace: bar
```

The error from decK would be the following:

```
$ deck sync -s meta.yaml -s foo.yaml

Error: it seems like you are trying to sync multiple workspaces at the same time ([ bar]).
decK doesn't support syncing multiple workspaces at the same time, please sync one workspace at a time
```

The reason why decK is failing is because it's adding an empty workspace to the array used for validation because the `foo.yaml` state file is missing the `_workspace` entry.

This commit makes sure that the "emtpy workspace" is not considered at workspace validation time.

### DIFF
--- a/file/readfile.go
+++ b/file/readfile.go
@@ -34,7 +34,9 @@ func getContent(filenames []string) (*Content, error) {
 		if err != nil {
 			return nil, fmt.Errorf("reading file: %w", err)
 		}
-		workspaces = append(workspaces, content.Workspace)
+		if content.Workspace != "" {
+			workspaces = append(workspaces, content.Workspace)
+		}
 		err = mergo.Merge(&res, content, mergo.WithAppendSlice)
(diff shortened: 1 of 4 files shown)


aboudreault (approved):
LGTM!

PR #847, opened 2023-02-13
fix: enable decK's convert subcommand to work with Konnect again

GGabriele:
After the deprecation of the old `deck konnect` command and the old Konnect format, the `deck convert` command with the Konnect format is not useful anymore in its current shape because the old Konnect format is not compatible anymore with the new workflow handled directly with the main `deck` command.

Despite the new Konnect format is mostly compatible with the 'regular' Kong format, it has a few differences:
- it needs the `_format_version` entry set to `"3.0"`
- it doesn't support the `_workspace` entry

This means that doing a `deck dump` and then a `deck sync` from one Gateway with workspaces to Konnect may not natively work because of these incompatibilities.

This commit makes sure to support the new Konnect format in the `deck convert` command. The new convertion does the following:
- set the `_format_version` entry to `"3.0"`
- handle the removal of the `_workspace` entry and the addition of the `_konnect.runtime_group_name` in the following manner:
  1. if present, remove the `_workspace` entry
  2. if `--konnect-runtime-group-name foo` flag is used, then set `_konnect.runtime_group_name: foo`
  3. if no `--konnect-runtime-group-name` flag is used, then set `_konnect.runtime_group_name` to the value `_workspace` was precedently set to
  4. if no `_workspace` was present and no `--konnect-runtime-group-name` flag is used, then leave the whole `_konnect` entry unset

### DIFF
--- a/cmd/convert.go
+++ b/cmd/convert.go
@@ -5,7 +5,6 @@ import (
 	"os"
 
 	"github.com/kong/deck/convert"
-	"github.com/kong/deck/cprint"
 	"github.com/kong/deck/utils"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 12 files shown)


hbagdi:

PR #868, opened 2023-03-17
feat: parsing numbers in templates

zekth:
Follow-up of https://github.com/Kong/deck/pull/867

### DIFF
--- a/file/readfile_test.go
+++ b/file/readfile_test.go
@@ -451,6 +451,61 @@ func Test_getContent(t *testing.T) {
 			},
 			wantErr: true,
 		},
+		{
+			name: "file with env var and parse Int",
+			args: args{[]string{"testdata/parseint/file.yaml"}},
+			envVars: map[string]string{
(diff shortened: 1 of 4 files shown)


file/readfile_test.go
```diff
@@ -451,6 +451,61 @@ func Test_getContent(t *testing.T) {
...
+			envVars: map[string]string{
+				"DECK_FOO_FLOAT": "1337",
+			},
+			want: &Content{
+				Plugins: []FPlugin{
+					{
+						Plugin: kong.Plugin{
+							Name: kong.String("foofloat"),
+							Config: kong.Configuration{
+								"foo": *kong.Float64(1337),
```
zekth, line 493: weirdly the assertions seems to fail on this one if i don't use the value. However it works fine for the `parseInt`.

GGabriele: That's because plugins' config stores actual values, not pointers (like for other non-plugin entities). You can rewrite this as `float64(1337)`.

GGabriele (approved):

PR #869, opened 2023-03-17
fix: return proper error when HTTP calls fail on validate

GGabriele:


rainest (approved):

rainest:
Rebased to origin/main since the Konnect failures look unrelated, presumably a change upstream that was fixed after this was opened.

rainest (approved):

aboudreault (approved):

PR #874, opened 2023-03-23
feat(file): add render command

zekth:
Proposal of a new format for `convert` command. Possible things to address:
- [ ] Mocking the env variables
- [ ] compatibility with older kong versions


convert/convert.go
```diff
@@ -1,18 +1,24 @@
...
 	"github.com/kong/deck/file"
+	"github.com/kong/deck/state"
 	"github.com/kong/deck/utils"
 	"github.com/kong/go-kong/kong"
 )
 
 type Format string
 
 const (
+	// FormatDeck represents the Deck configuration format.
```
GGabriele, line 20: Can you please fix this to reflect the new refactor?

convert/convert.go
```diff
@@ -37,38 +43,47 @@ func ParseFormat(key string) (Format, error) {
...
 		return err
 	}
 
 	switch {
 	case from == FormatKongGateway && to == FormatKonnect:
-		outputContent, err = convertKongGatewayToKonnect(inputContent)
-		if err != nil {
-			return err
+		if len(inputFilenames) > 1 {
+			return fmt.Errorf("only one input could be sent to convert kong config")
```
GGabriele, line 68:
```suggestion
			return fmt.Errorf("only one input file can be provided when converting from Kong to Konnect format")
```

convert/convert.go
```diff
@@ -37,38 +43,47 @@ func ParseFormat(key string) (Format, error) {
...
 		}
+		outputContent, err = convertKongGatewayToKonnect(inputContent)
 	case from == FormatKongGateway2x && to == FormatKongGateway3x:
-		outputContent, err = convertKongGateway2xTo3x(inputContent, inputFilename)
-		if err != nil {
-			return err
+		if len(inputFilenames) > 1 {
+			return fmt.Errorf("only one input could be sent to convert kong config")
```
GGabriele, line 73:
```suggestion
			return fmt.Errorf("only one input file can be provided when converting from Kong 2.x to Kong 3.x format")
```

convert/convert.go
```diff
@@ -195,3 +210,29 @@ func removeServiceName(service *file.FService) *file.FService {
 	serviceCopy.ID = kong.String(utils.UUID())
 	return serviceCopy
 }
+
+func convertDistributedToKong3(
```
GGabriele, line 214: nit for consistency with existing code:

```suggestion
func convertDistributedToKong3x(
```

convert/convert.go
```diff
@@ -195,3 +210,29 @@ func removeServiceName(service *file.FService) *file.FService {
...
+	}
+	targetState, err := state.Get(rawState)
+	if err != nil {
+		return nil, err
+	}
+
+	// file.KongStateToFile calls file.WriteContentToFile
+	return file.KongStateToContent(targetState, file.WriteConfig{
+		Filename:    outputFilename,
+		FileFormat:  format,
```
GGabriele, line 235: Thinking more about this, I think it would be better to either avoid setting the `_format_version` OR setting it properly based on input from the user (`--to kong-gateway-2.x` or `--to kong-gateway-3.x`).

Imagine having some "distributed" 2.x configuration:

```
services:
- name: svc1
  host: mockbin.org
  routes:
  - name: r1
    paths:
    - /r1
 ```

If we use the new functionality, we'd get something like this:

```
$ ./deck convert --from distributed --to kong-gateway --input-file kong.yaml --yes

$ cat kong.yaml
_format_version: "3.0"
services:
- connect_timeout: 60000
  host: mockbin.org
  name: svc1
  protocol: http
  read_timeout: 60000
  routes:
  - name: r1
    paths:
    - /r1
    preserve_host: false
    protocols:
    - http
    - https
    regex_priority: 0
    strip_path: false
  write_timeout: 60000
```

The input and output files are not the same when deployed to Kong 2.x or 3.x.

I think this may lead to some confusion and undesired outcome for the user.

What do you think?

File: file/readfile.go [hunk 6]
```diff
@@ -146,25 +146,57 @@ func getPrefixedEnvVar(key string) (string, error) {
...
-		"env":     getPrefixedEnvVar,
-		"toBool":  toBool,
-		"toInt":   toInt,
-		"toFloat": toFloat,
-	}).Delims("${{", "}}")
+func toFloatMocked(_ string) (float64, error) {
+	return 42, nil
+}
+
+func renderTemplate(content string, mockEnvVars bool) (string, error) {
```
GGabriele, line 182: Very nice!

file/readfile.go
```diff
@@ -146,25 +146,57 @@ func getPrefixedEnvVar(key string) (string, error) {
...
 
-func renderTemplate(content string) (string, error) {
-	t := template.New("state").Funcs(template.FuncMap{
-		"env":     getPrefixedEnvVar,
-		"toBool":  toBool,
-		"toInt":   toInt,
-		"toFloat": toFloat,
-	}).Delims("${{", "}}")
+func toFloatMocked(_ string) (float64, error) {
+	return 42, nil
```
GGabriele, line 179: I believe we should document this somehow.

convert/convert.go
zekth: Pushed a change handling this.

GGabriele:
@zekth it seems a test is failing due to some nil reference:

```
=== RUN   Test_Convert/converts_from_distributed_to_kong_gateway_(no_deck_specific_fields)
--- FAIL: Test_Convert (0.24s)
    --- PASS: Test_Convert/errors_out_due_to_invalid_conversion (0.03s)
    --- PASS: Test_Convert/errors_out_due_to_invalid_conversion#01 (0.02s)
    --- PASS: Test_Convert/errors_out_when_a_nameless_service_is_present_in_the_input (0.01s)
    --- PASS: Test_Convert/errors_out_when_input_file_doesn't_exist (0.00s)
    --- PASS: Test_Convert/converts_from_Kong_Gateway_to_Konnect_format (0.09s)
    --- PASS: Test_Convert/converts_from_Kong_Gateway_2.x_to_Kong_Gateway_3.x_format (0.04s)
    --- PASS: Test_Convert/converts_from_Kong_Gateway_2.x_to_Kong_Gateway_3.x_format_(no__format_version_input) (0.04s)
    --- FAIL: Test_Convert/converts_from_distributed_to_kong_gateway_(no_deck_specific_fields) (0.01s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xb421ef]
```

Would you mind checking?

Tieske:
As per internal discussion, this functionality is to land in its own command "render", which is to be a subcommand of the new "file" command. 

Since the "file" command is introduced in #939 , it makes sense to target that branch for now, instead of `main`.

Tieske:
closing in favour of #963

PR #882, opened 2023-04-04
Add support for license entity

rainest:
Support the license entity.

Syncing your license via deck CLI seems a bit odd, but I guess you can do it. We need this more for KIC to be able to support it in DB-less mode, and KIC uses bits of deck code to handle that path in addition to DB mode. We may as well get all the rest of it.

The Konnect integration tests are occasionally failing, but those appear to be flakes due to propagation issues (logs will show a created entity and then get a 404 on deletion attempts).


types/license.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	return license, nil
+}
+
+// Update updates a License in Kong.
+// The arg should be of type crud.Event, containing the license to be updated,
+// else the function will panic.
+// It returns a the updated *state.License.
+func (s *licenseCRUD) Update(ctx context.Context, arg ...crud.Arg) (crud.Arg, error) {
+	event := crud.EventFromArg(arg[0])
```
pmalek, line 58: - Can we check if `arg` has at least 1 element?
- Why not just use `crud.Event` type?

types/license.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	return &state.License{License: *createdLicense}, nil
+}
+
+// Delete deletes a License in Kong.
+// The arg should be of type crud.Event, containing the license to be deleted,
+// else the function will panic.
+// It returns a the deleted *state.License.
+func (s *licenseCRUD) Delete(ctx context.Context, arg ...crud.Arg) (crud.Arg, error) {
+	event := crud.EventFromArg(arg[0])
```
pmalek, line 44: ditto

state/types.go
```diff
@@ -1470,3 +1470,48 @@ func (v1 *Vault) EqualWithOpts(v2 *Vault, ignoreID, ignoreTS bool) bool {
...
+// Equal returns true if l1 and l2 are equal.
+// TODO add compare array without position
+func (l1 *License) Equal(l2 *License) bool {
+	return l1.EqualWithOpts(l2, false, false)
+}
+
+// EqualWithOpts returns true if l1 and l2 are equal.
+// If ignoreID is set to true, IDs will be ignored while comparison.
+// If ignoreTS is set to true, timestamp fields will be ignored.
+func (l1 *License) EqualWithOpts(l2 *License, ignoreID, ignoreTS bool) bool {
```
pmalek, line 1501: Instead of passing bools which might get used incorrectly why not use options (as the name suggests)? `WithIgnoreID` and `WithIgnoreTS`? At this point this function can be merged with `Equal`.

WDYT?

state/license_test.go
```diff
@@ -0,0 +1,291 @@
...
+			},
+		},
+	}
+	for _, s := range licenses {
+		assert.Nil(collection.Add(s))
+	}
+
+	allLicenses, err := collection.GetAll()
+
+	assert.Nil(err)
```
pmalek, line 289:
```suggestion
	assert.NoError(err)
```

state/license_test.go
```diff
@@ -0,0 +1,291 @@
...
+	assert.Equal(len(licenses), len(allLicenses))
```
pmalek, line 290: Would this work?

```suggestion
	assert.Len(licenses, allLicenses)
```

state/license_test.go
```diff
@@ -0,0 +1,291 @@
...
+	err = collection.Delete("does-not-exist")
+	assert.NotNil(err)
+	err = collection.Delete("first")
+	assert.Nil(err)
+
+	err = collection.Delete("first")
+	assert.NotNil(err)
+
+	err = collection.Delete("")
+	assert.NotNil(err)
```
pmalek, line 262:
```suggestion
	err := collection.Add(license)
	assert. NoError(err)

	err = collection.Delete("does-not-exist")
	assert.Error(err)
	err = collection.Delete("first")
	assert. NoError(err)

	err = collection.Delete("first")
	assert.Error(err)

	err = collection.Delete("")
	assert.Error(err)
```

state/license_test.go
```diff
@@ -0,0 +1,291 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			got, err := k.Get(tc.args.nameOrID)
+			if (err != nil) != tc.wantErr {
+				t.Errorf("LicensesCollection.Get() error = %v, wantErr %v", err, tc.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(got, tc.want) {
+				t.Errorf("LicensesCollection.Get() = %v, want %v", got, tc.want)
+			}
```
pmalek, line 140:
```suggestion
			assert.Equal(t, tc.want, got)
```

state/license.go
```diff
@@ -0,0 +1,167 @@
...
+	defer txn.Abort()
+
+	var searchBy []string
+	searchBy = append(searchBy, *license.ID)
+	_, err := getLicense(txn, searchBy...)
+	if err == nil {
+		return fmt.Errorf("inserting license %v: %w", license.Console(), ErrAlreadyExists)
+	} else if err != ErrNotFound {
+		return err
+	}
```
pmalek, line 45: I believe if this was consistent with what's used in `LicensesCollection.Get()` it would be a bit more readable.

```suggestion
	if err == nil {
		if err == ErrNotFound {
			return err
		}
		return fmt.Errorf("inserting license %v: %w", license.Console(), ErrAlreadyExists)
	}
```

Secondly, does that actually work? Don't we need to use `errors.Is()`?

GGabriele: Also, I think the `if err == nil` check should actually be `if err != nil`.

types/license.go
rainest: This is yet more copy pasta where every entity type in deck looks like this, but in this case I can't offer much any explanation as to why it's like this. The boolean positional arguments are at least a simple design that probably grew naturally out of initially only having one toggle.

This I can see that `crud.Arg` is an `interface{}` alias, which it probably doesn't need to be, and that I don't immediately see from [relevant code in the differ](https://github.com/Kong/deck/blob/a7cf38f623ad488cfffc1ecab21f32043fe7c301/diff/diff.go#L302-L352) where it'd ever be anything other than an Event or when you'd have more than one Event arg. I don't see anything else being passed as args.

The arg length check isn't going to get us much other than a cleaner error that's not user actionable, as any case where it's _not_ at least one item points to a bug further up. Your only recourse is to deliver replication steps to us and have us figure out what we did wrong in the diff library. Panicking on out of bounds is correctly indicating that there's something horribly wrong elsewhere in our code.

My best guess is that this was intended for some sort of expanded functionality beyond Events, but that said functionality was unknown at the time of design, and ultimately it turned out to be a YAGNI expansion port.

No follow-up issue here since I can't even point to a clearly good refactor, and would probably suggest a broader redesign of the differ. The type-related code is at least simple if ugly, whereas the differ code that calls it is confusing, difficult to debug, and apparently full of unnecessary room for future expansion that does nothing but make it impossible to have stronger compile-time validation that would suffice for what deck actually needs to do.

state/license.go
rainest: Reverted this after initially accepting it. Unless I'm missing something, nesting the `if err == ErrNotFound` under the `if err == nil` means we never hit the former case, correct? Was there something else intended here?

This shouldn't match `Get()`, since the good and bad conditions here are the reverse of `Get()`: not found is what we want to see, whereas a successful `getLicense()` response is a failure condition.

`Get()`, however, looks like it can be simplified, since `getLicense()` returns its own `nil` license value on errors, and we don't need an `if err == ErrNotFound { return ErrNotFound }` if we're going to return any error after anyway.

The equivalency check only needs `Is()` or `As()` if we wrap it, correct? We don't; there's only the one possible string for `ErrNotFound`.

rainest:
> Strictly speaking, a License is not real Gateway configuration.

It is the way the gateway implemented it :smile: but yeah I think that'd cause issues with the majority of licenses, which are loaded via environment but still visible (and can be deleted from) the API endpoint.

Testing the flag is hindered by our lack of a dedicated test license, however. There's no license I can provide that's accepted by the license endpoint but also not an actual license. I'm unsure if we have permission to stick a previously-valid test license in place and/or if it will cause other havoc if so (it would probably actually load the expired license, and lock out config for other tests).

The `sync()` test utility doesn't let me test failed loads either to confirm that the flag attempts it when you set `--skip-licenses=false`. I'd originally added one with the broken example license and a check for a non-nil error, but that yielded:

```
=== RUN   Test_Sync_SkipCLicense_3x/syncing_with_--skip-licenses_should_ignore_licenses#01
creating license b824e7c0-d116-4c03-9e27-de05c5d30083
Summary:
  Created: 0
  Updated: 0
  Deleted: 0
Error: 1 errors occurred:
	while processing event: {Create} license b824e7c0-d116-4c03-9e27-de05c5d30083 failed: HTTP status 400 (message: "schema violation (payload: Unable to validate license: could not decode license json)")

    sync_test.go:1915: syncing a broken license did not result in error
```
The command successfully executes and there's no error returned to the test, only the error text output to stderr. Is there some way we could confirm that short of screen scraping it somehow? I don't think we can override the stderr output to a buffer we can check in the test even.

state/license.go
pmalek: > The equivalency check only needs `Is()` or `As()` if we wrap it, correct? We don't; there's only the one possible string for `ErrNotFound`.

Yes, but using `Is()` or `As()` will be future proof in case that would change. 🤷

state/license_test.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+	assert.Nil(k.Add(lic1))
+
+	lic1.Payload = kong.String("other")
+	assert.Nil(k.Update(lic1))
+
+	r, err := k.Get("foo-id")
+	assert.Nil(err)
+	assert.NotNil(r)
+	assert.Equal(*r.Payload, "other")
```
pmalek, line 238:
```suggestion
	require.NoError(t, err)
	require.NotNil(t, r)
	require.Equal(t, *r.Payload, "other")
```

state/license_test.go
```diff
@@ -0,0 +1,289 @@
...
+func TestLicenseUpdate(t *testing.T) {
+	assert := assert.New(t)
+	k := licensesCollection()
+	lic1 := License{
+		License: kong.License{
+			ID:      kong.String("foo-id"),
+			Payload: kong.String("example"),
+		},
+	}
+	assert.Nil(k.Add(lic1))
```
pmalek, line 230:
```suggestion
	require.NoError(t, k.Add(lic1))
```

state/license_test.go
```diff
@@ -0,0 +1,289 @@
...
+
+	lic1.Payload = kong.String("other")
+	assert.Nil(k.Update(lic1))
```
pmalek, line 233:
```suggestion
	require.NoError(t, k.Update(lic1))
```

pmalek (commented):
Apart from my previous comments that were mostly addressed by debt backlog issues and minor nits suggesting `require` I believe this should be good (to an extent I am able to reason about this code which is not all that much).

Deferring to @GGabriele as I believe he's got more to say in here :)

cmd/diff.go
```diff
@@ -65,6 +65,8 @@ that will be created, updated, or deleted.
 			"and exit code 1 if an error occurs.")
 	diffCmd.Flags().BoolVar(&dumpConfig.SkipCACerts, "skip-ca-certificates",
 		false, "do not diff CA certificates.")
+	diffCmd.Flags().BoolVar(&dumpConfig.SkipLicenses, "skip-licenses",
+		true, "do not diff licenses.")
```
GGabriele, line 69: I think `skip-licenses` and `"do not diff licenses."` is a bit confusing (one is "positive", the other is "negative").

Should we do something like this instead:

`diffCmd.Flags().BoolVar(&dumpConfig.SkipLicenses, "no-skip-licenses", false, "do not diff licenses.")`

It feels more natural adding `--no-skip-licenses` (or similar) instead of `--skip-licenses=false`.

WDYT?

cmd/dump.go
```diff
@@ -175,5 +175,7 @@ configure Kong.`,
 		false, "assume `yes` to prompts and run non-interactively.")
 	dumpCmd.Flags().BoolVar(&dumpConfig.SkipCACerts, "skip-ca-certificates",
 		false, "do not dump CA certificates.")
+	dumpCmd.Flags().BoolVar(&dumpConfig.SkipLicenses, "skip-licenses",
+		true, "do not diff licenses.")
```
GGabriele, line 179: After we come up with a decision about [this](https://github.com/Kong/deck/pull/882/files#r1171240128), can you also please fix the doc strings? These are all referring to `diff` regardless of the command type

cmd/diff.go
rainest: In general, consistent language across like features is good UX. If we're adding a new switch that is similar to existing ones, it should normally follow the established pattern. Were we to change this, we should do so across the board and change the like flags for CA certificates and consumers.

I don't share the same confusion over both needing to be positive, but in either case, we should not change existing flags on a whim. Doing so would break any existing user scripts that use the existing flags, so I'd only recommend changing them if we have a documented history of user reports indicating that they failed to find the correct flag when attempting to skip these resources.

Changing the docstring would be a non-breaking change, but ultimately it's out of scope here: this PR seeks to add licenses and handle them similarly to other resources that can be excluded from actions, not to revise existing UX patterns. Please create a follow-up issue if you wish to change this.

rainest:
Something appears to be broken inside state and I'm not seeing how (the table is indeed declared) based on upstream failures when integrating this: https://github.com/Kong/kubernetes-ingress-controller/actions/runs/4738023905/jobs/8411439709

Holding pending that.

Go figure, apparently I should have just taken the "only add the file type" shortcut.

state/license_test.go
pmalek: I understand your point and the suggestion. It just has never been said explicitly and there's always been the implicit assumption that the PR owner whenever there's mutual agreement about a PR comment, makes the changes himself to said PR.

I'm fine with both approaches as long as it's clearly communicated (as in the above comment) what's the expectation.

cmd/diff.go
GGabriele: The difference between this case and the existing ones is the default value and the resulting usage of these flags.

For CA certificates, the default behaviour is to _include_ them into the computation. If you want to skip them, then you add `--skip-ca-certificates`.

For Licenses, the default behaviour is to _exclude_ them. With the current logic, if you want to include them, then you add `--skip-licenses=false`.

I don't think we should change any existing UX otherwise we are going to break existing usage and introduce unnecessary friction, so no follow-up issue will be created, but I do think that we should provide a good UX for new features. IMHO, either `--include-licenses` or `--no-skip-licenses` are more in line with this goal.

mikefero:
> Testing the flag is hindered by our lack of a dedicated test license, however. There's no license I can provide that's accepted by the license endpoint but also not an actual license. I'm unsure if we have permission to stick a previously-valid test license in place and/or if it will cause other havoc if so (it would probably actually load the expired license, and lock out config for other tests).

Could we add an environmental variable in our tests which if available is loaded and used; otherwise the license tests could be skipped automatically for local and community executions? This would then allow us to add a secret which could be used for CI/CD to ensure all the things.

rainest:
> Could we add an environmental variable in our tests which if available is loaded and used; otherwise the license tests could be skipped automatically for local and community executions? This would then allow us to add a secret which could be used for CI/CD to ensure all the things.

Not so much that (we already have Enterprise tests) as the nature of the test. Most of the Enterprise tests are fine accepting a license from an environment secret because they don't care about the license itself so long as it's valid. This cares about the license specifically, and needs to provide it via a file instead of an environment variable.

We could maybe set up something where CI injects the secret value into a file, but we don't have anything like that at present.

mikefero:
> We could maybe set up something where CI injects the secret value into a file, but we don't have anything like that at present. The other state file tests just use static files committed to the repo.

We should do this then; doesn't have to be in this PR, but we should create a task for this so it doesn't get lost. If the environment variable is present we could have the test framework generate the file needed either in `${TMP}` and then utilize it. This would make it work similarly for local dev and also in CI/CD; WDYT?

rainest:
Alright, false alarm on the integration test errors: I had been using an older version of this PR in the KIC branch, and that older version had a bug that I'd fixed since. I'd forgotten to pull the current tip of this into the KIC branch; after doing so the integration tests are passing normally again.

rainest:
Closed in favor of #900

PR #890, opened 2023-04-11
feat: add proxy support for konnect

alex-tectu:
This pull request adds a new parameter called "proxy-addr". This parameter is designed to help users who are trying to access Kong Konnect behind their company's network firewall. By specifying a proxy address, users can bypass the firewall and access Kong Konnect.

Example:
	deck ping --proxy-addr http://proxy.example.com:8080


utils/types.go
```diff
@@ -295,6 +296,13 @@ func GetKonnectClient(httpClient *http.Client, config KonnectConfig) (*konnect.C
 		httpClient = http.DefaultClient
 		httpClient.Transport = defaultTransport
 	}
+	if config.ProxyAddress != "" {
+		proxyURL, err := url.Parse(config.ProxyAddress)
+		if err == nil {
```
GGabriele, line 301: Can you please return an error in case `err` is not nil?

Tieske:
@GGabriele instead of adding a custom option, shouldn't we just honor the OS env variables for proxies? just a thought

Tieske:
e.g. use this function instead; https://pkg.go.dev/net/http#ProxyFromEnvironment

Tieske:
replaced by #952

PR #904, opened 2023-05-03
fix: handle nil pointer dereference in `Defaulter.getEntitySchema`

tjasko:
Corrects a nil pointer dereference when `Defaulter.getEntitySchema()` is called & the HTTP request fails.

### DIFF
--- a/utils/defaulter.go
+++ b/utils/defaulter.go
@@ -215,12 +215,15 @@ func (d *Defaulter) getEntitySchema(entityType string) (map[string]interface{}, 
 		return schema, err
 	}
 	resp, err := d.client.Do(d.ctx, req, &schema)
+	if err != nil {
+		return schema, err
+	}
 	// in case the schema is not found - like in case of EE features,
 	// no error should be returned.
 	if resp.StatusCode == http.StatusNotFound {
 		return schema, nil
 	}
-	return schema, err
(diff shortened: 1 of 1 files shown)


aboudreault (changes requested):
looking into the test failure..

GGabriele:
Due to the nature of the API, I think we should actually do something like this: https://github.com/Kong/deck/pull/869

utils/defaulter.go
```diff
@@ -215,12 +215,15 @@ func (d *Defaulter) getEntitySchema(entityType string) (map[string]interface{},
 		return schema, err
 	}
 	resp, err := d.client.Do(d.ctx, req, &schema)
+	if err != nil {
```
aboudreault, line 218: I think we want to preserve the desired behavior described here: https://github.com/Kong/deck/pull/904/files#diff-d85c74f21d13d17a0932f304636914d8aacd18acb50f86632aee600dd1080b20R221

looks like we are receiving a `*kong.APIError` with  the 404 code, so we could check this instead of checking the resp.StatusCode.

tjasko: @GGabriele already has PR up for this that I missed (#869). The `Client.Do()` method is hooked up in a bit of an abnormal way, so that threw me for a loop.

PR #909, opened 2023-05-15
Remove deprecated `konnect` subcommand

voigtjr:
The deprecated `konnect` subcommands rely on the deactivated `konnect.konghq.com` domain and is now broken because of that. These changes remove the deprecated subcommands.


voigtjr:
I'll reopen inside of the kong org.

PR #910, opened 2023-05-15
Remove deprecated `konnect` subcommands

voigtjr:
The deprecated konnect subcommands rely on the deactivated konnect.konghq.com domain and is now broken because of that. These changes remove the deprecated subcommands.


Tieske:
should this be targeted at the `v2` branch? see https://github.com/Kong/deck/pull/941

PR #918, opened 2023-05-22
feat: allow updating entities' IDs while keeping their names

czeslavo:
This adds an additional step to the `diff.Syncer`'s `diff` producer - `deleteDuplicates`. Its purpose is to clean up the entities in the **current** state that have duplicates with the same names in the **target** state before proceeding with updates/creates and regular deletes. Without that additional step, it's not possible to perform sync in which the IDs of entities are changed while their names are kept between syncs. 

An actual use-case that is the motivation for this change is Kong Ingress Controller where we recently started assigning entities' IDs explicitly. Without this change, existing setups break when KIC is upgraded from the version not setting the IDs to the new one because of breaking the uniqueness constraints of Services, Routes, and Consumers.

The added test cases represent the complete use case with all the affected entities: Services, Routes, and Consumers. We change their IDs, keeping the names and we expect the final state on the Kong side to be using the new IDs. 

The problematic part of the implementation is how Services are handled - we cannot safely delete them as they might be referred by Routes. When they do, deletes fail as there's no cascade delete for them configured and foreign key constraint is violated. Because of that, in the Service's `deleteDuplicateService` implementation, we have to manually generate delete events for the associated Routes and, in consequence, their Plugins. It is to make sure they'll be recreated in the later stage of the diff command - `createUpdate`.

That gives us more headache as that means we generate events that shouldn't be handled in parallel in the case of Service. Because of that, higher level events rearranging needs to happen to make sure they are all handled in the expected `reverseOrder`, just as the regular `delete` events.

It's a prerequisite for solving https://github.com/Kong/kubernetes-ingress-controller/issues/4025.

### DIFF
--- a/types/postProcess.go
+++ b/types/postProcess.go
@@ -2,6 +2,7 @@ package types
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/kong/deck/crud"
 	"github.com/kong/deck/state"
@@ -16,7 +17,20 @@ func (crud *servicePostAction) Create(_ context.Context, args ...crud.Arg) (crud
 }
 
 func (crud *servicePostAction) Delete(_ context.Context, args ...crud.Arg) (crud.Arg, error) {
-	return nil, crud.currentState.Services.Delete(*((args[0].(*state.Service)).ID))
+	serviceID := *args[0].(*state.Service).ID
+
+	// Delete all plugins associated with this service as that's the implicit behavior of Kong (cascade delete).
+	plugins, err := crud.currentState.Plugins.GetAllByServiceID(serviceID)
+	if err != nil {
+		return nil, fmt.Errorf("error looking up plugins for service '%v': %v", serviceID, err)
+	}
+	for _, plugin := range plugins {
+		err = crud.currentState.Plugins.Delete(*plugin.ID)
+		if err != nil {
+			return nil, fmt.Errorf("error deleting plugin '%v' for service '%v': %v", *plugin.ID, serviceID, err)
+		}
+	}
+	return nil, crud.currentState.Services.Delete(serviceID)
 }
 
 func (crud *servicePostAction) Update(_ context.Context, args ...crud.Arg) (crud.Arg, error) {
@@ -32,7 +46,20 @@ func (crud *routePostAction) Create(_ context.Context, args ...crud.Arg) (crud.A
 }
 
 func (crud *routePostAction) Delete(_ context.Context, args ...crud.Arg) (crud.Arg, error) {
-	return nil, crud.currentState.Routes.Delete(*((args[0].(*state.Route)).ID))
+	routeID := *args[0].(*state.Route).ID
+
+	// Delete all plugins associated with this route as that's the implicit behavior of Kong (cascade delete).
+	plugins, err := crud.currentState.Plugins.GetAllByRouteID(routeID)
+	if err != nil {
+		return nil, fmt.Errorf("error looking up plugins for route '%v': %v", routeID, err)
+	}
+	for _, plugin := range plugins {
+		err = crud.currentState.Plugins.Delete(*plugin.ID)
+		if err != nil {
+			return nil, fmt.Errorf("error deleting plugin '%v' for route '%v': %v", *plugin.ID, routeID, err)
+		}
+	}
+	return nil, crud.currentState.Routes.Delete(routeID)
 }
 
 func (crud *routePostAction) Update(_ context.Context, args ...crud.Arg) (crud.Arg, error) {
@@ -146,7 +173,19 @@ func (crud *consumerPostAction) Create(_ context.Context, args ...crud.Arg) (cru
 }
 
 func (crud *consumerPostAction) Delete(_ context.Context, args ...crud.Arg) (crud.Arg, error) {
-	return nil, crud.currentState.Consumers.Delete(*((args[0].(*state.Consumer)).ID))
+	consumerID := *args[0].(*state.Consumer).ID
+
+	// Delete all plugins associated with this consumer as that's the implicit behavior of Kong (cascade delete).
+	plugins, err := crud.currentState.Plugins.GetAllByConsumerID(consumerID)
+	if err != nil {
+		return nil, fmt.Errorf("error looking up plugins for consumer '%v': %v", consumerID, err)
+	}
+	for _, plugin := range plugins {
+		if err := crud.currentState.Plugins.Delete(*plugin.ID); err != nil {
+			return nil, fmt.Errorf("error deleting plugin '%v' for consumer '%v': %v", *plugin.ID, consumerID, err)
+		}
+	}
+	return nil, crud.currentState.Consumers.Delete(consumerID)
 }
 
 func (crud *consumerPostAction) Update(_ context.Context, args ...crud.Arg) (crud.Arg, error) {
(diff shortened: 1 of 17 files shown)


czeslavo:
It's ready for initial review - I'd like to validate whether the path I'm following here is sane and if there's no simpler way to achieve the possibility of changing entities' IDs without changing their names. 

I'd much appreciate input from @hbagdi who authored most of the code I was touching here. I was doing my best to not break existing scenarios after reading your comment on ec56add5 commit saying:

> Another change is create and update operations are now performed before
delete operations.
>This is to cover the following case:
>- assume service s1 with route r1 is present in kong
>- an update needs to perform: delete s1, create s2, associate r1 to s2
>
>This operation will fail if delete is first performed since s1 can't be
deleted as it has r1 associated with it.
>
>But if s2 is created first, then, r1 is moved from s1 to s2. then s1 can
be deleted safely.

czeslavo:
@aboudreault Cool! It's great you were able to catch it in manual tests. Indeed I made this change unnecessarily: 
```diff
+       updatedService, err := s.client.Services.Update(ctx, &service.Service)
-       updatedService, err := s.client.Services.Create(ctx, &service.Service)
```
I added a test case that covers updates of the entities with explicitly assigned IDs and fixed the regression.

aboudreault:
thanks @czeslavo! I will test this again on Monday.

types/postProcess.go
```diff
@@ -16,7 +17,20 @@ func (crud *servicePostAction) Create(_ context.Context, args ...crud.Arg) (crud
 }
 
 func (crud *servicePostAction) Delete(_ context.Context, args ...crud.Arg) (crud.Arg, error) {
-	return nil, crud.currentState.Services.Delete(*((args[0].(*state.Service)).ID))
+	serviceID := *args[0].(*state.Service).ID
+
+	// Delete all plugins associated with this service as that's the implicit behavior of Kong (cascade delete).
+	plugins, err := crud.currentState.Plugins.GetAllByServiceID(serviceID)
+	if err != nil {
+		return nil, fmt.Errorf("error looking up plugins for service '%v': %v", serviceID, err)
```
aboudreault, line 25:
```suggestion
		return nil, fmt.Errorf("error looking up plugins for service '%v': %w", serviceID, err)
```

types/postProcess.go
```diff
@@ -16,7 +17,20 @@ func (crud *servicePostAction) Create(_ context.Context, args ...crud.Arg) (crud
...
+	}
+	for _, plugin := range plugins {
+		err = crud.currentState.Plugins.Delete(*plugin.ID)
+		if err != nil {
+			return nil, fmt.Errorf("error deleting plugin '%v' for service '%v': %v", *plugin.ID, serviceID, err)
```
aboudreault, line 30:
```suggestion
			return nil, fmt.Errorf("error deleting plugin '%v' for service '%v': %w", *plugin.ID, serviceID, err)
```

types/postProcess.go
```diff
@@ -32,7 +46,20 @@ func (crud *routePostAction) Create(_ context.Context, args ...crud.Arg) (crud.A
...
+
+	// Delete all plugins associated with this route as that's the implicit behavior of Kong (cascade delete).
+	plugins, err := crud.currentState.Plugins.GetAllByRouteID(routeID)
+	if err != nil {
+		return nil, fmt.Errorf("error looking up plugins for route '%v': %v", routeID, err)
+	}
+	for _, plugin := range plugins {
+		err = crud.currentState.Plugins.Delete(*plugin.ID)
+		if err != nil {
+			return nil, fmt.Errorf("error deleting plugin '%v' for route '%v': %v", *plugin.ID, routeID, err)
```
aboudreault, line 59:
```suggestion
			return nil, fmt.Errorf("error deleting plugin '%v' for route '%v': %w", *plugin.ID, routeID, err)
```

types/postProcess.go
```diff
@@ -32,7 +46,20 @@ func (crud *routePostAction) Create(_ context.Context, args ...crud.Arg) (crud.A
 }
 
 func (crud *routePostAction) Delete(_ context.Context, args ...crud.Arg) (crud.Arg, error) {
-	return nil, crud.currentState.Routes.Delete(*((args[0].(*state.Route)).ID))
+	routeID := *args[0].(*state.Route).ID
+
+	// Delete all plugins associated with this route as that's the implicit behavior of Kong (cascade delete).
+	plugins, err := crud.currentState.Plugins.GetAllByRouteID(routeID)
+	if err != nil {
+		return nil, fmt.Errorf("error looking up plugins for route '%v': %v", routeID, err)
```
aboudreault, line 54:
```suggestion
		return nil, fmt.Errorf("error looking up plugins for route '%v': %w", routeID, err)
```

types/postProcess.go
```diff
@@ -146,7 +173,19 @@ func (crud *consumerPostAction) Create(_ context.Context, args ...crud.Arg) (cru
 }
 
 func (crud *consumerPostAction) Delete(_ context.Context, args ...crud.Arg) (crud.Arg, error) {
-	return nil, crud.currentState.Consumers.Delete(*((args[0].(*state.Consumer)).ID))
+	consumerID := *args[0].(*state.Consumer).ID
+
+	// Delete all plugins associated with this consumer as that's the implicit behavior of Kong (cascade delete).
+	plugins, err := crud.currentState.Plugins.GetAllByConsumerID(consumerID)
+	if err != nil {
+		return nil, fmt.Errorf("error looking up plugins for consumer '%v': %v", consumerID, err)
```
aboudreault, line 181:
```suggestion
		return nil, fmt.Errorf("error looking up plugins for consumer '%v': %w", consumerID, err)
```

types/postProcess.go
```diff
@@ -146,7 +173,19 @@ func (crud *consumerPostAction) Create(_ context.Context, args ...crud.Arg) (cru
...
+	}
+	for _, plugin := range plugins {
+		if err := crud.currentState.Plugins.Delete(*plugin.ID); err != nil {
+			return nil, fmt.Errorf("error deleting plugin '%v' for consumer '%v': %v", *plugin.ID, consumerID, err)
```
aboudreault, line 185:
```suggestion
			return nil, fmt.Errorf("error deleting plugin '%v' for consumer '%v': %w", *plugin.ID, consumerID, err)
```

aboudreault (changes requested):
Nice job here @czeslavo! I spend some time to review this change, did some additional testing and overall it seems to work well now.  One thing I noticed is that this "bug" also exists with all other entities that have more than the ID as unique constraint (e.g. Upstream). We should create a follow-up ticket for these ones.

types/postProcess.go
mmorel-35: Hi !
Adding [errorlint](https://golangci-lint.run/usage/linters/#errorlint) linter to golangci-lint could help you ensure that .

czeslavo:
@aboudreault I addressed all your suggestions and added a changelog entry, also created a tracking issue to cover the other entities (https://github.com/Kong/deck/issues/922).

aboudreault (approved):

types/postProcess.go
aboudreault: +1, will look to add this one

aboudreault: ref: https://github.com/Kong/deck/pull/923

PR #927, opened 2023-05-31
chore: downgrade goreleaser-action to v3

GGabriele:


.github/workflows/release.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
         with:
           go-version: 1.20
       - name: Run GoReleaser
-        uses: goreleaser/goreleaser-action@v4
+        uses: goreleaser/goreleaser-action@v3
```
czeslavo, line 21: Instead of this (which is just the action version) I think we should change the `with.version` from `latest` to the last version that worked for us (I believe it was v1.16.2 based on https://github.com/Kong/deck/actions/runs/4482371787/jobs/7880326738).

GGabriele: Let's try that. Updated

aboudreault (approved):

aboudreault (approved):

Issue #616, opened 2022-03-16
Multi-Line Environment Variables Validation Errors


probably-not:
In most companies, security policies dictate that we don't store secrets within a simple Git repository, and instead, inject the data via env variables to any CI/CD process that runs and requires the secrets.

While decK allows env variable injection, the injection is done in a simple way using Go Templates, without any regards to formatting of the file format. Subsequently, if the secret is a multi-line secret (such as a private or public RSA key necessary in certain plugins), then this breaks the file formatting and will cause errors in the file validation.

As an example, with a state file that looks like this:

```yaml
_format_version: "1.1"
consumers:
  - custom_id: consumer_1
    jwt_secrets:
      - algorithm: RS256
        key: consumer_1_iss_key
        rsa_public_key: ${{ env "DECK_RSA_PUBLIC_KEY" }}
        secret: consumer_1_secret
    username: consumer_1_user_name
```

Running `decK validate` fails with the following error:
```sh
$ DECK_RSA_PUBLIC_KEY=$(cat pubkey.pem) deck validate

Error: reading file: validating file content: unmarshaling file content: error converting YAML to JSON: yaml: line 9: could not find expected ':'
exit status 1
```

hbagdi:
I agree that the injection is a bit simplistic and we should be doing something more.
Very open to new ideas on making this better.

For the specific case, something like this _may_ get work:
```yaml
_format_version: "1.1"
consumers:
  - custom_id: consumer_1
    jwt_secrets:
      - algorithm: RS256
        key: consumer_1_iss_key
        rsa_public_key: |
          ${{ env "DECK_RSA_PUBLIC_KEY" }}
        secret: consumer_1_secret
    username: consumer_1_user_name

```

probably-not:
> For the specific case, something like this may get work:
Unfortunately that still fails with the same error, since the template is still injecting multiple lines into the file and the subsequent lines aren't indented.

> Very open to new ideas on making this better.
@hbagdi What do you think about first unmarshaling the YAML file into a temporary map immediately after reading in the `readContent` function, and then iterating over the keys and making any replacements you need while iterating? Once you complete the replacements you could then use something like mapstructure.Decode to decode from a map to a `Content` struct.

probably-not:
Either that, or the template funcs that are passed in need to be a lot more powerful so that they can enable the yaml (and JSON) formatting necessary

hbagdi:
[...]

PR #929, opened 2023-06-02
feat: Add indent function to support multi-line content

aparra:
We want to extract Lua code from `kong.yaml` file for two reasons:
1. Better syntax highlight in the IDE for .lua files
2. Write unit tests for the functions

So, we would like to have a setup like that:

**kong.yaml**
```yaml
---
_format_version: "3.0"
plugins:
  - name: pre-function
    config:
      log: |
        ${{ env "DECK_FILE_LOG_FUNCTION" | indent 8 }}
```

**file_log.lua**
```lua
function parse_traceid(str)str = string.sub(str,1,8)
  local uint = 0
  for i = 1, #str do
    uint = uint + str:byte(i) * 0x100^(i-1)
  end
  return string.format("%.0f", uint)
end

kong.log.set_serialize_value("trace_id", parse_traceid(ngx.ctx.KONG_SPANS[1].trace_id))
kong.log.set_serialize_value("span_id", parse_traceid(ngx.ctx.KONG_SPANS[1].span_id))
```

However, without the `indent` function, it is not possible to convert the yaml to json. It fails due indentation required by yaml. This issue https://github.com/Kong/deck/issues/616 worked around this problem, replacing yaml with json, where you have better control of quotes.

But we would like consistency in the setup; it means using yaml everywhere. Inspired by `k8s helm`, where function indent is available https://helm.sh/docs/chart_template_guide/function_list/#indent to deal with multi-line block, I added the indent function to deck.

### DIFF
--- a/file/readfile.go
+++ b/file/readfile.go
@@ -158,12 +158,18 @@ func toFloat(key string) (float64, error) {
 	return strconv.ParseFloat(key, 64)
 }
 
+func indent(spaces int, v string) string {
+	pad := strings.Repeat(" ", spaces)
+	return strings.Replace(v, "\n", "\n"+pad, -1)
+}
+
 func renderTemplate(content string) (string, error) {
 	t := template.New("state").Funcs(template.FuncMap{
 		"env":     getPrefixedEnvVar,
 		"toBool":  toBool,
 		"toInt":   toInt,
 		"toFloat": toFloat,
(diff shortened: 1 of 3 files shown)


aboudreault:
@aparra Thanks for this PR. tested quickly and it seems to work well. I will check to merge this one soon

PR #945, opened 2023-06-14
fix: handle missing service and route names detecting duplicates

### LINKED ISSUE #944: Nil pointer dereference for deleteDuplicateRoute
Hi

When running several differents commands like diff or sync, I'm getting the following nil pointer derefence. I'm getting this when creating routes for any of my services.
I also tested creating the route with other means and then dump with deck and sync, and I would get the same error.

This is an example for a service giving this error:

 ```
 - connect_timeout: 60000
    enabled: true
    host: myhost.svc.cluster.local
    name: myname
    port: 8000
    protocol: http
    read_timeout: 60000
    retries: 5
    routes:
      - https_redirect_status_code: 426
        id: 49d0a8fd-cb3b-4e73-b057-f1be9152f05d
        path_handling: v0
        paths:
          - /my-path
        preserve_host: false
        protocols:
          - http
          - https
        regex_priority: 0
        request_buffering: true
        response_buffering: true
        strip_path: true
    write_timeout: 60000
```

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x13ef82b]

goroutine 391 [running]:
github.com/kong/deck/types.(*routeDiffer).deleteDuplicateRoute(0x20?, 0xc0000f5400)
	github.com/kong/deck/types/route.go:191 +0x2b
github.com/kong/deck/types.(*routeDiffer).DuplicatesDeletes(0xc000b55580)
	github.com/kong/deck/types/route.go:178 +0x105
github.com/kong/deck/diff.(*Syncer).deleteDuplicates(0xc000b18980)
	github.com/kong/deck/diff/diff.go:190 +0x50e
github.com/kong/deck/diff.(*Syncer).diff(0x103df65?)
	github.com/kong/deck/diff/diff.go:174 +0xa6
github.com/kong/deck/diff.(*Syncer).Run.func2()
	github.com/kong/deck/diff/diff.go:322 +0x2b
created by github.com/kong/deck/diff.(*Syncer).Run
	github.com/kong/deck/diff/diff.go:321 +0x31f
```

Versions:
* Kong: 3.3.0
* Deck: decK v1.22.0 (7447a09)

czeslavo:
As both `Service` and `Route` names are allowed to be empty, we have to account for this case in the code detecting duplicates and short-circuit when the target entity has no name to prevent panics.

Fixes #944.

### DIFF
--- a/tests/integration/sync_test.go
+++ b/tests/integration/sync_test.go
@@ -3414,3 +3414,47 @@ func Test_Sync_UpdateWithExplicitIDs(t *testing.T) {
 		},
 	}, ignoreFieldsIrrelevantForIDsTests)
 }
+
+// test scope:
+//   - 3.0.0+
+//   - konnect
+func Test_Sync_UpdateWithExplicitIDsWithNoNames(t *testing.T) {
+	runWhenKongOrKonnect(t, ">=3.0.0")
+
+	client, err := getTestClient()
+	if err != nil {
+		t.Errorf(err.Error())
+	}
+
+	const (
+		beforeConfig = "testdata/sync/022-update-with-explicit-ids-with-no-names/before.yaml"
+		afterConfig  = "testdata/sync/022-update-with-explicit-ids-with-no-names/after.yaml"
+	)
+
+	// First, create entities with IDs assigned explicitly.
+	err = sync(beforeConfig)
+	require.NoError(t, err)
+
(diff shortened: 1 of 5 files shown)


Koumbaya (approved):
Nice

aboudreault (approved):

PR #963, opened 2023-07-14
feat(render): add new command to render final file

Tieske:
recovers #874 including tests


GGabriele (commented):
Does this PR still need all the changes to the `convert` package now that an entire new command (`render`) has been added?

Tieske:
> Does this PR still need all the changes to the `convert` package now that an entire new command (`render`) has been added?

I removed the duplicate code from `cmd/file_render.go`, and left the changes in `convert/convert.go`. Reasoning; the convert package has tests, where the cmd package has not. So this seemed the better option.

cmd/file_render.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+var (
+	fileRenderCmdKongStateFile  []string
+	fileRenderCmdKongFileOutput string
+	fileRenderCmdStateFormat    string
+)
+
+func executeFileRenderCmd(_ *cobra.Command, _ []string) error {
+	destinationFormat, err := convert.ParseFormat(fileRenderCmdStateFormat)
```
GGabriele, line 17:
```suggestion
	destinationFormat, err := convert.ParseFormat(string(convert.FormatKongGateway3x))
```

cmd/file_render.go
```diff
@@ -0,0 +1,53 @@
...
+
+func newFileRenderCmd() *cobra.Command {
+	renderCmd := &cobra.Command{
+		Use:   "render",
+		Short: "Render the configuration as Kong declarative config",
+		Long:  ``,
+		Args:  validateNoArgs,
+		RunE:  executeFileRenderCmd,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if len(fileRenderCmdStateFormat) == 0 {
```
GGabriele, line 34:
```suggestion
			if len(fileRenderCmdKongStateFile) == 0 {
```

cmd/file_render.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	renderCmd.Flags().StringSliceVarP(&fileRenderCmdKongStateFile,
+		"state", "s", []string{"-"}, "file(s) containing Kong's configuration.\n"+
+			"This flag can be specified multiple times for multiple files.\n"+
+			"Use `-` to read from stdin.")
+	renderCmd.Flags().StringVarP(&fileRenderCmdKongFileOutput, "output-file", "o",
+		"-", "file to which to write Kong's configuration."+
+			"Use `-` to write to stdout.")
+	renderCmd.Flags().StringVar(&fileRenderCmdStateFormat, "format",
```
GGabriele, line 49: I think this is not working as expected. Right now `convert` only supports [yaml](https://github.com/Kong/deck/blob/5ca2755b131ccc4d580ce240cb0484339bda62b0/convert/convert.go#L54) as file format. Either we introduce the support for `json` in this PR or we remove this option entirely (can't remember if that was here before the "force-push" of the old PR)

Tieske:
@GGabriele @zekth rebased and squashed. Should be ready to go now.

cmd/file_convert.go
```diff
@@ -37,7 +40,13 @@ func executeConvert(_ *cobra.Command, _ []string) error {
 			return nil
 		}
 
-		err = convert.Convert(convertCmdInputFile, convertCmdOutputFile, sourceFormat, destinationFormat)
+		err = convert.Convert(
+			[]string{convertCmdInputFile},
+			convertCmdOutputFile,
+			file.Format(strings.ToUpper(convertCmdStateFormat)),
```
zekth, line 46: what happens if the user uses `yml` instead of `yaml` does it panic? shouldn't we validate this input?

Tieske: That value is simply passed along all the way to writing the final file, where it gets validated;  https://github.com/Kong/deck/blob/main/file/writer.go#L737-L749

zekth: but `file.Format(...)` would pass?

zekth (dismissed):

GGabriele (approved):

PR #962, opened 2023-07-14
chore(cli): create 'gateway' subcommand for network commands

Tieske:
This introduces the `gateway` sub-command, and moves the existing commands under the new sub-command. Since aliases are in place this is non-breaking.

The old commands will now display a deprecation warning.

New structure:
```
- version     Print the decK version
- completion  Generate completion script
- help        Help about any command
​
- file        decK file operations
  - add-tags     Adds tags to objects in a decK file
  - list-tags    Lists current tags to objects in a decK file
  - merge        Merges multiple decK files into one
  - openapi2kong Convert OpenAPI files to Kong's decK format
  - patch        Applies patches on top of a decK file
  - remove-tags  Removes tags from objects in a decK file
  - convert      Convert files from one format into another format
  - render       renders a final decK file as it would be synced (for local diffs)
  - validate     Validate the state file (locally)
- gateway     online operations
  - diff         Diff the current entities in Kong with the one on disks
  - dump         Export Kong configuration to a file
  - ping         Verify connectivity with Kong
  - reset        Reset deletes all entities in Kong
  - sync         Sync performs operations to get Kong's configuration to match the state file
  - validate     Validate the state file (online)
- diff        ==> alias for backward compatibilty
- dump        ==> alias for backward compatibilty
- ping        ==> alias for backward compatibilty
- reset       ==> alias for backward compatibilty
- sync        ==> alias for backward compatibilty
- validate    ==> alias for backward compatibilty
- convert     ==> alias for backward compatibilty
```

Todo:

- [x] create "kong" subcommand
- [x] move kong commands under "kong"
- [x] change defaults of "kong xxx" commands to focus on stdin/stdout
- [x] change "kong" subcommand to "gateway", see [internal discussion](https://kongstrong.slack.com/archives/C037510LW2X/p1689629823348549?thread_ts=1689595986.005799&cid=C037510LW2X)
- [x] add `file validate` which was removed by #971 

Jira: APIOPS-98

PR #965, opened 2023-07-17
feat(cli): add short-hand '-o' flag to convert

Tieske:
also fixes the file-convert command to use positional args like the other "file" based commands. But since that hasn't been released yet, it is not a "user-facing" fix.


Tieske:
also moved to milestone 1.25 (following the base branch)

Tieske:
moved into #962

PR #966, opened 2023-07-17
fix(cli): oas2kong would not parse the --format flag

Tieske:


rspurgeon (approved):
LGTM, manually tested with a local build

GGabriele (approved):

PR #967, opened 2023-07-17
Add CLI tests

Tieske:
Here's an initial attempt at adding CLI tests of the actual binary build.

This uses the same bash-based test framework that's also used by Pongo and the Kong Docker images. Not sure this is suitable, hence providing a minimal test set to discuss, before adding more.

### DIFF
--- a/tests/cli/mytestfile.test.sh
+++ b/tests/cli/mytestfile.test.sh
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+
+
+function run_test {
+  # the suite name below will only be used when running this file directly, when
+  # running through "test.sh" it must be provided using the "--suite" option.
+  tinitialize "decK-cli test suite" "${BASH_SOURCE[0]}"
+
+  tchapter "deck file openapi2kong"
+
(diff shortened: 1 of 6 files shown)


Tieske:
The actual test file: https://github.com/Kong/deck/blob/tests/cli/tests/cli/mytestfile.test.sh
The test output: https://github.com/Kong/deck/actions/runs/5580371892/jobs/10197178252?pr=967#step:9:7

tests/cli/mytestfile.test.sh
```diff
@@ -0,0 +1,42 @@
...
+
+  tchapter "deck file openapi2kong"
+
+  ttest "accepts format 'json'" 
+  ./deck file openapi2kong --format json > /dev/null < tests/cli/fixtures/mock-a-rena-oas.yml
+  if [ $? -ne 0 ]; then
+    tfailure "--format=json was not accepted"
+  else
+    tsuccess
+  fi
```
mheap, line 17: The repeated checks would be error prone. Should we consider using something like https://github.com/bats-core/bats-core that has auto-detection?

PR #972, opened 2023-07-18
fix(cli): fix indentation in file-patch cli command

Tieske:


Tieske:
@lena-larionova please have a look. All whitespace indentation was changed to tabs, which caused it to look really bad in the CLI (defaults to 8 spaces). Hence here I changed it to 1 tab, and everything beyond that to 2 space-indents.

Will this render properly in the docs? If not, how can we solve this such that it works for CLI and docs?

cmd/file_addplugins.go
```diff
@@ -121,20 +121,20 @@ The plugin files have the following format (JSON or YAML) and are applied in the
 order they are given:
 
 	{ "_format_version": "1.0",
```
lena-larionova, line 123: Forgot to ask earlier: any reason that this is format version 1.0? If no specific reason, then can we put 3.0, to match latest?

lena-larionova (approved):
lgtm, tested it on docs and it generates fine 👍

PR #981, opened 2023-07-24
feat(linting): Add support for linting decK files

mheap:
Adds support for linting declarative configurations.

Example usage:

```bash
deck file lint -s ./openapi.yaml -r ruleset.yaml
```

You can control what's considered an error:

```bash
deck file lint -s ./openapi.yaml -r ruleset.yaml --fail-severity info
```

Or only show errors:

```bash
deck file lint -s ./openapi.yaml -r ruleset.yaml -D
```

It supports JSON and YAML output in addition to a human readable default:

```
Linting Violations: 2
Failures: 1

[9:15] Must use HTTPS protocol: `http` does not match the expression `^https`
[1:18] Check the version is correct: `3.0` does not match the expression `^1.1$`
exit status 1
````

Tested with the following input files:

```yaml
# openapi.yaml
_format_version: "3.0"
services:
  - host: mockbin.org
    id: b1525aee-d304-11ed-afa1-0242ac120002
    name: summer-time
    path: /requests
    plugins: []
    port: 443
    protocol: http
    routes:
      - id: b7d87736-d304-11ed-afa1-0242ac120002
        methods:
          - GET
        name: summer-time_get
        paths:
          - ~/summer-time$
        plugins: []
        regex_priority: 200
        strip_path: false
        tags: []
```

```yaml
# ruleset.yaml
rules:
  version-30:
    description: "Check the version is correct"
    given: $._format_version
    severity: error
    then:
      function: pattern
      functionOptions:
        match: "^1.1$"
  https-only:
    description: "Must use HTTPS protocol"
    given: $.services[*].protocol
    severity: warn
    then:
      function: pattern
      functionOptions:
        match: "^https"
```

Usage:

```bash
Usage:
  deck file lint [flags] lint-file

Flags:
  -D, --display-only-failures   only display failures
      --fail-severity string    minimum severity to fail on (default "warn")
      --format string           output format: default, json or yaml (default "default")
  -h, --help                    help for lint
  -r, --ruleset string          Ruleset to apply to the state file.
  -s, --state string            decK file to process. Use - to read from stdin. (default "-")
```


cmd/file_lint.go
```diff
@@ -0,0 +1,165 @@
...
+	"github.com/spf13/cobra"
+)
+
+var (
+	cmdLintInputFilename string
+	cmdLintInputRuleset  string
+)
+
+func executeLint(cmd *cobra.Command, _ []string) error {
+	severityList := map[string]int{
```
GGabriele, line 23: Can we maybe use `iota` instead? Something like this:

```
type Severity int

const (
	SeverityHint Severity = iota
	SeverityInfo
	SeverityWarn
	SeverityError
)

var severityStrings = [...]string{
	"hint",
	"info",
	"warn",
	"error",
}

func ParseSeverity(s string) Severity {
	for i, str := range severityStrings {
		if s == str {
			return Severity(i)
		}
	}
	return SeverityWarn
}
```

cmd/file_lint.go
```diff
@@ -0,0 +1,165 @@
...
+		"hint":  0,
+		"info":  1,
+		"warn":  2,
+		"error": 3,
+	}
+
+	verbosity, _ := cmd.Flags().GetInt("verbose")
+	logbasics.Initialize(log.LstdFlags, verbosity)
+
+	inputFilename, err := cmd.Flags().GetString("state")
```
GGabriele, line 33: As mentioned above, the file name is already available in `cmdLintInputFilename` (same for the other values below)

cmd/file_lint.go
```diff
@@ -0,0 +1,165 @@
...
+		if err != nil {
+			return fmt.Errorf("failed getting cli argument 'format'; %w", err)
+		}
+		outputFormat = strings.ToUpper(outputFormat)
+	}
+
+	inputRuleset, err := cmd.Flags().GetString("ruleset")
+
+	if err != nil || inputRuleset == "" {
+		return fmt.Errorf("failed getting cli argument 'ruleset';")
```
GGabriele, line 60: Wouldn't it be better to differentiate the two cases? e.g. signal `--ruleset` as required if missing and fail with a specific error in case of error while parsing the value.

Note,  all these `cmd.Flags().Get*` can be replaced by the appropriate `StringVarP` / `BoolVarP` etc which automatically parse the values.

cmd/file_lint.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	ruleSetBytes, err := ioutil.ReadFile(inputRuleset)
+	if err != nil {
+		panic(err.Error())
```
GGabriele, line 65: Let's avoid panic'ing (here and elsewhere too) and wrap the error so that's easier to understand what's happening (at what point the failure happened) for the user.

cmd/file_lint.go
```diff
@@ -0,0 +1,165 @@
...
+		})
+	}
+
+	lintErrs := map[string]interface{}{
+		"total_count": totalCount,
+		"fail_count":  failingCount,
+		"results":     lintResults,
+	}
+
+	if outputFormat == "DEFAULT" {
```
GGabriele, line 116: Let's make `DEFAULT` a constant

cmd/file_lint.go
```diff
@@ -0,0 +1,165 @@
...
+			}
+		}
+	} else {
+		err = filebasics.WriteSerializedFile("-", lintErrs, filebasics.OutputFormat(outputFormat))
+		if err != nil {
+			return err
+		}
+	}
+
+	if failingCount > 0 {
```
GGabriele, line 131: You can simply do:

```
	if failingCount > 0 {
		os.Exit(1)
	}
```

cmd/file_lint.go
```diff
@@ -0,0 +1,165 @@
...
+		Short: "Short description here",
+		Long:  `Long description Here`,
+		RunE:  executeLint,
+	}
+
+	lintCmd.Flags().StringVarP(&cmdLintInputFilename, "state", "s", "-",
+		"decK file to process. Use - to read from stdin.")
+	lintCmd.Flags().StringVarP(&cmdLintInputRuleset, "ruleset", "r", "",
+		"Ruleset to apply to the state file.")
+	lintCmd.Flags().StringP("format", "", "default", "output format: default, "+
```
GGabriele, line 159: Providing a `--format` different than `json` or `yaml` returns the following error:

```
$ ./deck file lint -s kong.yaml --ruleset ruleset.yaml --format foooo
Error: expected 'format' to be either 'yaml' or 'json', got: 'foooo'
```

The error suggests these two are the only allowed formats, while `default` is also supported.

GGabriele (changes requested):
Nice work!

I added a few suggestions and requests in the code. On top of those, can you please make sure to rebase on top of `main` and also add some basic test? Something [like this](https://github.com/Kong/deck/blob/main/tests/integration/dump_test.go) should be relatively easy to replicate.

cmd/file_lint.go
```diff
@@ -0,0 +1,165 @@
...
+
+func newLintCmd() *cobra.Command {
+	lintCmd := &cobra.Command{
+		Use:   "lint [flags] lint-file",
+		Short: "Short description here",
+		Long:  `Long description Here`,
+		RunE:  executeLint,
+	}
+
+	lintCmd.Flags().StringVarP(&cmdLintInputFilename, "state", "s", "-",
```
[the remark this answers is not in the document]
rspurgeon: I've only briefly considered this, but my gut says that linting over a single file should be sufficient.  In a pipeline, it seems we might generate->transform->lint and maybe have multiple of those that then merge->lint->sync or something like this.  If we support linting across multiple files, the output will need to clearly show which input file contains the warn/error.

My vote is single file until we have feedback that multiple is a necessary feature.

For positional arguments in the current `deck file ...` commands, it depends on which command.  Merge takes all files via positional args, but patch accepts the input on `-s, --state`.   The current usage `deck file lint` says:
```
deck file lint [flags] lint-file
``` 
but I'm not certain what `lint-file` is at this point.

cmd/file_lint.go
```diff
@@ -0,0 +1,165 @@
...
+
+	"github.com/daveshanley/vacuum/motor"
+	"github.com/daveshanley/vacuum/rulesets"
+	"github.com/kong/go-apiops/filebasics"
+	"github.com/kong/go-apiops/logbasics"
+	"github.com/spf13/cobra"
+)
+
+var (
+	cmdLintInputFilename string
```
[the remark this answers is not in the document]
rspurgeon: It's not clear what `lint-file` is in the usage, so this usage appears undefined to me.  If we want to allow positional arguments, it seems we need to decide if that is the input state file(s) or the rules file(s).   Can the underlying library support multiple rules files?  It may make sense to allow a single input state file via `-s,--state` and multiple rules files as the positional arguments.

cmd/file_lint.go
rspurgeon: I don't think we should support `--format default`, that's ambiguous. "default" should be specified in the help documentation as to what happens when you don't provide a value of either `json` or `yaml` for `--format`

GGabriele: I agree `--format default` shouldn't be supported, but I also think `default` is not a good name for this. If we change it to a better name, then the message may have this other option too. WDYT?

rspurgeon: I think this flag should be:
```
--format string          Output format: yaml or json. (default "yaml")
```

which matches the other deck file commands

GGabriele: So, just to clarify, no "human readable" option?

mheap: I think we need a human readable option. Agree with renaming away from default. The `plaintext` suggestion seems reasonable to me

rspurgeon: > So, just to clarify, no "human readable" option?

Sorry, I was confused by the purpose of the flag here, I understand now.   I agree @GGabriele, but maybe a value of `plain` is clear enough?  Or `plaintext' also works

cmd/file_lint.go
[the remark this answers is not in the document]
GGabriele: > Can the underlying library support multiple rules files?

The library gets a single input, so it's a matter of implementing the logic in decK to merge the rulesets.

It may make sense to just support a single state file and a single ruleset for starters. WDYT?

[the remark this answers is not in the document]
mheap: I don’t see a use case for supporting multiple at this point. Users can run the tool multiple times, or use the merge function to generate a single input

cmd/file_lint.go
```diff
@@ -147,19 +171,27 @@ func executeLint(cmd *cobra.Command, _ []string) error {
 func newLintCmd() *cobra.Command {
 	lintCmd := &cobra.Command{
 		Use:   "lint [flags] lint-file",
-		Short: "Short description here",
-		Long:  `Long description Here`,
-		RunE:  executeLint,
+		Short: "Lint a file against a ruleset",
+		Long: "Lint a file against a ruleset. \n" +
```
GGabriele, line 175: @mheap @rspurgeon can you please validate these?

rspurgeon: I'm still not sure what `lint-file` represents in this usage.  I think we're using flags for both the state file and the rules file, so what is `lint-file`?

As far as help text:
```suggestion
		Long: "Validate a decK state file against a linting ruleset, reporting any violations or failures. Report output can be returned in JSON, YAML, or human readable format (see --format). \n" +
```

rspurgeon: I also think it's important we document the `--fail-severity` flag and how it affects the return code of the command. I assume that users of this will want to branch behaviors based on command return code.   I think we should provide details here.  I would make a suggestion, but I'm not sure at this point how the code works in that regard.

GGabriele: Thanks for the suggestion!

Fixed this. https://github.com/Kong/deck/pull/981/commits/3bdd2fe76030a629dea7e41ca3b1cc94942219c1

GGabriele: > I would make a suggestion, but I'm not sure at this point how the code works in that regard.
Basically `--fail-severity` will gate what returns a non-zero (1) exit code.

Also, is `--display-only-failures` expected to care about `error` only or it should adapt to the `--fail-severity` value? If  the `--display-only-failures` flag should care only about `error` I suggest we change it to `--display-only-errors`.

Thoughts? @rspurgeon @mheap

mheap: `--fail-severity` and `--display-only-failures` are taken from Spectral, which is the defacto implementation of linting https://docs.stoplight.io/docs/spectral/9ffa04e052cc1-spectral-cli

We should adapt to the `--fail-severity` value

From Spectral:

> -D, --display-only-failures only output results equal to or greater than --fail-severity [boolean] [default: false]

GGabriele: @mheap thanks for the clarifying!

https://github.com/Kong/deck/pull/981/commits/a4e025cff686d8ec0e68422901e621c141a7cbc7

GGabriele (dismissed):

cmd/file_lint.go
rspurgeon: Can we just adopt what the spectral CLI supports directly?  Can we update our usage to be:
```
-F, --fail-severity results of this level or above will trigger a failure exit code
[string] [choices: "error", "warn", "info", "hint"] [default: "error"]
-D, --display-only-failures only output results equal to or greater than --fail-severity [boolean] [default: false]
```

rspurgeon:
Should we output the severity in plaintext mode?  It is provided in JSON or YAML format and I think is helpful for adjusting the rules or severity threshold.

```
./deck file lint -s ~/dev/tmp/kong-mock.yaml -r ~/dev/tmp/ruleset.yaml
Linting Violations: 2
Failures: 2

[10:15] Must use HTTPS protocol: `http` does not match the expression `^https`
[2:18] Check the version is correct: `3.0` does not match the expression `^1.1$`
```

Maybe something like:
```
[error][10:15] Must use HTTPS protocol: `http` does not match the expression `^https`
[warn][2:18] Check the version is correct: `3.0` does not match the expression `^1.1$`
```

GGabriele:
@rspurgeon I think I addressed both your latest comments. Let me know!

rspurgeon (approved):
LGTM, impressive work @mheap and @GGabriele

mikefero (approved):
LGTM. @mheap Do we plan on creating docs for this new feature and give some guidance to users (or even create a few real world examples)?

Tieske (dismissed):

Tieske:
> As discussed with @Tieske , I'd prefer we hold this for a new release version instead of merging into a revision release. We want to bundle this with other APIops related features that are slated for completion the coming weeks. Part of that work includes public docs, blog announcement, etc...

@rspurgeon this no longer holds right? we can merge once approved?

rspurgeon:
> @rspurgeon this no longer holds right? we can merge once approved?

Yes, thank you... We are clear to ship incremental (non-breaking) features in minor releases.   We should provide sufficient documentation both in product and on public docs site before merging, however.

GGabriele (approved):

PR #985, opened 2023-07-26
fix: correct diffs when comparing objects with empty tags

GGabriele:
Right now, when some declarative configuration having empty tags on any of its entities is deployed, subsequent `diff` / `sync` will show misleading results:

```
$ cat kong.yaml
_format_version: "3.0"
services:
  - name: foo
    protocol: http
    host: foo
    port: 8000
    tags: []

$ deck sync
creating service foo
Summary:
  Created: 1
  Updated: 0
  Deleted: 0

$ deck sync
updating service foo  {
   "connect_timeout": 60000,
   "enabled": true,
   "host": "foo",
   "id": "afadd2ed-2bcd-49fc-9000-adb58cd11b50",
   "name": "foo",
   "port": 8000,
   "protocol": "http",
   "read_timeout": 60000,
   "retries": 5,
   "write_timeout": 60000
 }

Summary:
  Created: 0
  Updated: 1
  Deleted: 0
```

The second `deck sync` run is showing an update without an actual diff. This is due to the way empty tags are treated internally in decK.

This commit fixes this defect.

### DIFF
--- a/state/types.go
+++ b/state/types.go
@@ -98,6 +98,13 @@ func (s1 *Service) EqualWithOpts(s2 *Service,
 	s1Copy := s1.Service.DeepCopy()
 	s2Copy := s2.Service.DeepCopy()
 
+	if len(s1Copy.Tags) == 0 {
+		s1Copy.Tags = nil
+	}
+	if len(s2Copy.Tags) == 0 {
+		s2Copy.Tags = nil
+	}
+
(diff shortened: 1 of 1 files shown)


Tieske (approved):
Should there be a change log entry?

GGabriele:
> Should there be a change log entry?

done

Tieske (approved):

PR #986, opened 2023-07-27
fix: correct consumers handling when custom_id is used

GGabriele:
Right now, when a consumer has the `username` equal to the `custom_id` of another consumer, decK fails because of the way the consumers are handled in the internal in-memory DB.

```
_format_version: "3.0"
consumers:
- custom_id: foo
  id: ce49186d-7670-445d-a218-897631b29ada
  username: Foo
- custom_id: bar
  id: 7820f383-7b77-4fcc-af7f-14ff3e256693
  username: foo
```

```
$ deck sync
Error: inserting consumer into state: inserting consumer Foo: entity already exists
```

This commit fixes this defect by modifying the way consumers are looked up in the in-memory DB.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,9 +67,10 @@
 
 ### Fixes
 
+- Fix Consumers handling when a consumer's `custom_id` is equal to the `username` of another consumer.
+  [#986](https://github.com/Kong/deck/pull/986)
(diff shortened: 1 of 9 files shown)


teb510 (approved):
looked at the code - didn't have time to test today but lgtm (for whatever that's worth!).

Tieske (approved):
lgtm

PR #987, opened 2023-07-28
feat: add wasm filter-chain entity support

hishamhm:
This brings support for wasm filter chains to deck.

Most of the groundwork for this was laid in the other repos:

* https://github.com/Kong/go-kong/pull/362
* https://github.com/Kong/go-database-reconciler/pull/72

Wasm filters were added in 3.4 and are an optional feature. ~Feature detection in deck works by fetching the root admin API endpoint and inspecting the configuration to see if `wasm` is enabled (this works for versions <3.4 and versions >=3.4 with wasm disabled). Feature detection does not run in Konnect mode; therefore Konnect is not yet supported, only Kong gateway OSS/EE~ **EDIT:** Feature detection works by calling `GET /filter-chains` and inspecting the response code ([link](https://github.com/Kong/go-database-reconciler/pull/101)).

For the sake of keeping things reasonable in size (we're at >3k new lines of code if you include the other 2 PRs), there are some validation steps that have been left out:

* validating filter chain names against those that the admin API tells us are installed/available
* validating filter configurations against the JSON schema from the admin API

These things are of course still validated by the Kong admin API, but at a later date it will be possible to validate them client-side for better UX.

---

KAG-4005
KOKO-1275


flrgh:
See also: https://github.com/Kong/go-database-reconciler/pull/72

flrgh:
https://github.com/Kong/go-database-reconciler/pull/72 has been merged. Awaiting the merge of https://github.com/Kong/go-database-reconciler/pull/90 and for a new release of go-database-reconciler now.

cmd/common.go
```diff
@@ -408,6 +416,63 @@ func determineSelectorTag(targetContent file.Content, config dump.Config) ([]str
 	return config.SelectorTags, nil
 }
 
+func determineFilterChainSupport(ctx context.Context, client *kong.Client) (bool, error) {
+	workspace := client.Workspace()
+	client.SetWorkspace("")
+	defer client.SetWorkspace(workspace)
```
flrgh, line 422: This feels like a mega hack, but I get test failures without it because sometimes tests run against a workspace that doesn't exist. Is there a better way?

flrgh:
re: `CI Test / test (pull_request) Failing after 4m`, this is due to us getting ratelimited by Codecov:

> [2024-05-21T16:04:17.142Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 429 - {'detail': ErrorDetail(string='Rate limit reached. Please upload with the Codecov repository upload token to resolve issue. Expected time to availability: 2452s.', code='throttled')}
Error: Codecov: Failed to properly upload: The process '/home/runner/work/_actions/codecov/codecov-action/v3/dist/codecov' failed with exit code 255

go.mod
```diff
@@ -4,6 +4,9 @@ go 1.21.6
 
 replace github.com/yudai/gojsondiff v1.0.0 => github.com/Kong/gojsondiff v1.3.0
 
+// TODO: remove me
+replace github.com/kong/go-database-reconciler v1.10.0 => github.com/kong/go-database-reconciler v0.0.0-20240513201015-0cc539e9aad8
```
flrgh, line 8: Note: this PR is blocked until we get a new release of go-database-reconciler (preferably after [this PR](https://github.com/Kong/go-database-reconciler/pull/90) is merged).

flrgh:
:wave: @GGabriele do you think you could review or help source a reviewer for this PR?

cmd/common.go
```diff
@@ -318,6 +325,9 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 	if err := checkForRBACResources(*rawState, dumpConfig.RBACResourcesOnly); err != nil {
 		return err
 	}
+	if err := checkFilterChainsAllowed(*rawState, dumpConfig, parsedKongVersion); err != nil {
```
GGabriele, line 328: Is this even needed? If decK attempts to create a filter chain on a Gateway that doesn't support it, can't we just return the API error back to the caller?

Right now decK is mostly best effort when it comes to features driven by Kong configuration, for example decK doesn't check whether a custom plugin exists before attempting to configure it.

cmd/common.go
GGabriele: I added a comment related to this here: https://github.com/Kong/deck/pull/987/files#r1629376245

In case we need to do this check, do you think we can modify the [fetchKongVersion](https://github.com/Kong/deck/blob/v1.38.1/cmd/common.go#L477) function to also deal with it?

GGabriele (commented):
Can we please add some integration tests too?

cmd/common.go
flrgh: If we don't do some kind of feature detection, pretty much all deck commands will throw an error when executing against a Kong node without wasm support (too old version, not turned on, etc) because `GET /filter-chains` will return 404/405.

So you are correct that `checkFilterChainsAllowed` is not _strictly_ required, but `determineFilterChainSupport` (feature detection) is. I would prefer to keep the `checkFilterChainsAllowed` guard in place because it's nicer on the user to fail fast and protects against things like incomplete sync, but I can remove it if you think it's just not a good fit with the rest of the codebase.

cmd/common.go
flrgh: Ah, so `fetchKongVersion()` is actually where I got the idea for this. It does something similar by [trying to fetch / before trying the /{workspace} endpoint](https://github.com/Kong/deck/blob/526baa380db173161fe113a2580a2dca7f472155/cmd/common.go#L469-L485).

GGabriele: I suggested to reuse `fetchKongVersion()` so that we will not have to run the same query twice.

Also, the current code would fail with 401 if decK doesn't have the permissions to query the default workspace.

cmd/common.go
GGabriele: > GET /filter-chains will return 404/405.

For reference, we do this already for Consumer Groups: https://github.com/Kong/go-database-reconciler/blob/main/pkg/dump/dump.go#L90

I'm okay if you prefer to do this extra check though.

flrgh: Ahhhhh interesting. I think I prefer it that way actually. I'll do some testing and push this logic to go-database-reconciler.

flrgh: https://github.com/Kong/go-database-reconciler/pull/101

flrgh:
Okay, found another wart in testing. PR for fixage his here:
https://github.com/Kong/go-database-reconciler/pull/103

tests/integration/testdata/filters/response_transformer.wasm
flrgh, line 1: It's only 2MB, but I acknowledge that putting a binary file into vcs is not great.

How do we feel about this? Okay to leave it as-is, or should I add a script/Makefile target to fetch it on demand when running the integration tests?

This one in particular is sourced from https://github.com/Kong/proxy-wasm-rust-response-transformer/releases/tag/0.2.0, so it is readily available for download without requiring any secrets.

.ci/lib.sh
```diff
@@ -0,0 +1,86 @@
+set -euo pipefail
```
flrgh, line 1: I found it rather clunky to get the integration tests running locally, so this includes changes I previously made to go-database-reconciler:

https://github.com/Kong/go-database-reconciler/commit/3bf528eff89a1b26eac17b67d4e399a404c16ba7

flrgh:
Okay, https://github.com/Kong/go-database-reconciler/pull/101 has been merged, but we're blocked on getting a version bump.

I've also submitted a semi-related bugfix PR to gdr ([link](https://github.com/Kong/go-database-reconciler/pull/103)). I think it would be wise to merge this before the version bump so that it is included with filter chain support from the get-go.

flrgh:
@GGabriele sorry for the turnaround time. This is ready for another round of review. Let me know what you think of the integration tests--anything else that you think would particularly benefit from more coverage?

tests/integration/testdata/filters/response_transformer.wasm
GGabriele: I personally don't have any qualms about directly embedding this here if needed for testing, both local and in the CI.

GGabriele:
@flrgh a new version of [go-database-reconciler](https://github.com/Kong/go-database-reconciler/releases/tag/v1.12.2) has been released now.

flrgh:
> @flrgh a new version of [go-database-reconciler](https://github.com/Kong/go-database-reconciler/releases/tag/v1.12.2) has been released now.

Awesome! Bumped and added test coverage.

PR #999, opened 2023-08-10
fix: set default strip_path to true

GGabriele:
In the past decK had to set the default value of the `strip_path` field to [false](https://github.com/Kong/deck/blob/v1.26.0/utils/constants.go#L22) because of some limitation in the library used for defaults injection: this was a workaround to allow this field to be set to `false`. Then, if the field was not set in the configuration, the field was then turned into `true` at runtime via inspection of the Kong Admin API.

This commit changes the decK default for the `strip_path` field to `true`, since now the underlying library has fixed the issue that triggered this workaround in the first place.

### DIFF
--- a/convert/testdata/5/output-expected.yaml
+++ b/convert/testdata/5/output-expected.yaml
@@ -22,7 +22,7 @@ services:
     - http
     - https
     regex_priority: 0
-    strip_path: false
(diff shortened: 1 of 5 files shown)


zekth (approved):
:v:

PR #1007, opened 2023-08-21
docs(cli): Add a long description for file render

lena-larionova:
The existing description for `deck file render` doesn't currently provide any detail on what the command does or how to use it. 
I was trying to test it out and couldn't even figure out how it was supposed to work, since there's no input file flag or example. 

Found the info I needed on Slack here, and based the update on that: https://kongstrong.slack.com/archives/C04349E4KRC/p1690921140291889

❗ I did run into one potential issue while testing: this command doesn't warn you when overwriting existing files, like `deck dump` or others do. Is that intentional?

### DIFF
--- a/cmd/file_render.go
+++ b/cmd/file_render.go
@@ -27,8 +27,19 @@ func executeFileRenderCmd(_ *cobra.Command, _ []string) error {
 func newFileRenderCmd() *cobra.Command {
 	renderCmd := &cobra.Command{
 		Use:   "render",
-		Short: "Render the configuration as Kong declarative config",
-		Long:  ``,
+		Short: "Combines multiple complete configuration files into one Kong declarative config file.",
+		Long:  `Combines multiple complete configuration files into one Kong
+	declarative config file.
+
+ This command can render the output in JSON or YAML format. Unlike 
+ "deck file merge", the render command accepts complete configuration files, 
+ while "deck file merge" only combines partial file snippets.
+
+ For example, the following command takes two input files and renders them as one 
+ combined JSON file:
+
+   deck file render kong1.yml kong2.yml -o kong3 --format json
+	`,
 		Args:  cobra.ArbitraryArgs,
 		RunE:  executeFileRenderCmd,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
--- a/cmd/file_render.go
+++ b/cmd/file_render.go
@@ -28,24 +28,24 @@ func newFileRenderCmd() *cobra.Command {
 	renderCmd := &cobra.Command{
 		Use:   "render",
 		Short: "Combines multiple complete configuration files and renders them as one Kong declarative config file.",
-		Long:  `Combines multiple complete configuration files and renders them as one Kong
-	declarative config file.
+		Long: `Combines multiple complete configuration files and renders them as one Kong
+declarative config file.
 
- This command renders a full declarative configuration in JSON or YAML format by assembling 
- multiple files and populating defaults and environment substitutions. 
- This command is useful to observe what configuration would be sent prior to synchronizing to 
- the gateway.
+This command renders a full declarative configuration in JSON or YAML format by assembling 
+multiple files and populating defaults and environment substitutions. 
+This command is useful to observe what configuration would be sent prior to synchronizing to 
+the gateway.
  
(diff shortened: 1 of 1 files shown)


rspurgeon:
cc. @zekth

cmd/file_render.go
```diff
@@ -27,8 +27,19 @@ func executeFileRenderCmd(_ *cobra.Command, _ []string) error {
...
+	declarative config file.
+
+ This command can render the output in JSON or YAML format. Unlike 
+ "deck file merge", the render command accepts complete configuration files, 
+ while "deck file merge" only combines partial file snippets.
+
+ For example, the following command takes two input files and renders them as one 
+ combined JSON file:
+
+   deck file render kong1.yml kong2.yml -o kong3 --format json
```
zekth, line 41:
```suggestion
   deck file render kong1.yml kong2.yml -o kong3.json --format json
```
Should be more accurate

lena-larionova: I tried both ways, and they both worked. I _also_ accidentally tried `kong3.yml --format json` and ended up with a file named kong3.yaml but the contents were json. Should there be any error or something around that 🤔

zekth: @Tieske is it an expected behavior?

zekth: @lena-larionova disregard

zekth (dismissed):

Tieske:
@zekth iirc this command also renders the files, as they would be synced to the control plane (kong or Konnect), such that it will also do substitutions etc. That is currently not in the description, should it be? including maybe a warning that secrets might be in the rendered file?

zekth:
currently render use `mocks` for values, so secrets won't be rendered.

ref: 
https://github.com/Kong/deck/blob/20ef81eb32b3de4023ea815d3db02dabf722fa02/file/reader.go#L36
https://github.com/Kong/deck/blob/20ef81eb32b3de4023ea815d3db02dabf722fa02/cmd/file_render.go#L24

rspurgeon:
@lena-larionova @zekth can we get this updated with the one suggestion (or something similar) and get it merged?

lena-larionova:
> this command also renders the files, as they would be synced to the control plane (kong or Konnect), such that it will also do substitutions etc.

How's this @Tieske @zekth :
Short description: "Combines multiple complete configuration files and renders them as one Kong declarative config file."

Then some questions about the rest:

> currently render use mocks for values, so secrets won't be rendered.

Does this mean the rendered file isn't in a ready state for syncing to a control plane, as the user has to replace mocked values with their own? Or am I misunderstanding something.

> such that it will also do substitutions etc.  That is currently not in the description, should it be?

What should be said about substitutions? What's the etc part? Is that something we should list out? What exactly should we add to the long description here?

Tieske:
> > currently render use mocks for values, so secrets won't be rendered.
> 
> Does this mean the rendered file isn't in a ready state for syncing to a control plane, as the user has to replace mocked values with their own? Or am I misunderstanding something.

No, the goal is to validate local, so users would not edit manually, but if successful do an actual sync with actual values (not mocked ones).

Deck does;
1. grab files
2. combine them
3. render variables and "other magic"
4. sync to CP

especially step 3 & 4 were a single operation, without an option to inspect the output of step 3. The new `render` command allows a user to execute the same steps, but without the sync, so they can validate locally. (otherwise the user would need a running Kong system to test the sync against)


> > such that it will also do substitutions etc.  That is currently not in the description, should it be?
> 
> What should be said about substitutions? What's the etc part? Is that something we should list out? What exactly should we add to the long description here?

Above I named it "other magic" because I do not know exactly what deck logic is executed here. Probably @GGabriele knows. If there is any additional logic, and if so, what it exactly is.

rspurgeon:
I think my suggestion above covers enough of the use case here to apply it and merge this down

rspurgeon (changes requested):
left a suggestion

lena-larionova:
@zekth @Tieske Added Rick's suggestion, how does that look?

Also added the term "render" back into the short description. That makes the description a bit long - is there a character limit I need to be aware of?

rspurgeon (approved):
LGTM

zekth (approved):
thanks

Tieske (approved):

GGabriele (approved):

PR #1015, opened 2023-08-30
fix(docs/cmd) Change go-apiops in remove-tags examples to deck file

lena-larionova:
Parallel to docs PR: https://github.com/Kong/docs.konghq.com/pull/6037

Examples are telling users to run `go-apiops`, which is only supposed to be used for dev/testing. Changing that to the standard `deck file`.


Tieske (approved):

Tieske:
good catch!

PR #1050, opened 2023-10-10
feat(convert): convert to KIC

Tieske:
New command "deck file kong2kic" Kong configuration file to Kong Ingress Controller manifests.

Currently, only ingress controller manifests are supported. Output format is always YAML.

```
Usage:
  deck file kong2kic [flags]

Flags:
  -f, --format string        output file format: json or yaml. (default "yaml")
  -h, --help                 help for kong2kic
  -o, --output-file string   Output file to write. Use - to write to stdout. (default "-")
  -s, --state string         decK file to process. Use - to read from stdin. (default "-")
      --style string         Generate manifests with annotations in Service and Ingress, or using the KongIngress CRD: annotation or crd. (default "annotation")
```

- Global Plugins map to KongClusterPlugins
- Kong Services map to K8s Services. Kong Routes map to K8s Ingresses. To specify Kong specific features, K8s Ingress and Service resources eventually have a KongIngress resource associated to them.
- Kong Consumer entities map to KongConsumer manifests.

Jira: APIOPS-8


mheap:
> To specify Kong specific features, K8s Ingress and Service resources eventually have a KongIngress resource associated to them.

Most things in `KongIngress` have been moved to annotations on the Ingress. Can you share the list of KongIngress uses and we can help migrate to the annotation format

go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	github.com/imdario/mergo v0.3.16
 	github.com/kong/go-apiops v0.1.21
 	github.com/kong/go-kong v0.46.0
+	github.com/kong/kubernetes-ingress-controller/v2 v2.11.1
```
mflendrich, line 22: This line is effectively adding a module dependency cycle between KIC and decK that we cannot make happen.

Since KIC uses decK as a library, it won't be possible for decK to import KIC as a module. A possible resolution for this would be:
* introduction of an "umbrella" project over the existing decK that would decouple "deck-as-a-library" from "deck-as-a-tool" or
* a massive refactor of deck or KIC.

mflendrich (changes requested):
Discussed with @Tieske on zoom earlier today.

We concluded that there are two blockers / next steps for this PR:
1. the implementation needs to avoid creating the deck <-> KIC dep cycle,
2. more alignment (an approved product brief / design doc) is needed.

go.mod
battlebyte: The reason for this dependency is to make use of the API Schema definitions for `KongClusterPlugin`, `KongConsumer`, `KongIngress`, `KongPlugin`, etc. Would it be possible to extract the API Schema definitions into its own module that is imported by both deck and KIC? Similarly to what k8s does here: https://github.com/kubernetes/api

battlebyte: Essentially, extract type definitions under `https://github.com/Kong/kubernetes-ingress-controller/tree/main/pkg/apis/configuration` into its own module.

mflendrich: The eng/PM decision reached is that we'll unblock `deck2kic`'s dependency on KIC by implementing the proposal tracked by https://github.com/Kong/deck/issues/1060.

Tieske:
@mflendrich @GGabriele @battlebyte 

I have cherry-picked the #1117 changes in here, rebased and fixed up.

Please give this another review. This should be close to merging now I think.

cmd/file_kong2kic.go
```diff
@@ -0,0 +1,122 @@
...
+		strings.ToUpper(cmdKong2KicAPI) == "GATEWAY" {
+		outputFileFormat = kong2kic.KICJSONGatewayAPI
+	} else if strings.ToUpper(cmdKong2KicOutputFormat) == "YAML" &&
+		strings.ToUpper(cmdKong2KicAPI) == "GATEWAY" {
+		outputFileFormat = kong2kic.KICYAMLGatewayAPI
+	} else {
+		return fmt.Errorf("invalid combination of output format and manifest style")
+	}
+
+	err = file.WriteContentToFile(outputContent, cmdKong2KicOutputFilename, outputFileFormat)
```
Tieske, line 74:
```suggestion
	err = kong2kic.WriteContentToFile(outputContent, cmdKong2KicOutputFilename, outputFileFormat)
```

cmd/file_kong2kic.go
```diff
@@ -0,0 +1,123 @@
...
+	} else {
+		kong2kic.GatewayAPIVersion = "gateway.networking.k8s.io/v1"
+	}
+	kong2kic.ClassName = CmdKong2KicClassName
+	verbosity, _ := cmd.Flags().GetInt("verbose")
+	logbasics.Initialize(log.LstdFlags, verbosity)
+
+	inputContent, err := file.GetContentFromFiles([]string{cmdKong2KicInputFilename}, false)
+	if err != nil {
+		return fmt.Errorf("failed reding input file '%s'; %w", cmdKong2KicInputFilename, err)
```
GGabriele, line 44:
```suggestion
		return fmt.Errorf("failed reading input file '%s'; %w", cmdKong2KicInputFilename, err)
```

cmd/file_kong2kic.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	outputContent = inputContent.DeepCopy()
+	if strings.ToUpper(cmdKong2KicOutputFormat) == "JSON" &&
```
GGabriele, line 48: Since these strings are reused more than once, whould we have these `JSON`, `CRD` etc defined as constants?

cmd/file_kong2kic.go
```diff
@@ -0,0 +1,123 @@
...
+		strings.ToUpper(cmdKong2KicManifestStyle) == "ANNOTATION" {
+		outputFileFormat = kong2kic.KICYAMLAnnotationIngressAPI
+	} else if strings.ToUpper(cmdKong2KicOutputFormat) == "JSON" &&
+		strings.ToUpper(cmdKong2KicAPI) == "GATEWAY" {
+		outputFileFormat = kong2kic.KICJSONGatewayAPI
+	} else if strings.ToUpper(cmdKong2KicOutputFormat) == "YAML" &&
+		strings.ToUpper(cmdKong2KicAPI) == "GATEWAY" {
+		outputFileFormat = kong2kic.KICYAMLGatewayAPI
+	} else {
+		return fmt.Errorf("invalid combination of output format and manifest style")
```
GGabriele, line 72: Should we suggest a way to fix the input?

cmd/file_kong2kic.go
GGabriele, line 48: Should we have all this into a separate `validateInput` function?

kong2kic/types.go
```diff
@@ -0,0 +1,225 @@
...
+	Ingresses          []k8snetv1.Ingress             `json:"ingresses,omitempty" yaml:",omitempty"`
+	Services           []k8scorev1.Service            `json:"services,omitempty" yaml:",omitempty"`
+	Secrets            []k8scorev1.Secret             `json:"secrets,omitempty" yaml:",omitempty"`
+	KongConsumers      []kicv1.KongConsumer           `json:"consumers,omitempty" yaml:",omitempty"`
+	KongConsumerGroups []kicv1beta1.KongConsumerGroup `json:"consumerGroups,omitempty" yaml:",omitempty"`
+	HTTPRoutes         []k8sgwapiv1.HTTPRoute         `json:"httpRoutes,omitempty" yaml:",omitempty"`
+}
+
+func (k KICContent) marshalKICContentToYaml() ([]byte, error) {
+	var kongIngresses []byte
```
GGabriele, line 29: nit: you could also do the following:

```
var (
  kongIngresses []byte
  kongPlugins []byte
...
...
)
```

kong2kic/types.go
```diff
@@ -0,0 +1,225 @@
...
+
+	}
+
+	for _, kongConsumer := range k.KongConsumers {
+		kongConsumers, err = yaml.Marshal(kongConsumer)
+		if err != nil {
+			return nil, err
+		}
+		output = append(output, kongConsumers...)
+		output = append(output, []byte("---\n")...)
```
GGabriele, line 119: maybe we should have `"---\n"` defined as a constant?

GGabriele (dismissed):
LGTM, but please get a +2 from someone from @Kong/team-k8s

rainest:
> This line is effectively adding a module dependency cycle between KIC and decK that we cannot make happen.

Oh, this is what prompted that split? We were probably fine without that, the API packages are pretty isolated from the rest of the code and don't import any of the parser code that loads deck. The parser code does import the API, but that's just a duplicate leaf, not a cycle.

Rest of my review was mostly just looking over output-expected.yamls, since AFAIK the code's already been reviewed and we're mostly concerned with the end output. Dunno if we'd block this as-ism but we'll eventually need KIC 3.x upgrades and should add some additional tests to confirm that the config is valid.

Development concurrent with KIC 3.x bit us:

- KongIngress still used. It looks like the one remaining one is for upstream configuration (the new Ingress annotations are used), which should be handled with [KongUpstreamPolicy](https://docs.konghq.com/kubernetes-ingress-controller/latest/guides/migrate/kongingress/#kongupstreampolicy) in 3.x.
- Several resources (at least non-global KongPlugins and credential Secrets--I didn't review exhaustively) include `kubernetes.io/ingress.class`, which they don't actually use. Including it is benign though, it just won't actually do anything. Offhand it doesn't look like we have a complete list of which do and do not need it, so probably just ignore this.
- Credential Secrets still use the `kongCredType` field instead of the [new label](https://docs.konghq.com/kubernetes-ingress-controller/latest/guides/migrate/credential-kongcredtype-label/). You can continue to include it for pre-3.0 compatibility but we will be removing support it within the next few 3.x releases.

Aside details, it looks like we just have expected output tests, not integration. Human review of large generated outputs will be error-prone, so I think we'd be better off tossing configuration into an actual instance to see if it's accepted. We have an [existing KIC E2E test](https://github.com/Kong/kubernetes-ingress-controller/blob/9ee78967746ea7c67a80d8cfdd951b52b283d6ce/test/e2e/features_test.go#L70-L232) that'd serve as a good starting point because it runs the admission webhook and is basically self-contained (the test function spawns its own Kubernetes cluster and installs KIC+Kong; it doesn't require any additional setup), though we'd probably want to simplify [the CI workflow](https://github.com/Kong/kubernetes-ingress-controller/blob/9ee78967746ea7c67a80d8cfdd951b52b283d6ce/.github/workflows/_e2e_tests.yaml) a bit. With that we can have the test exec a `kubectl apply` to apply the output resources and check if the command fails (the webhook or static validation rejected something) or if a [failure event](https://docs.konghq.com/kubernetes-ingress-controller/latest/reference/troubleshooting/#finding-problem-resource-events) shows up.

mflendrich (commented):
> the API packages are pretty isolated from the rest of the code and don't import any of the parser code that loads deck. The parser code does import the API, but that's just a duplicate leaf, not a cycle.

This can work in the context of low-level package mechanics but stops working when we consider the larger-scale dependency hell problem. If we allowed KIC version X to pull in deck version Y pulling in KIC version Z, this would mean an (effectively) irreversible need to make coordinated changes in lockstep (between separately governed projects) from this point onwards. The opportunity cost of a dependency hell that we've avoided is the real win here, not "code compiles but it otherwise wouldn't". 

To the point of the review:
- Now that the deck<->KIC issue is resolved, my reason for "changes requested" is gone.
- As Travis pointed out, KIC's API has evolved - `KongIngress` and `kongCredType` are deprecated as of KIC 3.0. This means that the proposed implementation will generate configuration that is considered deprecated and may cause validation warnings (+ will stop working in a future version of KIC). I think this issue is serious enough to fix in this PR, but I'll leave it to the APIOps folks so that we don't have too many cooks in the kitchen.
- Not necessarily for this PR: Travis's point about testing. Our k8s integration testing experience is that with new k8s versions and KIC/Kong versions you'll be bumping into the "configuration that was working has stopped working" class of errors that will be missed (like we could have missed the KIC 3.0 API changes in the bullet point above). My (soft) review recommendation is what Travis says: instead of "golden output" testing, have an integration test involving KIC.

TL;DR:
- the dependency blocker is gone, I'm removing my "changes requested"
- did not review for code style/correctness
- non-blocking (but very important) recommendation to upgrade the API to KIC 3.0+
- non-blocking suggestion to consider integration testing

kong2kic/testdata/custom_resources/json/output-expected.json
```diff
@@ -0,0 +1,835 @@
...
+        ],
+        "namespace": "example_namespace",
+        "strategy": "local",
+        "sync_rate": -1,
+        "window_size": [
+            30
+        ]
+    },
+    "plugin": "rate-limiting-advanced",
+    "status": {}
```
czeslavo, line 168: I think we should make sure that no `status` or `metadata.creationTimestamp` fields (or any other of that kind) are part of the output. These are fields that are meant to be set by the Kubernetes API server or the owning controller. 

I imagine this could be done generically for every object serialized like so:
```go
func SerializeObjectDropping(obj interface{}, fieldsToDrop ...string) ([]byte, error) {
	objBytes, err := json.Marshal(obj)
	if err != nil {
		return nil, err
	}
	genericObj := map[string]interface{}{}
	if err := json.Unmarshal(objBytes, &genericObj); err != nil {
		return nil, err
	}
	for _, field := range fieldsToDrop {
		delete(genericObj, field)
	}
	return yaml.Marshal(genericObj) // or json.Marshal, depending on your needs.
}
```
Potentially this could be also done with some `reflect` usage to avoid the first `json.Marshal`, but I'm not sure which one would be more performant and how much we're concerned about the performance in this case.

mheap:
@czeslavo I believe @battlebyte has made the requested changes. Could you take a final look and ✅ if you're happy?

mheap:
Ah, I just noticed no additional commits on the branch. I couldn't see `status` in the docs examples and assumed that this was complete. @battlebyte Did you see Greg's comment above?

mheap:
See #1218 for the implementation of all feedback

Tieske:
@czeslavo @battlebyte @GGabriele @mflendrich 
This now incorporates the changes in #1218 . Should be good now, re-requested reviews, so please give it one more pass.

Then we can finally merge.

battlebyte (dismissed):
Looks good to me!

battlebyte:
This is the final command reference:
```
Convert Kong configuration files to Kong Ingress Controller (KIC) manifests.

The kong2kic subcommand transforms Kong's configuration files, written in the deck format, 
into Kubernetes manifests suitable for the Kong Ingress Controller. By default kong2kic generates 
manifests for KIC v3.x using the Kubernetes Gateway API. Only HTTP/HTTPS routes are supported.

Usage:
  deck file kong2kic [flags]

Flags:
      --class-name string    Value to use for "kubernetes.io/ingress.class" ObjectMeta.Annotations and for
                                        "parentRefs.name" in the case of HTTPRoute. (default "kong")
  -f, --format string        output file format: json or yaml. (default "yaml")
  -h, --help                 help for kong2kic
      --ingress              Use Kubernetes Ingress API manifests instead of Gateway API manifests.
      --kicv2                Generate manifests compatible with KIC v2.x.
  -o, --output-file string   Output file to write. Use - to write to stdout. (default "-")
  -s, --state string         decK file to process. Use - to read from stdin. (default "-")
```

battlebyte:
@rspurgeon yes, that is something I considered as well. I'm fine changing it, no problem. Any thoughts on that @mheap ?

mheap:
Big +1 for that change

mheap:
@GGabriele @czeslavo Could you please take a look? I'd like to get this released with the other changes on `main`

cmd/file_kong2kic.go
```diff
@@ -0,0 +1,139 @@
...
+	outputContent = inputContent.DeepCopy()
+	err = kong2kic.WriteContentToFile(outputContent, cmdKong2KicOutputFilename, outputFileFormat, yamlOrJSON)
+	if err != nil {
+		return fmt.Errorf("failed converting Kong to Ingress '%s'; %w", cmdKong2KicInputFilename, err)
+	}
+
+	return nil
+}
+
+func validateInput(cmdKong2KicOutputFormat string) (
```
GGabriele, line 61: I think this function can benefit from a small refactor using a switch/case which would limit some repetitions (e.g. set `yamlOrJSON` only once)

This is tot a blocking comment.

cmd/file_kong2kic.go
```diff
@@ -0,0 +1,139 @@
...
+		outputFileFormat file.Format
+		yamlOrJSON       string
+	)
+
+	kong2kic.ClassName = cmdKong2KicClassName
+	verbosity, _ := cmd.Flags().GetInt("verbose")
+	logbasics.Initialize(log.LstdFlags, verbosity)
+
+	// if cmdKong2KicVersion is not 2 or 3 return an error
+	if cmdKong2KicVersion != "2" && cmdKong2KicVersion != "3" {
```
GGabriele, line 38: since you already have a `validateInput` helper function, it would make sense to consolidate all input validation there.

cmd/file_kong2kic.go
```diff
@@ -0,0 +1,139 @@
...
+	} else {
+		err = fmt.Errorf("invalid combination of parameters. Please use --help for more information")
+	}
+
+	return outputFileFormat, yamlOrJSON, err
+}
+
+//
+//
+// Define the CLI data for the openapi2kong command
```
GGabriele, line 107:
```suggestion
// Define the CLI data for the kong2kic command
```

kong2kic/types.go
```diff
@@ -0,0 +1,259 @@
...
+	Ingresses            []k8snetv1.Ingress              `json:"ingresses,omitempty" yaml:",omitempty"`
+	Services             []k8scorev1.Service             `json:"services,omitempty" yaml:",omitempty"`
+	Secrets              []k8scorev1.Secret              `json:"secrets,omitempty" yaml:",omitempty"`
+	KongConsumers        []kicv1.KongConsumer            `json:"consumers,omitempty" yaml:",omitempty"`
+	KongConsumerGroups   []kicv1beta1.KongConsumerGroup  `json:"consumerGroups,omitempty" yaml:",omitempty"`
+	HTTPRoutes           []k8sgwapiv1.HTTPRoute          `json:"httpRoutes,omitempty" yaml:",omitempty"`
+	KongUpstreamPolicies []kicv1beta1.KongUpstreamPolicy `json:"upstreamPolicies,omitempty" yaml:",omitempty"`
+}
+
+func (k KICContent) marshalKICContentToYaml() ([]byte, error) {
```
GGabriele, line 30: are `marshalKICContentToJSON` and `marshalKICContentToYaml` the same except for the format used? If so, we should consolidate them into a single one:

```
func (k KICContent) marshalKICContentToFormat(format string) ([]byte, error) {
	var output []byte

	const (
		yamlSeparator = "---\n"
	)

	for _, kongIngress := range k.KongIngresses {
		kongIngresses, err := SerializeObjectDroppingFields(kongIngress, format)
		if err != nil {
			return nil, err
		}
		output = append(output, kongIngresses...)
		if format == file.YAML {
			output = append(output, []byte(yamlSeparator)...)
		}
	}
...
...
}
```

kong2kic/writer.go
```diff
@@ -0,0 +1,102 @@
...
+
+	"github.com/kong/go-database-reconciler/pkg/file"
+	"github.com/kong/go-database-reconciler/pkg/utils"
+)
+
+func WriteContentToFile(content *file.Content, filename string, format file.Format, yamlOrJSON string) error {
+	var c []byte
+	var err error
+
+	switch format {
```
GGabriele, line 17: Isn't it possible to do the following instead? (untested)

```
	switch yamlOrJSON {
	case file.YAML:
		c, err = MarshalKongToKICYaml(content, string(format))
	case file.JSON:
		c, err = MarshalKongToKICJson(content, string(format))
	}
	if err != nil {
		return err
	}
```

kong2kic/writer.go
```diff
@@ -0,0 +1,102 @@
...
+	// case YAML:
```
GGabriele, line 18: Leftover comments?

kong2kic/writer.go
```diff
@@ -0,0 +1,102 @@
...
+		if err != nil {
+			return err
+		}
+	default:
+		return fmt.Errorf("unknown file format: " + string(format))
+	}
+
+	if filename == "-" {
+		if _, err := fmt.Print(string(c)); err != nil {
+			return fmt.Errorf("writing file: %w", err)
```
GGabriele, line 70:
```suggestion
			return fmt.Errorf("writing to stdout: %w", err)
```

GGabriele (commented):
Added a few comments / suggestions.

I haven't reviewed `kong2kic/kong2kic.go` though, I'll leave that to @czeslavo since I lack domain knowledge to properly do that.

battlebyte:
Thanks @GGabriele, all your suggestions applied in commit d49ecd0.

GGabriele:
> Thanks @GGabriele, all your suggestions applied in commit [d49ecd0](https://github.com/Kong/deck/commit/d49ecd0019d4b3b65875c32913ea0d468155c059).

Much cleaner, thanks Jordi!

czeslavo (commented):
I left some comments. It's a massive volume of code to review and I'm not sure how deep we should go with the review here. 

Note: Given this is not integration-tested, I wouldn't be very comfortable with advertising the functionality as something ready to use in prod. Are we going to implement tests that apply the resulting manifests to a K8s cluster with KIC running, ensuring all the routes work as expected before releasing it? If not, I think this should be advertised as an experimental feature.

mheap:
RE: Integration tests, I'm comfortable with manual testing for the initial release. We should add integration tests as a followup

battlebyte:
Thanks @czeslavo, all your suggestions applied in commit [ba63ef9](https://github.com/Kong/deck/pull/1050/commits/ba63ef96b3fde973d8f4d58846848ea4447f65a1). I'm currently working on integration testing using the kubernetes testing framework. I will ping you in Slack for your advice on how to best approach this if that is fine with you.

kong2kic/upstream.go
```diff
@@ -0,0 +1,214 @@
...
+	if content.Upstreams != nil {
+		var kongIngress kicv1.KongIngress
+		kongIngress.APIVersion = KICAPIVersion
+		kongIngress.Kind = IngressKind
+		if service.Name != nil {
+			kongIngress.ObjectMeta.Name = calculateSlug(*service.Name + "-upstream")
+		} else {
+			log.Println("Service name is empty. This is not recommended." +
+				"Please, provide a name for the service before generating Kong Ingress Controller manifests.")
+		}
```
czeslavo, line 180:
```suggestion
			log.Println("Service name is empty. This is not recommended." +
				"Please, provide a name for the service before generating Kong Ingress Controller manifests.")
			continue
		}
```

kong2kic/consumer_group.go
```diff
@@ -0,0 +1,82 @@
...
+	// iterate over the consumer groups and create a KongConsumerGroup for each one
+	for _, consumerGroup := range content.ConsumerGroups {
+		var kongConsumerGroup kicv1beta1.KongConsumerGroup
+		kongConsumerGroup.APIVersion = "configuration.konghq.com/v1beta1"
+		kongConsumerGroup.Kind = "KongConsumerGroup"
+		if consumerGroup.Name != nil {
+			kongConsumerGroup.ObjectMeta.Name = calculateSlug(*consumerGroup.Name)
+		} else {
+			log.Println("Consumer group name is empty. This is not recommended." +
+				"Please, provide a name for the consumer group before generating Kong Ingress Controller manifests.")
```
czeslavo, line 23:
```suggestion
			log.Println("Consumer group name is empty. This is not recommended." +
				"Please, provide a name for the consumer group before generating Kong Ingress Controller manifests.")
			continue
```

kong2kic/route.go
```diff
@@ -0,0 +1,502 @@
...
+			)
+
+			k8sIngress.TypeMeta.APIVersion = "networking.k8s.io/v1"
+			k8sIngress.TypeMeta.Kind = "Ingress"
+			if service.Name != nil && route.Name != nil {
+				k8sIngress.ObjectMeta.Name = calculateSlug(*service.Name + "-" + *route.Name)
+			} else {
+				log.Println("Service name or route name is empty. This is not recommended." +
+					"Please, provide a name for the service and the route before generating Kong Ingress Controller manifests.")
+			}
```
czeslavo, line 41:
```suggestion
				log.Println("Service name or route name is empty. This is not recommended." +
					"Please, provide a name for the service and the route before generating Kong Ingress Controller manifests.")
				continue
			}
```

kong2kic/route.go
```diff
@@ -0,0 +1,502 @@
...
+				httpRoute.APIVersion = GatewayAPIVersionV1
+			} else {
+				httpRoute.APIVersion = GatewayAPIVersionV1Beta1
+			}
+			if service.Name != nil && route.Name != nil {
+				httpRoute.ObjectMeta.Name = calculateSlug(*service.Name + "-" + *route.Name)
+			} else {
+				log.Println("Service name or route name is empty. This is not recommended." +
+					"Please, provide a name for the service and the route before generating HTTPRoute manifests.")
+			}
```
czeslavo, line 292:
```suggestion
			} else {
				log.Println("Service name or route name is empty. This is not recommended." +
					"Please, provide a name for the service and the route before generating HTTPRoute manifests.")
				continue
			}
```

kong2kic/service.go
```diff
@@ -0,0 +1,143 @@
...
+	for _, plugin := range service.Plugins {
+		var kongPlugin kicv1.KongPlugin
+		kongPlugin.APIVersion = KICAPIVersion
+		kongPlugin.Kind = KongPluginKind
+		if plugin.Name != nil && service.Name != nil {
+			kongPlugin.ObjectMeta.Name = calculateSlug(*service.Name + "-" + *plugin.Name)
+		} else {
+			log.Println("Service name or plugin name is empty. This is not recommended." +
+				"Please, provide a name for the service and the plugin before generating Kong Ingress Controller manifests.")
+		}
```
czeslavo, line 121:
```suggestion
			log.Println("Service name or plugin name is empty. This is not recommended." +
				"Please, provide a name for the service and the plugin before generating Kong Ingress Controller manifests.")
			continue
		}
```

czeslavo (commented):
Thanks Jordi! Left some more about the missing `continue`s where names are missing (objects without names will fail to be created).

battlebyte:
Thanks @czeslavo , missing continues fixed.

czeslavo (approved):
🚀

battlebyte (approved):
Changes after review by Gabriele and Greg.

PR #1051, opened 2023-10-10
feat(analytics): add analytics for local operations

Tieske:
Jira: APIOPS-86


RobSerafini:
@Kong/team-deck - Would appreciate your review on this. It seems safe enough to me and we could always revert if there are problems.

GGabriele (approved):

PR #1056, opened 2023-10-16
feat(patch): allow arrays to be specified on the CLI

Tieske:
requires apiops > 0.1.22 for this pr: https://github.com/Kong/go-apiops/pull/100

values were specified on the CLI as `--value 'key:value'`, and arrays could not. But arrays
were possible in the patch file format. This adds the format `--value '[1,2,3]'` to specify the arrays on the CLI as well.

### DIFF
--- a/cmd/file_patch.go
+++ b/cmd/file_patch.go
@@ -32,7 +32,7 @@ func executePatch(cmd *cobra.Command, args []string) error {
 	{
 		var err error
 		valuesPatch.SelectorSources = cmdPatchSelectors
-		valuesPatch.ObjValues, valuesPatch.Remove, err = patch.ValidateValuesFlags(cmdPatchValues)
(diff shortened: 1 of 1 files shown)


cmd/file_patch.go
```diff
@@ -120,21 +120,30 @@ to the output file. The patches can be specified by a '--selector' and one or mo
...
 
 Examples of valid values:
 
+  # set field "read_timeout" to a numeric value of 10000
 	--selector="$..services[*]" --value="read_timeout:10000"
+
+	# set field "_comment" to a string value
 	--selector="$..services[*]" --value='_comment:"comment injected by patching"'
+
+	# set field "_ignore" to an array of strings
```
rspurgeon, line 138: Instead of saying "set" should this say "append"?

PR #1061, opened 2023-10-25
fix: do not change created_at on entity updates

GGabriele:


file/builder.go
```diff
@@ -188,6 +188,9 @@ func (b *stateBuilder) consumerGroups() {
 				return
 			}
 		}
+		if current != nil && current.CreatedAt != nil {
+			cgo.ConsumerGroup.CreatedAt = current.CreatedAt
+		}
```
aboudreault, line 193: why do we need to check `current.CreatedAt != nil`? if set to nil, kong or konnect will set it properly to now(), no?

GGabriele: Of course 🤦

GGabriele: @aboudreault can you take another look? https://github.com/Kong/deck/pull/1061/commits/8cd9e138161ba442ab14dabbd6177f6088c0b4a5

aboudreault (approved):

PR #1074, opened 2023-11-01
fix(stderr): ensure we output error to stderr, not stdout

Tieske:
This updates the `cprint` package to explicitly specify the output stream.

All occurrences have been updated accordingly, explicit warnings are now going to `stderr`. A few statements remain to stdout.

Is this breaking?
- it changes signatures in the internal helper package `cprint`, is this part of the external API?
- the `Syncer` (see changes in `diff/diff.go`) remains backward compatible, since that is part of the external API.

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -85,7 +85,7 @@ func getWorkspaceName(workspaceFlag string, targetContent *file.Content,
 		if enableJSONOutput {
 			jsonOutput.Warnings = append(jsonOutput.Warnings, warning)
 		} else {
-			cprint.DeletePrintf("Warning: " + warning + "\n")
+			cprint.DeletePrintf(os.Stderr, "Warning: "+warning+"\n")
 		}
 		return workspaceFlag
 	}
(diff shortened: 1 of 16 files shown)


GGabriele:
> is this part of the external API?

Yes, this is used by [KIC](https://github.com/Kong/kubernetes-ingress-controller/blob/main/internal/manager/setup.go#L13)

@Kong/team-k8s

mheap:
I think this would be a breaking change for people that expect warnings on `stdout` for older commands and have scripts that rely on this output.

I'd expect the following behaviour:

* Deprecation notices that were added prior to 1.28 remain on `stdout`
* Deprecation notices added in 1.28 are sent to `stderr`

rainest:
I don't think we care much on the KIC side. We [disable deck logs](https://github.com/Kong/kubernetes-ingress-controller/blob/e933bda55d2447b9a51256b8adc67e9aa7c1fd48/internal/manager/setup.go#L47-L50) above debug and cannot parse them.

I don't think the changes other than those in `diff/diff.go` would affect us, since we don't use the `cmd` package and instead build our own Syncer instance. As of this comment those don't actually appear to change anything (they're still going to stdout with a TODO to change them later). stderr is correct for them (they are informational messages, not output).

The multi-line diff output KIC gets isn't really suitable for log daemon consumption either way, so I'd be fine changing them immediately. It's breaking in the strictest sense, but not really given the practical use case (a human looks at container logs directly for debugging) those messages fit at present.

---

One of the limitations of our current deck interface for KIC is that deck doesn't let you inject a logger, as much of deck doesn't actually use a log library--the cprint stuff is just a basic wrapper around `fmt.Println()` with a very basic notion of verbosity.

The lines KIC ends up showing are logged within the Syncer rather than returned from it, which is one of the things we'd want to change if moving to a more library-oriented design. For now we only have the global cprint on/off toggle.

Tieske:
closing this since it is considered breaking

PR #1084, opened 2023-11-03
fix(cli): validate was setting wrong filenames

Tieske:


CHANGELOG.md
```diff
@@ -79,6 +79,12 @@
 - Add support for konnect AU region.
   [#1082](https://github.com/Kong/deck/pull/1082)
 
+### Fixes
+
+- Fix validate command setting the wrong filelist (and effectively do nothing)
```
rspurgeon, line 84: Changelog entries should describe issues and resolutions from the perspective of a user.  A user won't know what `filelist` is and so it doesn't help them.  This should say something like : 

```suggestion
- Resolved an issue in the `deck file validate` and `deck gateway validate` commands that prevented them from properly processing the provided file arguments.
```

aboudreault (approved):

PR #1087, opened 2023-11-06
add share tag feature

AntoineJac:
add share tag feature: https://konghq.atlassian.net/browse/APIOPS-100


AntoineJac:
Close as handle now by https://github.com/Kong/deck/pull/1097

PR #1101, opened 2023-11-09
tests: fix konnect integration tests

pmalek:


tests/integration/test_utils.go
```diff
@@ -258,7 +261,7 @@ func setup(t *testing.T) {
 
 func sync(kongFile string, opts ...string) error {
 	deckCmd := cmd.NewRootCmd()
-	args := []string{"sync", "-s", kongFile}
+	args := []string{"gateway", "sync", kongFile}
```
GGabriele, line 264: We are keeping the "old" syntax on purpose here since we care more about not breaking existing workflows than introducing possible bugs into new functionalities. @Tieske has some WIP task to improve our posture here.

pmalek:
These tests seem rather brittle anyway, probably due to the nature of using an external service like Konnect. 🤔 Might need more work to make them bullet proof.

hbagdi:

PR #1103, opened 2023-11-09
ci: use license action

pmalek:
Use https://github.com/Kong/kong-license/ instead of manually updated secret. 🤖 

Also while at it, change the order of steps in tests to utilize Go cache. Without this change repo is not checked out when go is being set up:

```
Restore cache failed: Dependencies file is not found in /home/runner/work/deck/deck. Supported file pattern: go.sum
```

### DIFF
--- a/.github/workflows/integration-enterprise.yaml
+++ b/.github/workflows/integration-enterprise.yaml
@@ -28,22 +28,29 @@ jobs:
     env:
       KONG_ANONYMOUS_REPORTS: "off"
       KONG_IMAGE: ${{ matrix.kong_image }}
-      KONG_LICENSE_DATA: ${{ secrets.KONG_LICENSE_DATA }}
 
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
       - name: Setup go
         uses: actions/setup-go@v4
         with:
           go-version: '^1.20'
-      - name: Checkout repository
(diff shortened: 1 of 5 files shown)


GGabriele (approved):
huge!

.github/workflows/integration-enterprise.yaml
```diff
@@ -28,22 +28,29 @@ jobs:
...
         with:
           username: ${{secrets.DOCKERHUB_PULL_USERNAME}}
           password: ${{secrets.DOCKERHUB_PULL_TOKEN}}
+      - uses: Kong/kong-license@master
+        id: license
+        with:
+          password: ${{ secrets.PULP_PASSWORD }}
       - name: Setup Kong
+        env:
+          KONG_LICENSE_DATA: ${{ steps.license.outputs.license }}
```
Tieske, line 51: see https://github.com/Kong/kong-license#github-action

This shouldn't be necessary to set the ENV explicitly, the action will already have it set (exported). There are multiple occurrences of this.

Unless the explicitness of the current approach is preferred, then this is fine.

pmalek: I forgot about this feature of license action but yes:

> Unless the explicitness of the current approach is preferred, then this is fine.

IMHO this might make it easier for the readers that do not know this hidden (albeit documented!) behavior of this action.

PR #1109, opened 2023-11-13
chore: extract code to new library

rainest:
Following the extraction of [v1.29.2](https://github.com/Kong/deck/releases/tag/v1.29.2) library code to a separate repository in https://github.com/Kong/go-database-reconciler/pull/9, update deck module references to use that code's new home.

Remove copies of that code from the deck repo and clean up modules not used by the remaining packages in `kong/deck`.

As-is this PR relies on a commit hash module version for the new repo. We should merge https://github.com/Kong/go-database-reconciler/pull/9 first, tag an actual version, and replace the temporary module version with that.

Removes codegen-related make targets, scripts, and CI configuration, as that does not interact with the remaining code here. I'm not sure how we were using the `kong_json_schema.json` output from one of those. There don't seem to be references here. Is there anything external we need to update with that having moved, or is `kong_json_schema.json` actually internal to the `file` package and Go code that imports it?

Golang and package upgrades appear to have broken several tests. This includes additional changes to address those:

The `fatih/color` module [made changes](https://github.com/fatih/color/commit/a2983f4b633d1e9926360624a7b25886a43114ca) in v1.16.0 that had the (maybe) unexpected side effect of removing spaces between arguments. AFAICT the insertion of `c.Sprint()` means that it calls [`doPrint()` instead of `doPrintln()`](https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/fmt/print.go;l=1203-1226). The former inserts spaces between non-string arguments only, whereas the latter inserts spaces between all arguments. There's no immediate need to upgrade this, so I just downgraded it back to v1.15.0, the version deck is currently using.

Something (dunno what, possibly the Golang upgrade?) changed how tests see deck output. This change was originally reporting errors due to actual output including color codes, whereas the expected output does not contain color codes. I think something in an older version was stripping those, since they are indeed present in the output and the tests never had anything to strip them before. I added a small MIT-licensed helper lib to just strip them, since I don't think we care about color in those integration tests and writing the codes into the expected strings is kinda annoying.


GGabriele:
> I'm not sure how we were using the `kong_json_schema.json` output from one of those. 

This file pulled [here](https://github.com/Kong/deck/blob/main/file/schema.go#L6) and used as part of the [input validation](https://github.com/Kong/deck/blob/v1.29.2/file/validate.go#L29) logic.

> There's no immediate need to upgrade this, so I just downgraded it back to v1.15.0, the version deck is currently using.

Good call, thanks!

> Something (dunno what, possibly the Golang upgrade?) changed how tests see deck output. This change was originally reporting errors due to actual output including color codes, whereas the expected output does not contain color codes. I think something in an older version was stripping those, since they are indeed present in the output and the tests never had anything to strip them before. I added a small MIT-licensed helper lib to just strip them, since I don't think we care about color in those integration tests and writing the codes into the expected strings is kinda annoying.

SGTM!

GGabriele (dismissed):

rainest:
Okay, so the schema JSON should be part of the lib then AFAICT. Bumped to an actual go-database-reconciler tag, sorted out some module weirdness, should be good to go.

GGabriele (dismissed):

GGabriele (approved):

bastantoine:
Hello, I just found out about this MR. The file `kong_json_schema.json` is referenced in the [documentation of the DB-less and declarative config](https://docs.konghq.com/gateway/latest/production/deployment-topologies/db-less-and-declarative-config/) (under *Declarative configuration format*, search for *"See the declarative configuration schema for all configuration options"*).

Could you update the documentation ([here](https://github.com/Kong/docs.konghq.com/blob/main/app/_src/gateway/production/deployment-topologies/db-less-and-declarative-config.md?plain=1#L162)) to either remove the link, or bring back the reference file somewhere and update its location in the docs ?
Do you required an issue about this?

Fun fact, I was using this file yesterday, so imagine my surprise when I didn't found it this morning 😅.

Thank you!

PR #1115, opened 2023-11-19
chore(cli): improve warnings

Tieske:
2 commits;

- describe changes when switching to new commands
- diplay message when reading from stdin

EDIT: 2nd commit moved here: https://github.com/Kong/go-database-reconciler/pull/17

### DIFF
--- a/cmd/file_convert.go
+++ b/cmd/file_convert.go
@@ -6,7 +6,6 @@ import (
 	"strings"
 
 	"github.com/kong/deck/convert"
-	"github.com/kong/go-database-reconciler/pkg/cprint"
 	"github.com/kong/go-database-reconciler/pkg/file"
 	"github.com/kong/go-database-reconciler/pkg/utils"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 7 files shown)


mheap:
Based on feedback so far I think we should keep `-s` as the default, even in the new aliases. It should be a 1:1 implementation for the existing command (with positional arguments in place so that we don't break BC for people that already migrated).

With that in mind, this PR provides inaccurate guidance and shouldn't be merged.

rspurgeon:
As discussed, let's make `-s/--state` and positional arguments for the `gateway` commands both supported and I think we should change the documentation to say something like the following that indicates why a user may want to switch from the `-s` to positional:

```
Use 'deck gateway diff' instead.
Note:
  - see 'deck gateway diff --help' for changes to the command\n"+
  - The new command supports the '-s/--state' flag for backward compatibility but also supports specifying state files as positional arguments which allows for alternative shell expansions (`*.yaml`).
```

mheap:
We shouldn’t provide two ways to do the same thing long term. If positional arguments are the future, then the PR is ok, but I’d like @hbagdi to sign off. We chose flags for a reason in the original design

rspurgeon:
My feedback is driven to help a user understand the differences in capabilities of the two options, which it seems we have established are staying in the tool (`-s` to maintain BC and positional to maintain BC with early adopters).  IIUC the two options provide different capabilities on the shell, so we should document that.

rspurgeon:
Here is the change that needs to occur with the current deprecation messages for all commands that have been marked deprecated by this change.

Currently the messages read as follows:
```
Warning: 'deck sync' is DEPRECATED and will be removed in a future version. Use 'deck gateway sync' instead.
```
These will change to:
```
Info: 'deck sync' functionality has moved to `deck gateway sync` and will be removed in a future MAJOR version of deck. Migration to `deck gateway sync` is recommended.
    Note:
        - see 'deck gateway sync --help' for changes to the command
	- files changed to positional arguments without the '-s/--state' flag
	- the default changed from 'kong.yaml' to '-' (stdin/stdout)
```

rspurgeon (dismissed):
lgtm

RobSerafini:
@mikefero  @mflendrich - I think this just needs an approval and merge and a new build of Deck. Can you guys help move this along?

cmd/file_convert.go
```diff
@@ -91,12 +90,15 @@ func newConvertCmd(deprecated bool) *cobra.Command {
...
 	fileOutDefault := "-"
 	if deprecated {
-		short = "[deprecated] use 'file convert' instead"
+		short = "[deprecated] see 'deck file convert --help' for changes to the command"
 		execute = func(cmd *cobra.Command, args []string) error {
 			convertCmdInputFile = convertCmdInputFileDeprecated
 			convertCmdOutputFile = convertCmdOutputFileDeprecated
-			cprint.UpdatePrintf("Warning: 'deck convert' is DEPRECATED and will be removed in a future version. " +
-				"Use 'deck file convert' instead.\n")
+			fmt.Fprintf(os.Stderr, "Info: 'deck convert' functionality has moved 'deck file convert' and will be removed\n"+
```
GGabriele, line 97:
```suggestion
			fmt.Fprintf(os.Stderr, "Info: 'deck convert' functionality has moved to 'deck file convert' and will be removed\n"+
```

GGabriele (approved):

PR #1117, opened 2023-11-22
Support for Gateway API HTTPRoute objects.

battlebyte:
Added support for the HTTPRoute object according to the Gateway API.


rspurgeon:
@battlebyte please convert to Draft for now so we can evaluate UX and behavior

Tieske:
this work has been cherry-picked into #1050 

closing this.

PR #1206, opened 2024-02-05
feat: auto-generate rla's namespace on 'convert' subcommand

GGabriele:


mheap:
@GGabriele Thank you! Are there any other plugins that you know of that could benefit from the same conversion?

GGabriele:
The only other EE Plugins with `auto = true` fields are the following, but none of these fields are required, so the plugin creation would be successful even without them:
- `graphql-rate-limiting-advanced` (namespace)
- `kafka-log` (cluster_name)
- `kafka-upstream` (cluster_name)
- `openid-connect` (cache_tokens_salt)

mheap:
Does the standard `rate-limiting` plugin have the same required field (and/or can you explain how I can check myself so that I can save you some time?)

Tieske (approved):

PR #1230, opened 2024-02-29
Handle nil error when port not defined in service with Gateway API.

battlebyte:
Fixes 3 issues in kong2kic:

1. Handles nil pointer error when service port is not specified and output style is Gateway API
2. Handles the case in which no HTTP method is specified in a Route
3. Add annotation from service to upstream only when relationship is found.

### DIFF
--- a/kong2kic/route.go
+++ b/kong2kic/route.go
@@ -356,12 +356,21 @@ func populateKICIngressesWithGatewayAPI(content *file.Content, kicContent *KICCo
 			})
 
 			// add service details to HTTPBackendRef
-			portNumber := k8sgwapiv1.PortNumber(*service.Port)
-			backendRef := k8sgwapiv1.BackendRef{
-				BackendObjectReference: k8sgwapiv1.BackendObjectReference{
-					Name: k8sgwapiv1.ObjectName(*service.Name),
-					Port: &portNumber,
-				},
+			var backendRef k8sgwapiv1.BackendRef
+			if service.Port != nil {
+				portNumber := k8sgwapiv1.PortNumber(*service.Port)
+				backendRef = k8sgwapiv1.BackendRef{
+					BackendObjectReference: k8sgwapiv1.BackendObjectReference{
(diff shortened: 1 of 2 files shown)


Tieske:
@czeslavo would you mind having a look at this, lgtm, but would like a k8s/kic double-check

kong2kic/route.go
```diff
@@ -356,12 +356,21 @@ func populateKICIngressesWithGatewayAPI(content *file.Content, kicContent *KICCo
...
+						Port: &portNumber,
+					},
+				}
+			} else {
+				backendRef = k8sgwapiv1.BackendRef{
+					BackendObjectReference: k8sgwapiv1.BackendObjectReference{
+						Name: k8sgwapiv1.ObjectName(*service.Name),
+					},
+				}
 			}
```
czeslavo, line 374: Just to simplify:
```suggestion
			backendRef := k8sgwapiv1.BackendRef{
				BackendObjectReference: k8sgwapiv1.BackendObjectReference{
					Name: k8sgwapiv1.ObjectName(*service.Name),
				},
			}
			if service.Port != nil {
				portNumber := k8sgwapiv1.PortNumber(*service.Port)
				backendRef.Port = &portNumber
			}
```

czeslavo (commented):
Apart from one cosmetic nit, LGTM 👍

kong2kic/route.go
battlebyte: Thanks @czeslavo. Should line 366 be `backendRef.BackendObjectReference.Port = &portNumber`? Strangely enough, your code compiles, although I don't see any `Port` field in the `BackendRef` struct.

czeslavo: It would be equivalent - it's because `BackendObjectReference` is embedded in `BackendRef` structure (https://gobyexample.com/struct-embedding).

battlebyte: True! Suggestion applied in commit e5c9ed1

Tieske (approved):

PR #1235, opened 2024-03-04
tests: add integration test for kong2kic

battlebyte:
- Granular tests based on each resource type
- Integration test deploys generated manifests to k8s cluster
- Refactor code to add annotations for tags
- Fix consumer group plugin bug
- Inform plugin attributes: enabled, runon, ordering, protocols
- Generate external service when host is informed and there is no upstream match


rainest (commented):
Is there a command to regenerate output? `kong2kic/testdata/README.md` doesn't mention one and I don't see an existing `make` target for it. It's unclear how you should update expected output if input or transform code changes. If the answer is "by hand", it shouldn't be.

I don't much love golden output tests for purposes other than showing what a diff is, and they're especially iffy if presented in isolation and without comment on what a particular input/output test is supposed to demonstrate. I'd prefer having more focused units around "field X in input transforms to fields Y+Z in output as expected" as a base, with golden full input/output comparison as backup material to highlight any differences we may not have accounted for in the units.

Golden-style tests end up being particularly challenging for tools that expect to support multiple versions of adjacent software, where we end up with things like [this sync test change](https://github.com/Kong/go-database-reconciler/commit/599ea6575353588a26fab808ac011504d96653ef#diff-b09c72ddce91eda1770990ad2744a3867eb59e006e6842fc4d01b8d3ce5eec77) that are essentially 98% the same as some other version's test case, but require significant input/output duplication because the surrounding checks aren't flexible enough to handle the 2% that is different between versions.

hbagdi:

battlebyte:
@hbagdi I will implement some of the suggestions shortly.

mheap (approved):

PR #1245, opened 2024-03-13
ci(tests): test against go `stable` & `oldstable`

gszr:
* Change the tests and codeql-analysis workflows to run against both Go `stable` and `oldstable`. (If we do not want to test against both, please let me know.)
* Change remaining workflows to run against `stable` - rather than hardcoding

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -34,6 +34,7 @@ jobs:
       fail-fast: false
       matrix:
         language: [ 'go' ]
+        golang: ["oldstable", "stable"]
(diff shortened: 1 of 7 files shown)


Prashansa-K:
@gszr Could you please resolve the conflicts that are being shown here?

I had added some changes here: https://github.com/Kong/deck/pull/1336 to address [this issue](https://github.com/Kong/deck/issues/1282). So, in some places we may have to keep the version same as go.mod so that the released version and project version golang match.

gszr:
I think we can close this one out; still, it could be valuable to test against both stable and oldstable, but I am not sure how to achieve that if we pin the Go version to the one in go.mod.

PR #1312, opened 2024-05-30
added vacuum custom function loader and command line arg to file lint

srb3:


cmd/file_lint.go
```diff
@@ -110,11 +113,16 @@ func executeLint(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to read input file '%s'; %w", cmdLintInputFilename, err)
 	}
 
+	functionsFlag, _ := cmd.Flags().GetString("functions")
```
rspurgeon, line 116: `functions` can be a constant

cmd/file_lint.go
```diff
@@ -220,6 +247,8 @@ func newLintCmd() *cobra.Command {
 	lintCmd.Flags().BoolVarP(&cmdLintOnlyFailures,
 		"display-only-failures", "D", false,
 		"only output results equal to or greater than --fail-severity")
+	lintCmd.Flags().StringVarP(&cmdCustomFunctionsPath, "functions", "f", "",
```
rspurgeon, line 250: I would prefer a descriptive argument name. `--custom-function-path` and then a short option, possibly `-f` for this command.

cmd/file_lint.go
```diff
@@ -20,6 +22,7 @@ var (
 	cmdLintFailSeverity   string
 	cmdLintOutputFilename string
 	cmdLintOnlyFailures   bool
+	cmdCustomFunctionsPath string
```
Tieske, line 25: please use `cmdLint` as prefix, those variables are global in decK, so must minimize probability of colliding with other command options.

cmd/file_lint.go
Tieske, line 116: No need to create this `functionsFlag` variable. When defining the flag like this below:

`lintCmd.Flags().StringVarP(&cmdCustomFunctionsPath, ...`

you pass the address of the global variable `cmdCustomFunctionsPath`, so that is where the value of the flag will be stored. So instead of creating a new variable `functionsFlag` you can directly use the value of `cmdCustomFunctionsPath`.

cmd/file_lint.go
```diff
@@ -184,6 +192,25 @@ func executeLint(cmd *cobra.Command, args []string) error {
...
+) (map[string]model.RuleFunction, error) {
+	// check custom functions
+	if functionsFlag != "" {
+		pm, err := plugin.LoadFunctions(functionsFlag, silence)
+		if err != nil {
+			fmt.Printf("Error: Unable to open custom functions: %v\n", err)
+			fmt.Println()
+			return nil, err
+		}
+		fmt.Printf("Info: Loaded %d custom function(s) successfully.\n", pm.LoadedFunctionCount())
```
Tieske, line 208: do not directly print. Use `logbasics.Info` instead (output is then controlled by the `verbose` cli flag).

cmd/file_lint.go
```diff
@@ -184,6 +192,25 @@ func executeLint(cmd *cobra.Command, args []string) error {
...
+	functionsFlag string,
+	silence bool,
+) (map[string]model.RuleFunction, error) {
+	// check custom functions
+	if functionsFlag != "" {
+		pm, err := plugin.LoadFunctions(functionsFlag, silence)
+		if err != nil {
+			fmt.Printf("Error: Unable to open custom functions: %v\n", err)
+			fmt.Println()
+			return nil, err
```
Tieske, line 206: do not print, return an error instead:

```suggestion
			return fmt.Errorf("unable to open custom functions: %w", err)
```

Tieske (changes requested):

srb3:
Thank you for your feedback! I've addressed the requested changes in the latest commit:

   * Removed the unnecessary functionsFlag variable and directly used cmdCustomFunctionsPath.
   * Updated error handling
   * Switched to logbasics instead of printing directly
   * Simplified the flag initialisation and retrieval.

Please review the changes in [this commit](https://github.com/Kong/deck/pull/1312/commits/1e3719291664ae506da33e5ca70302303a3667c0). Let me know if there are any additional adjustments needed.

cmd/file_lint.go
rspurgeon: lower case vars are not exported outside the package, however this is visible across the `cmd` package so I agree we should follow the suggested and current prefix pattern.

srb3: Thanks @rspurgeon this should have been addressed in the latest commit

rspurgeon:
@srb3 please mark this PR Draft until we have time to discuss this as a product feature, align with documentation, and other items.  Thank you

hbagdi:

PR #1328, opened 2024-06-28
ci: use internal worker for release job

GGabriele:


Tieske (dismissed):
if we update the timeout, should we maybe set it very high first to ensure the workflow works? we can then later scale it back once we know what a proper run-time would be for that flow.

PR #1335, opened 2024-07-08
feat: support online validation with Konnect

GGabriele:


validate/validate.go
```diff
@@ -181,6 +181,9 @@ func (v *Validator) Validate(formatVersion semver.Version) []error {
 	if err := v.entities(v.state.FilterChains, "filter_chains"); err != nil {
 		allErr = append(allErr, err...)
 	}
+	if err := v.entities(v.state.Vaults, "vaults"); err != nil {
```
Prashansa-K, line 184: Sorry, I lack some context here. How is this addition related to this PR? @GGabriele

cmd/gateway_validate.go
```diff
@@ -45,7 +42,11 @@ func executeValidate(cmd *cobra.Command, _ []string) error {
 	ctx := cmd.Context()
 	var kongClient *kong.Client
 	if validateOnline {
-		kongClient, err = getKongClient(ctx, targetContent)
+		if validateCmdRBACResourcesOnly && mode == modeKonnect {
+			return fmt.Errorf("rbac validation not yet supported in konnect mode")
```
Prashansa-K, line 46: @GGabriele Added this condition here, as you mentioned that rbac stuff is not supported by Konnect. Let me know if error needs to be rephrased. Without this, users would get random errors/outputs when the flag is present.

Prashansa-K:
Other e2e tests for deck validate commands are added here: https://github.com/Kong/deck/pull/1415

Kept this PR scoped to Konnect stuff.

Prashansa-K:
@pmalek Could you please help us with a review here? We have extended `deck gateway validate` command to work with Konnect as well. Changelog will be added during release time, as is the practice we follow here with deck releases currently.

pmalek (commented):
I've left some minor comments but to me this looks pretty solid 👍

validate/validate.go
pmalek: ping: @GGabriele. Do you mind chiming in here? 🙇

tests/integration/test_utils.go
```diff
@@ -374,3 +374,23 @@ func render(opts ...string) (string, error) {
 
 	return stripansi.Strip(string(out)), cmdErr
 }
+
+func validate(online bool, opts ...string) error {
```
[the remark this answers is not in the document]
Prashansa-K: I have added two consts ONLINE and OFFLINE instead of passing boolean values. I think that will make the intent clear as well.

validate/validate.go
GGabriele: Vaults validation is currently completely missing from decK. This is not strictly related to the `konnect` implementation, so we can move it into a separate commit if we want.

pmalek: I'd prefer to extract it if that's the case.

Additionally, if possible I'd suggest to mark this with a comment why this is needed.

Prashansa-K: I have created a separate PR for this @pmalek 
https://github.com/Kong/deck/pull/1421

mheap (approved):

Issue #1282, opened 2024-04-22
Please update go version to 1.21.9 or 1.22.2


akoshok1:
This is for CVE : https://github.com/advisories/GHSA-4v7x-pqxf-cx7m
This is for decK 1.37.

Prashansa-K:
@akoshok1 @mheap 
The latest deck version (1.39) already uses `1.22.4`.
If we do a patch release for 1.37 with this update, would we need to do the same for 1.38?

Also, prior versions (1.36 and lower) have 1.21 as well. Do we still support these or are these deprecated / EOL? How many simultaneous versions do we support at a time?

mheap:
@Prashansa-K It looks like the release job uses 1.20 (see https://github.com/Kong/deck/blob/main/.github/workflows/release.yaml#L20)

We should update the jobs to read from `go.mod` and release a new version. Asking people to upgrade to latest is acceptable for deck

Prashansa-K:
This is fixed in 1.39.3

PR #1336, opened 2024-07-16
#1282 Updated git workflows to use go.mod for go versions

Prashansa-K:
This would fix issue #1282.
Fetching the version directly from go.mod would help us to maintain consistency in the project and the released versions.


GGabriele (approved):
Thanks for this fix!

Issue #1313, opened 2024-06-03
decK colours do not work correctly in non-tty environment


calumrussell:
decK uses the fatih/color library for colouring command output. The behaviour of this library is to check whether the terminal is a tty. If it is then do not show colours unless the user provides an override.

This is demonstrated here -  https://pkg.go.dev/github.com/fatih/color#readme-disable-enable-color - the library can write colours in non-tty envs but must be explicitly overwritten.

The bug in decK is that if you are in a non-tty env then you can never display colours. Passing `--no-color` does not override because if you pass `false`, as the docs explain, then it is interpreted as `true`.

The bug is here:

https://github.com/Kong/deck/blob/526baa380db173161fe113a2580a2dca7f472155/cmd/root.go#L376

If you are in a non-tty environment, color.NoColor is always `true` so it is impossible to override this value or change it in anyway. I overrode this value to always be `false` locally, and I was able to get colours in a non-tty environment (by piping the result of my decK command to `cat`).

The expected behaviour is either that decK supports some kind of override or that by passing `--no-color=false` it would be possible to override in a non-tty env.

PR #1339, opened 2024-07-17
Fixed --no-color flag behaviour in non-tty environments

Prashansa-K:
This fixes issue #1313 and makes the flag behaviour consistent across tty and non-tty.

Screenshot 2024-07-17 at 11 11 02 AM

The attached image shows the changed behaviour.
Now, the non-tty output also comes up as colored by default, same as tty output. This can be overridden by passing the `--no-color` flag.

I have a doubt though.
If our customers pass the deck output to a non-tty (may be a file), this can cause things to break if `--no-color` is not passed, as the colorised output may cause formatting issues in a file.
Is this doubt valid?
Should I check for non-tty environments separately and handle this, if this is indeed a valid usecase?
@GGabriele @mheap

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -373,7 +373,9 @@ func initConfig() {
 	// cookie-jar support
 	rootConfig.CookieJarPath = viper.GetString("kong-cookie-jar-path")
 
-	color.NoColor = (color.NoColor || viper.GetBool("no-color"))
+	if viper.IsSet("no-color") {
+		color.NoColor = viper.GetBool("no-color")
+	}
 
 	if err := initKonnectConfig(); err != nil {
 		fmt.Println(err)


GGabriele:
> If our customers pass the deck output to a non-tty (may be a file), this can cause things to break if --no-color is not passed, as the colorised output may cause formatting issues in a file.
Is this doubt valid?

I'd worried only if this commit changes the default behaviour of the tool. Is it the case?

> Should I check for non-tty environments separately and handle this, if this is indeed a valid usecase?

I'd let @mheap chime in, but I believe predictability is a very important feature. I'd vote against doing special handling on a case-by-case way since we should let the input define the expected behaviour.

===

On a side note: please check [here](https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#commit-message-format) for the convention we use for commit messages. This will help us have a consistent and clear "git story".

Thanks!

mheap:
This does change the default behaviour. The expected behaviour from the issue is as follows:

* Default: TTY gets colour, non-TTY doesn't
* `--no-color=true`: Both TTY and non-TTY don't get colour
* `--no-color=false`: Both TTY and non-TTY get colour

There may not be a way to differentiate `--no-color=false` from the default.If so, how would you recommend explicitly enabling colours without changing the default behaviour?

Prashansa-K:
For retaining the default behaviour, I can check if the flag was passed or not before changing the value for `color.NoColor.`
Submitted the new change with commit message as per conventions.

GGabriele (approved):
I didn't know about `viper.IsSet`, nice!!

Good job!

Issue #1341, opened 2024-07-18
Require better validation for patch files for deck file patch command


Prashansa-K:
Right now, we don't validate if the file has selectors and values. This can result into garbled output files.
A validation check can be added for patch files.

PR #1342, opened 2024-07-18
fix: validate presence of flags and args in file patch command

### LINKED ISSUE #1130: Selector CLI flag on file patch and file add-plugin not working when using a patch/plugin YAML file
I am trying to use 2 of the new commands, `file patch` and `file add-plugin` and have tried using the `--selector` flag but its being ignored

I am using a file with the info for the patch and plugin but then the selector via the CLI

`
cat kong.yaml | deck file patch --selector="$..services[*]" patch.yaml
`

If I put that selector into the patch.yaml file then it works, can I not use CLI flag --selector with a patch file?

Prashansa-K:
The command intends to patch input files either via selector-value flags or patch file arguments. The change ensures that at least one of these is present, but not both at the same time. Till now, the command accepted patches from flags and patch files at the same time, but ignored the patches passed via flags. This was a confusing behaviour for customers.

Fix #1130

@rspurgeon I have created a new issue for patch file validation [here](https://github.com/Kong/deck/issues/1341).
We can discuss its feasibility and usecase there. For this particular issue (#1130), I felt erroring out with a proper reasoning given to user should be the best way to move forward.

### DIFF
--- a/cmd/file_patch.go
+++ b/cmd/file_patch.go
@@ -167,6 +167,18 @@ If the 'values' object instead is an array, then any arrays returned by the sele
 will get the 'values' appended to them.
 `,
 		RunE: executePatch,
+		PersistentPreRunE: func(_ *cobra.Command, args []string) error {
+			if len(args) > 0 && (len(cmdPatchSelectors) > 0 || len(cmdPatchValues) > 0) {
+				return fmt.Errorf("cannot use patch file argument along with '--selector' and '--value'")
+			}
+
+			if len(args) == 0 && len(cmdPatchSelectors) == 0 && len(cmdPatchValues) == 0 {
+				return fmt.Errorf("must specify at least one of these: " +
+					"a patch file argument or a '--selector' and '--value' combination")
+			}
+
+			return nil
+		},
 		Example: "# update the read-timeout on all services\n" +
 			"cat kong.yml | deck file patch --selector=\"$..services[*]\" --value=\"read_timeout:10000\"",
 	}


Prashansa-K:
Here's the command in action:
Screenshot 2024-07-18 at 10 46 17 AM

cmd/file_patch.go
```diff
@@ -167,6 +167,18 @@ If the 'values' object instead is an array, then any arrays returned by the sele
 will get the 'values' appended to them.
 `,
 		RunE: executePatch,
+		PersistentPreRunE: func(_ *cobra.Command, args []string) error {
```
GGabriele, line 170: jfyi: this all could also be achieve with the following

```
patchCmd.MarkFlagsOneRequired("selector", "value")
patchCmd.MarkFlagsMutuallyExclusive("selector", "value")
```

```
$ ./deck file patch --selector foo --value foo
Error: if any flags in the group [selector value] are set none of the others can be; [selector value] were all set
```

```
$ ./deck file patch
Error: at least one of the flags in the group [selector value] is required
```

Anyway, I find the custom messages better than the default ones from cobra

GGabriele (commented):
Changes look good to me!

Nit: can you please make sure commit messages follow [the convention](https://github.com/Kong/kong/blob/master/CONTRIBUTING.md#commit-message-format) even for body length? e.g. `The body of your message should not contain lines longer than 72 characters`

This makes everything more readable.

For example:

```
commit a0388610ea9dc259e14c205e85811aac8bd7ed10
Author: Gabriele Gerbino <gabrielegerbino@gmail.com>
Date:   Fri Mar 15 12:57:02 2024 +0100

    fix: do not fetch kong version when running `validate`

    decK used to fetch the running Kong version in order to gate
    route regex patterns validation. This is problematic in
    the case users don't have access to the `/` endpoint, which
    is the one used to fetch the version of the runtime.

    This commit removes the need of fetching the Kong version
    entirely by relying on the `_format_version` field from the
    config file instead.
```

vs.

```
commit 47ab479635509980b1ba7ca6b42e5e28bc443789
Author: Prashansa Kulshrestha <prashkulshrestha@gmail.com>
Date:   Thu Jul 18 10:47:17 2024 +0530

    fix: validate presence of flags and args in file patch command

    The command intends to patch input files either via selector-value flags or patch file arguments. The change ensures that at least one of these is present, but not both at the same time. Till now, the command accepted patches from flags and patch files at the same time, but igno
red the patches passed via flags. This was a confusing behaviour for customers.

    Fix #1130
```

cmd/file_patch.go
Prashansa-K: These methods are limited to flags. They won't be able to compare flags with command arguments.
In the PreRun func, we are comparing selector-value flags with patch file argument.

Prashansa-K:
Updated the commit message.

cmd/file_patch.go
GGabriele: Ah, I misread the issue. Ignore me :D

GGabriele (approved):

PR #1375, opened 2024-08-22
fix: clarify validate konnect behaviors

johnharris85:
Few quick fixes around Konnect behavior:
- The help description for `gateway validate` vs `file validate` was incorrect (said both would contact a running CP to validate, `file validate` does not do this, `gateway validate` does).
- Message about the `konnect` section being missing has been changed to a warning (as you can specify konnect cli flags still, and it now returns a 0 exit code for automation purposes) and the message has been clarified.
- `--konnect-compatibility` and `--parallelism` flags gated to their relevant parent commands (`file validate` and `gateway validate` respectively).

### DIFF
--- a/cmd/gateway_validate.go
+++ b/cmd/gateway_validate.go
@@ -173,7 +173,7 @@ this command unless --online flag is used.
 			return preRunSilenceEventsFlag()
 		}
 
-		if validateOnline {
+		if online {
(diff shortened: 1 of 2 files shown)


Prashansa-K:
@johnharris85 Could you please check why the tests are not running for this particular PR?

Issue #1364, opened 2024-08-21
Handling Konnect API rate limiting


pmalek:
### Problem statement

Konnect API imposes rate limiting on users. CI in this repo uses a Konnect token to run integration tests and sometimes does get 429 in response due to hitting quota:

```
=== RUN   Test_Sync_ConsumerGroupConsumerFromUpstream
    sync_test.go:4992: 
        	Error Trace:	/home/runner/work/deck/deck/tests/integration/sync_test.go:4992
        	Error:      	Received unexpected error:
        	            	authenticating with Konnect: HTTP status 429 (message: "API rate limit exceeded")
        	Test:       	Test_Sync_ConsumerGroupConsumerFromUpstream
Info: 'deck reset' functionality has moved to 'deck gateway reset' and will be removed
in a future MAJOR version of deck. Migration to 'deck gateway reset' is recommended.
Error: authenticating with Konnect: HTTP status 429 (message: "API rate limit exceeded")
    test_utils.go:266: authenticating with Konnect: HTTP status 429 (message: "API rate limit exceeded")%!(EXTRA string=failed to reset Kong's state)
```

https://github.com/Kong/deck/actions/runs/10488966058/job/29052543949

### Acceptance criteria

- [ ] Handle Konnect API throttling in the client or in tests so that hitting the quota do not result in failed tests.

PR #1381, opened 2024-08-27
fix: creating a retryable kong client for tests and retry logic in konnect authentication for CI rate-limit errors

Prashansa-K:
- chore: creating a retryable kong client for tests to account for rate-limit errors
- fix: added retry and backoff logic in Konnect authentication layer to account for rate-limit errors during authentication.

Fixes [#1364](https://github.com/Kong/deck/issues/1364)

### DIFF
--- a/cmd/common_konnect.go
+++ b/cmd/common_konnect.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/kong/go-database-reconciler/pkg/diff"
 	"github.com/kong/go-database-reconciler/pkg/dump"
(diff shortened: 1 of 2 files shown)


GGabriele:
Don't we already have _retries_ supported in our client? https://github.com/Kong/go-database-reconciler/blob/25ccc04e6c42e1b447070391e3938913deaed9fb/pkg/utils/types.go#L182

Prashansa-K:
> Don't we already have _retries_ supported in our client? https://github.com/Kong/go-database-reconciler/blob/25ccc04e6c42e1b447070391e3938913deaed9fb/pkg/utils/types.go#L182

Yeah the client does have it, but the authentication layer still returns 429. The Konnect's login_service uses a normal http request [here](https://github.com/Kong/go-database-reconciler/blob/25ccc04e6c42e1b447070391e3938913deaed9fb/pkg/konnect/login_service.go#L55) and is not retried. Does the client retry it by default?
I added my own retry layer but it still fails with 429. Trying with an increased backoff period.

Prashansa-K:
Works fine with an increased backoff period.
I tried the [pipeline](https://github.com/Kong/deck/actions/runs/10676866469/usage?pr=1381) run thrice and it succeeds each time.
Screenshot 2024-09-03 at 10 51 55 AM

GGabriele (approved):

PR #1390, opened 2024-09-08
Replace templated kong2tf with recursive implementation

mheap:
I've replaced the template approach with an implementation that recursively walks input data. This means that `kong2tf` will work with any entity that deck supports. It will even work with new fields automatically without templates needing to be updated.

In addition, I added support for importing every supported entity (we don't support important consumer group members in the provider).

I've tested this by creating one of every entity in the Konnect UI, running `deck dump --konnect-token $KONNECT_TOKEN --konnect-control-plane-name "CP With Everything" --with-id`, then `kong2tf` and the imports all work as expected with `go run main.go file kong2tf -s ~/tfimport-check/kong.yaml --generate-imports-for-control-plane-id "0dea9abf-074a-4988-bdd6-c9ea1ea25d4b" --ignore-credential-changes`


mheap:
I'm going to merge to the branch - wanted to submit a PR to make it clear what changed. Please feel free to review on here even after merge

Issue #1385, opened 2024-08-28
Check if `--konnect-addr` is set


mheap:
Using the command

```
deck ping --konnect-addr https://us.api.konghq.com
```

returns the error message

```
Error: reading Kong version: making HTTP request: Get "http://localhost:8001/": dial tcp [::1]:8001: connect: connection refused
```

In the above use case, the user is trying to ping konnect, however, the error message appears to be responding to the default --kong-addr value which is a self-managed setup ( and a completely different URL )

**Analysis**

The problem in this case is that decK has no way to determine whether the user wanted to ping Konnect or Kong: the https://us.api.konghq.com value passed via the --konnect-addr is the default one decK uses for Konnect, so the command the customer ran is actually equivalent to just deck ping.

**Acceptance Criteria**

- [ ] Detect the presence of `--konnect-addr` using `viper.isSet` and change the error message as needed
- [ ] All commands detect `--konnect-addr` being provided, not just `deck ping`

PR #1398, opened 2024-09-11
fix: default konnect url with konnect addr flag not accounted for.

Prashansa-K:
Previously, if a deck command was issued with --konnect-addr flag but with the default konnectUrl https://us.api.konghq.com, the flag value was getting ignored as deck could not figure out if it was passed by flag or not. This change ensures that if the flag is set, deck commands are issued against Konnect.

Fixes: https://github.com/Kong/deck/issues/1385


pmalek (approved):
Ideally we'd add a test to for that if possible but I'm not sure how realistic it is given that we rely on package level variables and adding tests (in this PR) could require a fair amount of refactoring.

Please add a changelog entry for this.

PR #1407, opened 2024-09-26
chore: release prep for v1.40.3

Prashansa-K:


Prashansa-K:
@pmalek This PR is ready for review. I have removed the conflicts that were coming up earlier.

README.md
```diff
@@ -73,7 +73,7 @@ the GitHub [release page](https://github.com/kong/deck/releases)
 or install by downloading the binary:
 
 ```shell
-$ curl -sL https://github.com/kong/deck/releases/download/v1.40.0/deck_1.40.0_linux_amd64.tar.gz -o deck.tar.gz
```
pmalek, line 76: I believe we should change those links at some point so that they do not need updating on each release (e.g using latest release link and not adding the version information to the binary file names)

pmalek (approved):

README.md
Prashansa-K: Good idea! I also want to update the release process such that `homebrew/deck` doesn't require manual merge each time. I will keep this in mind too. Thanks!

PR #1424, opened 2024-10-22
fix: konnect cli flags getting considered for online validation.

Prashansa-K:
Earlier, if konnect control plane information was passed via cli flags, online validation for Konnect was failing. This change fixes the issue.

### DIFF
--- a/validate/konnect_compatibility.go
+++ b/validate/konnect_compatibility.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strconv"
 
+	"github.com/kong/go-database-reconciler/pkg/dump"
 	"github.com/kong/go-database-reconciler/pkg/file"
 	"github.com/kong/go-kong/kong"
 )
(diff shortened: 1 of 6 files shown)


validate/konnect_compatibility.go
```diff
@@ -39,14 +40,14 @@ func checkPlugin(name *string, config kong.Configuration) error {
...
-func KonnectCompatibility(targetContent *file.Content) []error {
+func KonnectCompatibility(targetContent *file.Content, dumpConfig dump.Config) []error {
 	var errs []error
 
 	if targetContent.Workspace != "" {
 		errs = append(errs, errors.New(errWorkspace))
 	}
 
-	if targetContent.Konnect == nil {
+	if targetContent.Konnect == nil && dumpConfig.KonnectControlPlane == "" {
```
samugi, line 50: is it always going to be `""` or could it also be `nil`?

samugi (approved):
I don't have much background knowledge about the change but it looks good to me. I left a small comment.

validate/konnect_compatibility.go
Prashansa-K: It's a string value. So, always "".

PR #1450, opened 2024-11-22
chore(deps): import kubernetes-configuration instead of KIC

programmer04:
There is new source of truth - [kubernetes-configuration](https://github.com/Kong/kubernetes-configuration)  for all CRDs maintained by team-k8s (KIC and KGO related), so it should be imported and used instead of KIC.

Closes https://github.com/Kong/kubernetes-ingress-controller/issues/6641


czeslavo (commented):
Please consider adding a golangcilint rule that will prevent using KIC types.

.golangci.yml
```diff
@@ -106,3 +106,15 @@ linters-settings:
...
+          recommendations:
+          - github.com/kong/kubernetes-configuration
+      - github.com/kong/gateway-operator:
+          recommendations:
+          - github.com/kong/kubernetes-configuration
+  importas:
+    no-unaliased: true
+    alias:
+    - pkg: github.com/kong/kubernetes-configuration/api/configuration/(v[\w\d]+)
+      alias: kc${1}
```
czeslavo, line 119: For consistency's sake, I suggest using the same alias as we do in KGO.
```suggestion
      alias: configuration${1}
```

programmer04: Ok, I'm changing it, but TBH, this alias `configuration` makes names unnecessary long and unconvenient to use in function signatures, etc. Maybe something to discuss during team-sync...

czeslavo (approved):

PR #1508, opened 2025-01-27
Fix `deck gateway apply`

mheap:
The existing `deck gateway apply` command works when applying a resource for the first time, but error due to duplicate entities when trying to update the resource.

This PR moves the partial apply logic to `go-database-reconciler` (https://github.com/Kong/go-database-reconciler/pull/182) and removes the `isPartialApply` logic from decK. This also removes the license checks that were previously added for Consumer Group fetching as GDR now handles this.

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -521,7 +521,6 @@ func performDiff(ctx context.Context, currentState, targetState *state.KongState
 	dry bool, parallelism int, delay int, client *kong.Client, isKonnect bool,
 	enableJSONOutput bool, applyType ApplyType,
 ) (int, error) {
-
 	shouldSkipDeletes := applyType == ApplyTypePartial
 
 	s, err := diff.NewSyncer(diff.SyncerOpts{
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -521,13 +521,7 @@ func performDiff(ctx context.Context, currentState, targetState *state.KongState
 	dry bool, parallelism int, delay int, client *kong.Client, isKonnect bool,
 	enableJSONOutput bool, applyType ApplyType,
 ) (int, error) {
-	shouldSkipDeletes := false
-	diffApplyType := diff.ApplyTypeFull
-
-	if applyType == ApplyTypePartial {
-		shouldSkipDeletes = true
-		diffApplyType = diff.ApplyTypePartial
-	}
+	shouldSkipDeletes := applyType == ApplyTypePartial
 
 	s, err := diff.NewSyncer(diff.SyncerOpts{
 		CurrentState:  currentState,
@@ -542,7 +536,7 @@ func performDiff(ctx context.Context, currentState, targetState *state.KongState
 		return 0, err
 	}
 
-	stats, errs, changes := s.Solve(ctx, parallelism, dry, enableJSONOutput, diffApplyType)
+	stats, errs, changes := s.Solve(ctx, parallelism, dry, enableJSONOutput)
 	// print stats before error to report completed operations
 	if !enableJSONOutput {
 		printStats(stats)
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -131,7 +131,7 @@ func RemoveConsumerPlugins(targetContentPlugins []file.FPlugin) []file.FPlugin {
 }
 
 func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
-	delay int, workspace string, enableJSONOutput bool,
+	delay int, workspace string, enableJSONOutput bool, isPartialApply bool,
 ) error {
 	// read target file
 	if enableJSONOutput {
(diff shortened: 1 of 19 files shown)


cmd/common_konnect.go
```diff
@@ -246,7 +246,7 @@ func syncKonnect(ctx context.Context,
 		return err
 	}
 
-	stats, errs, _ := s.Solve(ctx, parallelism, dry, false)
+	stats, errs, _ := s.Solve(ctx, parallelism, dry, false, false)
```
pmalek, line 249: Style: the new `bool` parameter could get its own type (e.g. `type ApplyType bool`) and then a set of named constants so that we don't end up with call sites that just use verbatim `true` / `false` which are hard to reason about unless function signature is looked up.

tests/integration/apply_test.go
```diff
@@ -64,4 +78,28 @@ func Test_Apply_3x(t *testing.T) {
...
 		})
 	}
+
+	t.Run("updates existing entities", func(t *testing.T) {
+		runWhen(t, "kong", ">=3.0.0")
+		setup(t)
+
+		apply("testdata/apply/007-update-existing-entity/service-01.yaml")
+
+		out, _ := dump()
```
pmalek, line 88: Why not check the error?

```suggestion
		out, err := dump()
		require.NoError(t, err)
```

tests/integration/apply_test.go
```diff
@@ -64,4 +78,28 @@ func Test_Apply_3x(t *testing.T) {
...
+		expectedOriginal, err := readFile("testdata/apply/007-update-existing-entity/expected-state-01.yaml")
+		if err != nil {
+			t.Fatalf("failed to read expected state: %v", err)
+		}
```
pmalek, line 92:
```suggestion
		expectedOriginal, err := readFile("testdata/apply/007-update-existing-entity/expected-state-01.yaml")
		require.NoError(t, err, "failed to read expected state")
```

tests/integration/apply_test.go
```diff
@@ -64,4 +78,28 @@ func Test_Apply_3x(t *testing.T) {
...
+
+		assert.Equal(t, expectedOriginal, out)
+
+		apply("testdata/apply/007-update-existing-entity/service-02.yaml")
+		expectedChanged, err := readFile("testdata/apply/007-update-existing-entity/expected-state-02.yaml")
+		if err != nil {
+			t.Fatalf("failed to read expected state: %v", err)
+		}
+
+		outChanged, _ := dump()
```
pmalek, line 102:
```suggestion
		outChanged, err := dump()
		require.NoError(t, err)
```

tests/integration/apply_test.go
```diff
@@ -64,4 +78,28 @@ func Test_Apply_3x(t *testing.T) {
...
+			t.Fatalf("failed to read expected state: %v", err)
+		}
+
+		assert.Equal(t, expectedOriginal, out)
+
+		apply("testdata/apply/007-update-existing-entity/service-02.yaml")
+		expectedChanged, err := readFile("testdata/apply/007-update-existing-entity/expected-state-02.yaml")
+		if err != nil {
+			t.Fatalf("failed to read expected state: %v", err)
+		}
```
pmalek, line 100:
```suggestion
		require.NoError(t, err, "failed to read expected state")
```

tests/integration/apply_test.go
pmalek, line 1: I've noticed that `sync()` and `dump()` are missing context plumbing so I've added them in #1509. PTAL

cmd/common.go
```diff
@@ -272,24 +256,21 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 		return err
 	}
 
-	// Consumer groups are an enterprise 3.4+ feature
-	if parsedKongVersion.GTE(reconcilerUtils.Kong340Version) && isLicensedKongEnterprise {
```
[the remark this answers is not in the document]
mheap: This is something that I added in the previous PR that is no longer needed as I've moved this logic in to GDR so consumer groups are only loaded if referenced rather than loading all at runtime

cmd/common_konnect.go
mheap: Agreed. I've pushed an update for this in https://github.com/Kong/deck/pull/1508/commits/152977482599301b8559d3c7f70801bcb5222a3a

tests/integration/apply_test.go
mheap: Merged and this PR is rebased

pmalek (approved):
Please add a [changelog](https://github.com/Kong/deck/blob/main/CHANGELOG.md) entry for this.

👍

mheap:
It looks as though the changelog is added at release time based on https://github.com/Kong/deck/pull/1502

Merging, and can issue a followup PR if needed

PR #1518, opened 2025-02-05
feat: adding consumer-group policy overrides support in deck

Prashansa-K:
Proposal Draft: https://konghq.atlassian.net/wiki/spaces/A/pages/4357751187/Consumer+Group+Policy+Overrides+with+Deck

Related GDR changes: https://github.com/Kong/go-database-reconciler/pull/191

### DIFF
--- a/tests/integration/sync_test.go
+++ b/tests/integration/sync_test.go
@@ -2173,7 +2173,7 @@ func Test_Sync_ServicesRoutes_From_3x(t *testing.T) {
 	}
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
-			runWhenKongOrKonnect(t, ">=3.0.0")
+			runWhen(t, "kong", ">=3.0.0")
 			setup(t)
 
 			require.NoError(t, sync(context.Background(), tc.kongFile))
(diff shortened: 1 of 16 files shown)


Prashansa-K:
@harshadixit12 I have added the fixes to Konnect integration test in this separate PR: https://github.com/Kong/deck/pull/1519
Since we retrieve kong version now (refer: https://github.com/Kong/go-database-reconciler/pull/195), Konnect related tests had to be separated out for testing.

tests/integration/sync_test.go
```diff
@@ -1925,6 +1925,11 @@ const complexQueryForDegraphqlRoute = `query SearchPosts($filters: PostsFilters)
 }
 `
 
+const errorConsumerGroupPolicies = "a rate-limiting-advanced plugin with config.consumer_groups\n" +
```
harshadixit12, line 1928: Should we capitalise the first letter of error message?

harshadixit12:
@Prashansa-K had a few questions - 
1. Why do we have separate tests for gateway versions 3.8 and 3.9?
2. Do we need tests for the diff command?

tests/integration/sync_test.go
Prashansa-K: This error comes from GDR. It is copied as it is. https://github.com/Kong/go-database-reconciler/blob/main/pkg/utils/utils.go#L29
Also, errors in go are not capitalised by convention. You get lint warnings for the same.

Prashansa-K:
> @Prashansa-K had a few questions -
> 
> 1. Why do we have separate tests for gateway versions 3.8 and 3.9?
> 2. Do we need tests for the diff command?

1. 3.9 introduces some schema changes in the RLA plugin. So, different files and expected state need to be maintained for both these versions.
2. Not required. It is a dry-run version of sync and we have covered sync tests.

harshadixit12 (approved):

Issue #1522, opened 2025-02-11
image: kong/deck:latest ( v1.44.0 ) Error 403 command deck gateway diff/sync to Kong Opensource


athapornpong:
Issue: Inconsistent Behavior with deck gateway diff Command Between Local and Job Script
Description:
When I run the deck gateway diff command locally, it works as expected without any errors:

```paintext
deck gateway diff --kong-addr 'https://kong-admin-oss-nonprod' --headers 'Kong-Admin-Token: PASSWORD'  service-tmp.yaml --workspace default


Output:
Summary:
  Created: 0
  Updated: 0
  Deleted: 0
```

However, when using the same command in a CI job script with the Docker image kong/deck:latest, I receive the following error:
```paintext
 docker run --rm -v $(pwd):/deck kong/deck:latest gateway diff --kong-addr "https://kong-admin-oss-nonprod" --headers 'Kong-Admin-Token: PASSWORD' /deck/service-tmp.yaml --tls-skip-verify

Error: custom entity degraphql_routes: HTTP status 403 (message: "Enterprise license missing or expired")
```

Steps to Reproduce:
Environment: Kong OSS setup

Kong Control Plane OSS:
  values:
    image:
      repository: kong/kong-gateway
      tag: "3.4.3.13"
Kong Data Plane OSS:
  values:
    image:
      repository: kong/kong-gateway
      tag: "3.4.3.13"

Job Script: Using kong/deck:latest Docker image
Command 
```
deck gateway sync --kong-addr $KONG_ADMIN_OSS_API_URL --headers "Kong-Admin-Token: $KONG_ADMIN_OSS_TOKEN"  /etc/kong/kong.yml --workspace default
```

Observed Error (in ArgoCD logs):
```
Error: custom entity degraphql_routes: HTTP status 403 (message: "Enterprise license missing or expired")
```

Expected Behavior:
The command should work in both local and CI environments without triggering the Enterprise license error, as we're using Kong OSS.

Actual Behavior:
✅ Works fine when executed locally.
❌ Fails with a 403 error when run inside a CI job using the kong/deck:latest Docker image.

Additional Information:
Kong Version: 3.4.3.13 (OSS)
Deck Version (Docker Image): kong/deck:latest
CI Tool: ArgoCD

🙏 Request:
Please help identify why the behavior is inconsistent between the local environment and the Docker job. Is there a version mismatch or an Enterprise feature accidentally enabled in the Docker image?

Prashansa-K:
[...]

PR #1531, opened 2025-02-13
chore: added new gh workflow for improving OSS gateway testing.

Prashansa-K:
At the moment, we only rely on kong/kong image
for testing for OSS gateway. We have found cases
where users use kong/kong-gateway image, without
a license, to use as an OSS. To deal with these
scenarios and ensure deck works seemlessly with
them, we are adding this new workflow which uses
the EE image but has no license.

### DIFF
--- a/.github/workflows/integration-ee-image-no-license.yaml
+++ b/.github/workflows/integration-ee-image-no-license.yaml
@@ -1,0 +1,57 @@
+name: Integration Test (EE Image, No License)
+
+concurrency:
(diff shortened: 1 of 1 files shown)


Prashansa-K:
Found the need for this from here: https://github.com/Kong/deck/issues/1522

Prashansa-K:
The check for enterprise in tests relies on the presence of string "enterprise" in kong version. So, enterprise tests are bound to fail here.
Thus, closing this PR for now.

PR #1534, opened 2025-02-13
Revert "chore(deps): bump github.com/daveshanley/vacuum from 0.9.15 to 0.16.3"

Prashansa-K:
Reverts Kong/deck#1528


mheap:
@Prashansa-K Please talk to @czeslavo as this is a required change for KIC

czeslavo:
It looks like `daveshanley/vacuum` still has a bug causing _another_ race condition than the one for which I contributed a fix. It happens locally every 3rd/4th test run. In this case we can revert it. I'll look into the upstream code to see if it's easy to fix. If not, I will report a bug. For KIC, if we bump deck to a proper version in GDR, it will be enough.

czeslavo (approved):

Issue #1524, opened 2025-02-11
Add --skip-consumers-with-consumer-groups flag to sync


Prashansa-K:
At the moment, this flag works only for deck dump

PR #1545, opened 2025-02-24
feat: added --skip-consumers-with-consumer-groups to sync and diff commands

Prashansa-K:
This flag was earlier enabled on dump command.
We are adding it to sync and diff commands too
as it provides performance benefits in case an
enterprise has thousands of consumers per group.
It ensures that list_consumers=false while
querying a consumer-group. This is beneficial for
enterprise customers who sync consumers separately on a different cadence 
but require frequent changes in other entity configurations. 
Not listing consumers leads to a boost in performance in such cases.

Related changes:
https://github.com/Kong/go-database-reconciler/pull/159
https://github.com/Kong/go-kong/pull/494

Why was this change necessary?
Check linked FTI: https://konghq.atlassian.net/browse/FTI-6183

More info: https://kongstrong.slack.com/archives/C07AQH7SAF8/p1736409331920239

For #1524

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -26,8 +26,9 @@ import (
 const exitCodeDiffDetection = 2
 
 var (
-	assumeYes    bool
-	noMaskValues bool
+	assumeYes        bool
+	noMaskValues     bool
+	syncCmdAssumeYes bool
 )
 
 type mode int
@@ -449,7 +450,7 @@ func validateSkipConsumersWithConsumerGroups(targetContent file.Content, config 
 		printFn("--skip-consumers-with-consumer-groups flag is added, but no select tags are specified.\n" +
 			"This can lead to unintended changes.\n\n")
 
-		ok, err := reconcilerUtils.Confirm("> Do you wish to continue? ")
+		ok, err := confirmPrompt("> Do you wish to continue? ", syncCmdAssumeYes)
 		if err != nil {
 			return false, err
 		}
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -439,6 +439,14 @@ func validateSkipConsumersWithConsumerGroups(targetContent file.Content, config 
 		printFn := color.New(color.FgYellow, color.Bold).PrintfFunc()
 		printFn("--skip-consumers-with-consumer-groups flag is added, but no select tags are specified.\n" +
 			"This can lead to unintended changes.\n\n")
+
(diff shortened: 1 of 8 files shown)


cmd/common.go
```diff
@@ -178,6 +179,10 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 	var kongClient *kong.Client
 	mode := getMode(targetContent)
 	if mode == modeKonnect {
+		if dumpConfig.SkipConsumersWithConsumerGroups {
+			return errors.New("the flag --skip-consumers-with-consumer-groups can not be used with Konnect")
```
mheap, line 183: Is this because Konnect does not support the query parameter?

Prashansa-K: Yes.

cmd/common.go
```diff
@@ -422,6 +434,24 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 	return nil
 }
 
+func validateSkipConsumersWithConsumerGroups(targetContent file.Content, config dump.Config) error {
+	if len(config.SelectorTags) == 0 {
+		printFn := color.New(color.FgYellow, color.Bold).PrintfFunc()
+		printFn("--skip-consumers-with-consumer-groups flag is added, but no select tags are specified.\n" +
```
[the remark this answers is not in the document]
harshadixit12: Would prompting user to confirm they want to go ahead make sense?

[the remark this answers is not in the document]
Prashansa-K: This isn't an error in "deck's sense". So, I thought printing a warning and adding the necessary description in help command should be sufficient. The error case which can lead to problematic output is when this flag is used along with `consumers.groups` in the file. So, we are erroring out there.

> Would prompting user to confirm they want to go ahead make sense?
That's a good idea. I will add that and push the changes.

cmd/common.go
```diff
@@ -422,6 +435,32 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
...
+			"This can lead to unintended changes.\n\n")
+
+		ok, err := reconcilerUtils.Confirm("> Do you wish to continue? ")
```
harshadixit12, line 444: This would come up on diff command as well even though no writes to gateway are being made, correct?

Prashansa-K: Yes, it would. We keep diff and sync similar in terms of output shown to the end-user.

harshadixit12 (approved):

PR #1555, opened 2025-03-10
fix: process top-level plugins and routes

battlebyte:
Correctly process top level definitions of plugins and routes. Previously, top level definitions of plugins and routes were not processed.

### DIFF
--- a/kong2kic/route.go
+++ b/kong2kic/route.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/kong/go-database-reconciler/pkg/file"
+	"github.com/kong/go-kong/kong"
 	configurationv1 "github.com/kong/kubernetes-configuration/api/configuration/v1"
 	k8snetv1 "k8s.io/api/networking/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
@@ -515,6 +516,12 @@ func processGatewayAPIPlugin(
 	}
 	if plugin.RunOn != nil {
 		kongPlugin.RunOn = *plugin.RunOn
+	}
+	if plugin.Ordering != nil {
+		kongPlugin.Ordering = &kong.PluginOrdering{
+			Before: plugin.Ordering.Before,
+			After:  plugin.Ordering.After,
+		}
 	}
 	if plugin.Protocols != nil {
 		var protocols []string
--- a/kong2kic/route.go
+++ b/kong2kic/route.go
@@ -9,7 +9,6 @@ import (
 	"strings"
 
 	"github.com/kong/go-database-reconciler/pkg/file"
-	"github.com/kong/go-kong/kong"
 	configurationv1 "github.com/kong/kubernetes-configuration/api/configuration/v1"
 	k8snetv1 "k8s.io/api/networking/v1"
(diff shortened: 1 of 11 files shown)


kong2kic/route.go
```diff
@@ -518,43 +460,97 @@ func addBackendRefs(httpRoute *k8sgwapiv1.HTTPRoute, service file.FService, rout
...
+	}
+	kongPlugin.Config = apiextensionsv1.JSON{Raw: configJSON}
+
+	// Populate additional plugin fields if they exist
+	if plugin.Enabled != nil {
+		kongPlugin.Disabled = !*plugin.Enabled
+	}
+	if plugin.RunOn != nil {
+		kongPlugin.RunOn = *plugin.RunOn
+	}
```
Prashansa-K, line 518: Don't we need similar handling for plugin.Ordering?

kong2kic/utils.go
```diff
@@ -145,3 +148,64 @@ func createKongPlugin(plugin *file.FPlugin, ownerName string) (*configurationv1.
...
+}
+
+// Utility function to process top-level entitites in a file.Content.
+// Find the top-level plugins and routes associated with the service and add them to the service-level entities.
+// Find the top-level plugins associated with the routes and add them to the route-level plugins.
+func processTopLevelEntities(content *file.Content) {
+	for _, route := range content.Routes {
+		// Find the service associated with this route
+		for idx, service := range content.Services {
+			if route.Service != nil && route.Service.Name != nil && service.Name != nil && *route.Service.Name == *service.Name {
```
Prashansa-K, line 183: If we shift this condition `if route.Service != nil && route.Service.Name != nil` out of the second loop and `continue` in case it is false, we can avoid going into the second loop entirely for those cases and simplify the condition as well, making it more readable.

```
for _, route := range content.Routes {
   if route.Service == nil && route.Service.Name == nil {
      continue
   }
   
   for idx, service := range content.Services {
      if service.Name != nil && *route.Service.Name == *service.Name {
             ... rest of the code ...
      }
   }
}
```

kong2kic/utils.go
```diff
@@ -145,3 +148,64 @@ func createKongPlugin(plugin *file.FPlugin, ownerName string) (*configurationv1.
...
+				break
+			}
+		}
+	}
+	// Process top-level plugins
+	for _, plugin := range content.Plugins {
+		if plugin.Service != nil {
+			// Find the service associated with this plugin
+			for idx, service := range content.Services {
+				if plugin.Service.ID != nil && service.Name != nil && *plugin.Service.ID == *service.Name {
```
Prashansa-K, line 194: An early check for `if plugin.Service.ID == nil` before the second loop would help here as well.

Prashansa-K: Same for `plugin.Route.ID` on L203

kong2kic/utils.go
battlebyte: Makes sense, but adjusted the first condition to:
if route.Service == nil || (route.Service != nil && route.Service.Name == nil) {
	continue 
}

kong2kic/utils.go
battlebyte: That is right, both modified.

Prashansa-K (approved):

PR #1570, opened 2025-03-26
feat: added support for partials in deck

Prashansa-K:


tests/integration/testdata/sync/039-partials/kong-partials-no-name.yaml
```diff
@@ -0,0 +1,6 @@
+_format_version: "3.0"
```
harshadixit12, line 1: If partial can be created without providing name/id, how does it get linked to a plugin 🤔

harshadixit12 (approved):

tests/integration/testdata/sync/039-partials/kong-partials-no-name.yaml
Prashansa-K: In that case, a user would have to take a gateway dump and use those ids.
Can be useful in distributed config.

Issue #1543, opened 2025-02-19
sync with --select-tag can replace existing certificate without indicating an update / delete if ID is mixed up by mistake


harshadixit12:
**Issue**
We have a customer report where two separate teams were running deck gateway syncs (from separate files) with `--select-tag` and a certificate object. Somehow, they crossed wires and the ID became the same for both files (but different tags)
The console output for both indicates it is "creating" the object, however, it was actually overwriting/updating the original object.

**How to recreate?**
Have two files for sync
On deck file A - Select tag was "TestA" and the ID of certificate was "XYZ"
On deck file B - Select tag was "TestB" and the ID of certificate was "XYZ"
Now, run `deck gateway sync --select-tag {tag} {file}` for each file one after the other

This would be the response for both
creating certificate XYZ
Summary:
  Created: 1
  Updated: 0
  Deleted: 0

Now, run `deck gateway dump` and check the output - the first certificate is replaced.

Indication that the certificate is being replaced is missing.

**Expected behaviour**
When a resource is being updated / replaced, we should indicate on the console.

mheap:
I'd go one step further. If the tag on the entity does not match what is provided in `select_tags`, throw an error. This should be true for all entities, not just certificates

PR #1572, opened 2025-03-26
fix select tag mismatch leading to replacement of entity with ID

harshadixit12:
Fixes https://github.com/Kong/deck/issues/1543

Pending:
- [x] Add tests for konnect

Depends on:
GDR: https://github.com/Kong/go-database-reconciler/pull/214
Go Kong: https://github.com/Kong/go-kong/pull/526

Note: This is to be merged after partials are supported on deck


go.mod
```diff
@@ -11,10 +11,10 @@ require (
 	github.com/blang/semver/v4 v4.0.0
 	github.com/daveshanley/vacuum v0.9.15
 	github.com/fatih/color v1.18.0
-	github.com/google/go-cmp v0.6.0
+	github.com/google/go-cmp v0.7.0
 	github.com/kong/go-apiops v0.1.41
-	github.com/kong/go-database-reconciler v1.20.3
-	github.com/kong/go-kong v0.63.0
+	github.com/kong/go-database-reconciler v1.20.4-0.20250326070856-1248921909f8
```
harshadixit12, line 16: To be updated once go-kong and GDR are released

tests/integration/sync_test.go
```diff
@@ -7917,3 +7917,352 @@ func Test_Sync_Consumers_Default_Lookup_Tag(t *testing.T) {
...
+			require.Error(t, err)
+			assert.ErrorContains(t, err, tc.errorExpected)
+		})
+	}
+}
+
+// test scope:
+//   - konnect
+func Test_Sync_Avoid_Overwrite_On_Select_Tag_Mismatch_With_ID_Konnect(t *testing.T) {
+	runWhenKonnect(t)
```
Prashansa-K, line 8038: Can't we do `runWhenKongorKonnect` for this?

Prashansa-K: I meant we keep just one function with all the test-cases and use `runWhenKongOrKonnect`. 😅 

Any entity that konnect doesn't support can have a diff function. It can reduce a lot of duplication.

harshadixit12: Ah, support for the entities also depends on the kong version - for example, vault is supported from >=v3, partials from v3.10, consumer groups on enterprise, etc.
I could probably reduce some duplication by using `runWhenKongorKonnect` for https://github.com/Kong/deck/pull/1572/files#diff-b09c72ddce91eda1770990ad2744a3867eb59e006e6842fc4d01b8d3ce5eec77R7924, and create a separate suite for testing only vault, consumer group, and partials

Prashansa-K: Yupp

Prashansa-K (approved):

PR #1586, opened 2025-04-04
fix: tests failing due to missing timeout key

harshadixit12:
Some tests for partials seem to be failing in konnect since a new property `timeout` is present in the response. ([link](https://github.com/Kong/deck/actions/runs/14245134517/job/39966070393?pr=1585) to a failing job)
The same is not present in Kong Enterprise GW.

So, I've separated the tests, and included the new property in the konnect tests.

### DIFF
--- a/tests/integration/sync_test.go
+++ b/tests/integration/sync_test.go
@@ -7732,7 +7732,7 @@ func Test_Sync_SkipConsumersWithConsumerGroups_Konnect(t *testing.T) {
 }
 
 func Test_Sync_Partials_Plugins(t *testing.T) {
-	runWhenEnterpriseOrKonnect(t, ">=3.10.0")
+	runWhen(t, "enterprise", ">=3.10.0")
 
 	client, err := getTestClient()
 	require.NoError(t, err)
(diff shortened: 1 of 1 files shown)


tests/integration/sync_test.go
```diff
@@ -7732,7 +7732,114 @@ func Test_Sync_SkipConsumersWithConsumerGroups_Konnect(t *testing.T) {
...
+	t.Run("linking to a plugin fails in case of non-existent partial", func(t *testing.T) {
+		err := sync(ctx, "testdata/sync/039-partials/kong-wrong.yaml")
+		require.Error(t, err)
+		assert.ErrorContains(t, err, "partial non-existent-partial for plugin rate-limiting-advanced: entity not found")
+	})
+}
+
+func Test_Sync_Partials_Plugins_Konnect(t *testing.T) {
+	runWhen(t, "konnect", "")
+	setDefaultKonnectControlPlane(t)
```
harshadixit12, line 7842: calling `setDefaultKonnectControlPlane(t)` inside runWhenKonnect led to other tests failing which use a different control plane.
IMO, `setDefaultKonnectControlPlane(t)` is best called wherever needed, rather than by default for all konnect tests - what do you think? @Prashansa-K

Prashansa-K (approved):

PR #1597, opened 2025-04-10
fix: fixed foreign key reference under plugin config

Prashansa-K:
This change updates gdr version to v1.22.4
Here, we have relaxed the restrictions
added in an earlier deck version to error
out in case of foreign keys under a plugin
config. Now, errors come up only if same
type of entity is nested under an entity.
This change has also fixed the resolution
of foreign keys in case names are used,
instead of IDs.

### DIFF
--- a/tests/integration/testdata/diff/004-no-diff-plugin/protocol-initial-order-plugins.yaml
+++ b/tests/integration/testdata/diff/004-no-diff-plugin/protocol-initial-order-plugins.yaml
@@ -13,7 +13,6 @@ services:
     - grpc
     - https
   - name: prometheus
-    service: mockbin
     tags:
     - o11y
     config:
--- a/tests/integration/testdata/sync/041-plugins-nested-foreign-keys/1_1/kong-consumer-groups-via-ids.yaml
+++ b/tests/integration/testdata/sync/041-plugins-nested-foreign-keys/1_1/kong-consumer-groups-via-ids.yaml
@@ -1,0 +1,44 @@
+_format_version: "1.1"
+
+consumer_groups:
+  - name: group-1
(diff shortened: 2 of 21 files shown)


tests/integration/testdata/diff/004-no-diff-plugin/protocol-initial-order-plugins.yaml
```diff
@@ -13,7 +13,6 @@ services:
     - grpc
     - https
   - name: prometheus
-    service: mockbin
```
harshadixit12, line 16: @Prashansa-K why was this change needed 🤔  - iirc these files were used for plugin protocols ordering fix

Prashansa-K, line 16: 'service' under 'service' is an error now.

harshadixit12, line 16: We've modified this because service is referenced again here for a service scoped plugin?

tests/integration/testdata/sync/041-plugins-nested-foreign-keys/1_1/kong-consumer-groups-via-ids.yaml
```diff
@@ -0,0 +1,44 @@
...
+          limit_by: consumer
+          minute: 6
+          policy: local
+        consumer_group: 8ca63651-4068-4baa-b2b9-08dc99c296e0 # group-1's id
+        enabled: true
+        name: rate-limiting
+        protocols:
+          - http
+
+routes:
```
harshadixit12, line 29: Why do we want multiple routes - each having a plugin referencing different consumer group?

harshadixit12 (approved):

Issue #1554, opened 2025-03-07
deck command for config checking and auto-fixes to help in migrations


Prashansa-K:
This utility can help users to check for lint errors or any deprecated/unsupported config in their deck files and also auto-fix whatever possible. It can be useful for migrations between versions.
Extending `deck file convert` can also be explored here.

Prashansa-K:
Internal doc for tracking what configs we will auto-fix is added here along with a sample output: https://konghq.atlassian.net/wiki/spaces/A/pages/4563304450/Configuration+Reference+for+Migration+Utility+2.8+to+3.4

A test binary is shared with the stakeholders involved along with a sample file. Reviews pending.

PR #1610, opened 2025-04-23
feat: migration utility for converting configuration from 2.8 to 3.4

Prashansa-K:
Internal Reference: https://konghq.atlassian.net/wiki/spaces/A/pages/4476502132/decK+config+migration+utility

For https://github.com/Kong/deck/issues/1554


cmd/file_convert.go
```diff
@@ -114,8 +114,14 @@ can be converted into a 'kong-gateway-3.x' configuration file.`,
 		Args: validateNoArgs,
```
harshadixit12, line 114: The new valid formats should be added to suggestions when `deck file convert --help` is run?

28x.yaml
```diff
@@ -0,0 +1,145 @@
+_format_version: "1.1"
```
harshadixit12, line 1: this and check.yaml - are they used anywhere?

Prashansa-K: Ahh! Bad commits. I will remove them.

convert/convert.go
```diff
@@ -27,6 +28,10 @@ const (
 	FormatKongGateway2x Format = "kong-gateway-2.x"
 	// FormatKongGateway3x represents the Kong gateway 3.x format.
 	FormatKongGateway3x Format = "kong-gateway-3.x"
+
+	// Adding LTS version strings
+	FormatKongGatewayVersion28x Format = "2.8"
```
harshadixit12, line 33: We dont want to prefix these with `kong-gateway-`?

convert/plugin_updates.go
```diff
@@ -94,3 +99,90 @@ func autoGenerateNamespaceForRLAPluginConsumerGroups(plugin *kong.ConsumerGroupP
...
+	oldField, newField, pluginName string,
+) kong.Configuration {
+	var changes int
+	if _, ok := pluginConfig[oldField]; ok {
+		pluginConfig[newField] = pluginConfig[oldField]
+		delete(pluginConfig, oldField)
+		changes++
+	}
+	if changes > 0 {
+		cprint.UpdatePrintf("Automatically converted legacy configuration field \"%s\" to the new field %s in plugin %s\n",
```
harshadixit12, line 172: let's add quotes for the new field as well?

convert/plugin_updates.go
```diff
@@ -94,3 +99,90 @@ func autoGenerateNamespaceForRLAPluginConsumerGroups(plugin *kong.ConsumerGroupP
...
+			})
+		}
+	}
+}
+
+func updateLegacyPluginConfig(plugin *file.FPlugin) {
+	if plugin != nil && plugin.Config != nil {
+		config := plugin.Config.DeepCopy()
+
+		config = updateLegacyFieldToNewField(config, "blacklist", "deny", "")
```
harshadixit12, line 144: We don't want to print plugin name to console here?

convert/rulesets/280-to-340/plugins.yaml
```diff
@@ -0,0 +1,98 @@
+rules:
```
harshadixit12, line 1: @Prashansa-K what would our source of truth be to come up with these rules - ie, how do we know what has changed for each entity from gateway version `a` to `b`?
I can think of `/schemas/{entity}` endpoint on admin API, is there anything else?

convert/plugin_updates.go
```diff
@@ -94,3 +99,90 @@ func autoGenerateNamespaceForRLAPluginConsumerGroups(plugin *kong.ConsumerGroupP
...
+
+		config = updateLegacyFieldToNewField(config, "whitelist", "allow", "")
+
+		if plugin.Name != nil {
```
harshadixit12, line 148: Was pluginName not a required property in 2.8x?

convert/plugin_updates.go
```diff
@@ -94,3 +99,90 @@ func autoGenerateNamespaceForRLAPluginConsumerGroups(plugin *kong.ConsumerGroupP
...
+	}
+	if changes > 0 {
+		cprint.UpdatePrintf("Automatically converted legacy configuration field \"%s\" to the new field %s in plugin %s\n",
+			oldField, newField, pluginName)
+	}
+	return pluginConfig
+}
+
+func removeDeprecatedFields3x(pluginConfig kong.Configuration, fieldName, pluginName string) kong.Configuration {
+	var changes int
```
harshadixit12, line 179: If removing one field at a time, this function can be simplified, no? `changes` and the if condition based on it can be removed?

convert/rulesets/280-to-340/plugins.yaml
Prashansa-K: https://github.com/Kong/kong-lts-migration-preflight/blob/main/deck-rulesets/28-to-34/rules/plugins.yaml
We are using the same rules that the TSE teams use.

convert/plugin_updates.go
Prashansa-K: I would rather have a check here then let my code panic in-case of a schema change.

convert/plugin_updates.go
Prashansa-K: Many plugins require these updates. So, I left it out.
I have added changes so that pluginName gets printed now.

tests/integration/file_convert_test.go
```diff
@@ -72,3 +73,69 @@ func Test_FileConvert(t *testing.T) {
...
+			name:               "auto-fixes plugin configuration for 2.8x",
+			inputFile:          "testdata/file-convert/002-kong-gateway-28x-to-34x-migration/28x-plugins.yaml",
+			errorExpected:      false,
+			expectedOutputFile: "testdata/file-convert/002-kong-gateway-28x-to-34x-migration/34x-expected-plugins.yaml",
+		},
+		{
+			name:               "auto-fixes route configuration for 2.8x",
+			inputFile:          "testdata/file-convert/002-kong-gateway-28x-to-34x-migration/28x-routes.yaml",
+			errorExpected:      false,
+			expectedOutputFile: "testdata/file-convert/002-kong-gateway-28x-to-34x-migration/34x-expected-routes.yaml",
```
harshadixit12, line 103: Lets add consumer and consumer group scoped plugins in the tests too?

harshadixit12:
LGTM, one suggestion for tests

tests/integration/file_convert_test.go
Prashansa-K: The plugins that we are auto-fixing right now aren't supported for consumer-group scoping as checked from here:
https://docs.konghq.com/hub/plugins/compatibility/#scopes

I have added a possible one for consumer-scoping. Also, expanded the tests for route-scoped and service+route scoped plugin.

harshadixit12 (approved):

PR #1622, opened 2025-05-05
tests: Add a simple regression test for sync performance

harshadixit12:
This PR contains a very basic implementation for performance - measuring duration of execution of sync command.
There is code duplication, and it's been implemented to run on only latest enterprise gateway - so needs polishing.

IMO, depending on duration could lead to flaky tests - as network latency is unpredictable.

Ideally, I'd want to have the following in a performance test:
1. No flakiness
2. Key performance indicators - CPU consumption, memory consumption, Network throughput, and execution time, for example
3. Representative state files to measure each - designed to stress each of the key performance indicators
4. Storing past metrics and easy comparison - for example, i'd want the execution time on `main` available to compare with my `feat/slow-feature` as a comment on my PR - so it's easy to compare what is changing.

Feedback and thoughts on how this can be improved welcome 🙏 

The test passes on this branch, I'll link another where it fails - based on v1.46.1

### DIFF
--- a/tests/performance/sync_test.go
+++ b/tests/performance/sync_test.go
@@ -4,9 +4,12 @@ package performance
 
 import (
 	"context"
+	"io"
+	"os"
 	"testing"
-	"time"
 
+	"github.com/acarl005/stripansi"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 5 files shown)


tests/performance/sync_test.go
```diff
@@ -0,0 +1,41 @@
...
+func Test_Sync_Execution_Duration_Simple(t *testing.T) {
+	tests := []struct {
+		name                        string
+		stateFile                   string
+		acceptableExecutionDuration int64
+	}{
+		{
+			name: "Entities with UUIDs",
+			// This file contains 100 services, 10 consumer groups, and 100 consumers in total.
+			// Note that real world latency for http response will be about 10x of local instance (which is used in testing)
```
harshadixit12, line 22: I want to avoid unnecessary consumption of github actions usage by automated tests.

harshadixit12:
While still relevant, perhaps not necessary in its current form.

PR #1629, opened 2025-05-12
chore: updated golang to v1.24.3

Prashansa-K:
This change would also ensure
that we account for CVEs
https://nvd.nist.gov/vuln/detail/cve-2025-22871
and https://nvd.nist.gov/vuln/detail/CVE-2025-22866

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.24.0 AS build
+FROM golang:1.24.3 AS build
 WORKDIR /deck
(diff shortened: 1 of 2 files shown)


Prashansa-K:
According to the CVE reports here: https://www.cve.org/CVERecord?id=CVE-2025-22871
and https://www.cve.org/CVERecord?id=CVE-2025-22866, this go version 1.24.3 should target both the CVEs.

harshadixit12 (approved):

PR #1630, opened 2025-05-12
fix: consumer group consumer create request body

harshadixit12:


Prashansa-K:
This PR wouldn't be required anymore, right? The module update is already merged and afair we don't require new tests for this. Feel free to close this if that's the case.

PR #1662, opened 2025-06-16
fix: added dump config option to skip custom entities if select tags …

Prashansa-K:
…are set

### DIFF
--- a/tests/integration/sync_test.go
+++ b/tests/integration/sync_test.go
@@ -8909,8 +8909,9 @@ func Test_Sync_Partials_Tagging_Konnect(t *testing.T) {
 // test scope:
 //
 // - >=3.0.0 enterprise
+// - konnect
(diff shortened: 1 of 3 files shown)


tests/integration/sync_test.go
```diff
@@ -8905,3 +8905,39 @@ func Test_Sync_Partials_Tagging_Konnect(t *testing.T) {
 		})
 	}
 }
+
+// test scope:
+//
+// - >=3.0.0 enterprise
+func Test_Sync_SkipCustomEntitiesWithSelectorTags(t *testing.T) {
+	runWhen(t, "enterprise", ">=3.0.0")
```
harshadixit12: We don't need this for konnect?

Prashansa-K: will add it with the mod update

Issue #1668, opened 2025-06-19
file validation fails with vague error while using lookup tags


Prashansa-K:
"default_lookup_tags" can't work with file (offline) validation as it involves getting a dumpState from the kong gateway. Right now, this fails with 
`Error: building state: ....`

It's better to give users a proper message regarding this.

PR #1669, opened 2025-06-19
fix: file validation with lookup tags

Prashansa-K:
For #1668

### DIFF
--- a/tests/integration/validate_test.go
+++ b/tests/integration/validate_test.go
@@ -325,9 +325,7 @@ func Test_Validate_PartialLookupTags(t *testing.T) {
 			err := validate(tc.mode, validateOpts...)
 			if tc.errorExpected {
 				require.Error(t, err)
-				if tc.errorString != "" {
-					assert.Contains(t, err.Error(), tc.errorString)
-				}
+				assert.Contains(t, err.Error(), tc.errorString)
 				return
 			}
 			require.NoError(t, err)


tests/integration/validate_test.go
```diff
@@ -304,7 +322,14 @@ func Test_Validate_PartialLookupTags(t *testing.T) {
 			}
 			validateOpts = append(validateOpts, tc.additionalArgs...)
 
-			err := validate(ONLINE, validateOpts...)
+			err := validate(tc.mode, validateOpts...)
+			if tc.errorExpected {
+				require.Error(t, err)
+				if tc.errorString != "" {
```
harshadixit12: Is this if condition required? 👀

Prashansa-K: I had added it to avoid panics for future, but since this is a test, we can let it go. Updated the code.

tests/integration/validate_test.go
```diff
@@ -282,16 +282,34 @@ func Test_Validate_PartialLookupTags(t *testing.T) {
...
 		},
 		{
 			name:      "validate default_lookup_tags with partials",
 			stateFile: "testdata/validate/001-partials/partial-lookup-tags.yaml",
+			mode:      ONLINE,
+		},
+		{
+			name:           "validate partials",
+			stateFile:      "testdata/validate/001-partials/partials.yaml",
+			additionalArgs: []string{"--online-entities-list=Partials"},
```
harshadixit12, line 303: Is the arg relevant for offline validation - since we are anyway printing an error message

harshadixit12: Approving the PR, this can be fixed on main PR if needed.

harshadixit12 (approved):

PR #1675, opened 2025-06-25
fix: keys keysets creation without names

Prashansa-K:
For https://github.com/Kong/go-database-reconciler/issues/300

### DIFF
--- a/tests/integration/sync_test.go
+++ b/tests/integration/sync_test.go
@@ -8403,6 +8403,27 @@ func Test_Sync_KeysAndKeySets(t *testing.T) {
 				},
 			},
 		},
+		{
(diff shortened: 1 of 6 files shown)


tests/integration/sync_test.go
```diff
@@ -8489,6 +8589,83 @@ func Test_Sync_KeysAndKeySets_Konnect(t *testing.T) {
 	}
 }
 
+// test scope:
+//
+// - konnect
+// TODO: Update test once konnect supports key_set creation without names.
```
Prashansa-K, line 8595: Ref: https://kongstrong.slack.com/archives/C04RXLGNB6K/p1750749600325499

PR #1677, opened 2025-06-26
chore: remove unused error handling code

asartori86:
The `err` variable was not re-computed, ensuring that it is guaranteed to be non-nil at this point in the code.


asartori86:
superseeded by #1685

PR #1686, opened 2025-06-30
feat: add konnect-control-plane-id flag

asartori86:
Introduce the --konnect-control-plane-id flag, which is mutually exclusive with the --konnect-control-plane-name flag. While names are more user-friendly, IDs are better suited for automation scripts due to their simplicity (no spaces, special characters, or sensitive names). This flag is intended for internal use and is therefore hidden.

requires https://github.com/Kong/go-database-reconciler/pull/303


cmd/common.go
```diff
@@ -127,6 +127,21 @@ func evaluateTargetRuntimeGroupOrControlPlaneName(targetContent *file.Content) e
 	return nil
 }
 
+func evaluateControlPlaneID(targetContent *file.Content) error {
+	idFromFile := targetContent.Konnect.ControlPlaneID
+	if idFromFile != "" && konnectControlPlaneID != "" && idFromFile != konnectControlPlaneID {
+		return fmt.Errorf("warning: control plane ID '%v' specified via "+
```
harshadixit12: let's not have `warning: ` in the error string - it is leading to 
```
Error: warning: control plane ID '41fafd27-1c5f-4954-9c39-4d4df68e16bc' specified via --konnect-control-plane-id flag is different from '41fafd27-1c5f-4954-9c39-4d4df68e16bd' found in state file(s)
```

harshadixit12:
`gateway_validate` command seems to fail when id is given
```
Error: [konnect] section not specified - ensure details are set via cli flags
```

asartori86:
> `gateway_validate` command seems to fail when id is given
> 
> ```
> Error: [konnect] section not specified - ensure details are set via cli flags
> ```
> 
> Also, some tests wherever relevant would be great!

could you share the full command line used?

this works for me

```shell
$ ./deck gateway validate --konnect-addr https://us.api.konghq.tech --konnect-token $KONNECT_TOKEN_DEV  test.json 

$ cat test.json
{ "_format_version": "3.0"
, "_konnect": {"control_plane_id": "c4fe2fd4-c3e1-494e-b1ea-40a73415c456"}
}
```

cmd/common.go
asartori86: sure, I've adopted the same formatting used few lines above. Amended both

PR #1689, opened 2025-07-04
Fix: lint invalid ruleset panic

harshadixit12:
Todo:
Verify that the version bump does not cause breaking changes


mheap:
I need to drop this for a while. The next steps are to upgrade the tests to respect the new default rules for OAS

harshadixit12:
> I need to drop this for a while. The next steps are to upgrade the tests to respect the new default rules for OAS

Update on this - the output from linting doesn't seem to be deterministic anymore on v0.17, and there are issues with concurrency due to which tests are failing. Doesn't seem to be a quick fix.

mheap:
Closing in favour of https://github.com/Kong/deck/pull/1748

PR #1701, opened 2025-07-23
chore: restructured dump command code

Prashansa-K:
This change reduces duplication in the dump command code and does some basic restructuring.
This would be useful when we add sanitisation capabilities.


Prashansa-K:
Changed the base branch to a copy of main for now. @harshadixit12 
If the changes look okay, I would like to merge it to base.

PR #1709, opened 2025-07-30
feat: added basic sanitizer struct for file.Content sanitizing

Prashansa-K:


sanitize/exempts.go
```diff
@@ -0,0 +1,10 @@
+package sanitize
+
+// Top-level content fields that should not be sanitized
+var topLevelExemptedFields = map[string]struct{}{
+	"Info":          {},
```
Prashansa-K, line 5: Info sanitisation would be done at last to retain referential integrity.
https://konghq.atlassian.net/browse/APIOPS-128

harshadixit12: What will we be sanitizing in `info`? 👀

harshadixit12 (approved):

sanitize/exempts.go
Prashansa-K: lookup tags and select-tags information.

PR #1711, opened 2025-07-30
feat: added basic sanitiser for fields

Prashansa-K:


sanitize/exempts.go
```diff
@@ -0,0 +1,31 @@
...
+	"FormatVersion": {},
+	"Konnect":       {},
+	"Transform":     {},
+	"Workspace":     {},
+}
+
+// Fields that should be skipped for specific entities
+// These fields either require special handling (like Keys)
+// or need not be sanitized like plugin.name
+var entityLevelExemptedFields = map[string]map[string]struct{}{
```
harshadixit12, line 15: Would it be better if we make it clear in code which fields need special handling and which are exempted?

sanitize/sanitize.go
```diff
@@ -0,0 +1,143 @@
...
+	contentType := content.Type()
+
+	for i := 0; i < content.NumField(); i++ {
+		fieldName := contentType.Field(i).Name
+		if _, exempt := topLevelExemptedFields[fieldName]; exempt {
+			continue
+		}
+		fieldValueSet := content.Field(i)
+
+		if fieldValueSet.IsValid() && fieldValueSet.CanInterface() && !fieldValueSet.IsZero() {
```
harshadixit12: Why do we want to check for CanInterface() here?

harshadixit12: Also, wouldn't https://pkg.go.dev/reflect#Value.IsValid already cover checking for zero value?

sanitize/utils.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (s *Sanitizer) sanitizeValue(value string) string {
+	hashedPath := s.hashValue(value)
+	if !strings.Contains(value, "/") {
+		return hashedPath
+	}
+
+	var redactedPath string
+
+	// this means it is a path field
```
harshadixit12: Why do we want special handling for path fields? Could there be any other fields like this which need special handling?

sanitize/sanitize.go
```diff
@@ -0,0 +1,143 @@
...
+		sanitizedValue, exists := s.sanitizedMap[originalValue]
+		if !exists {
+			sanitizedValue = s.sanitizeValue(originalValue)
+			s.sanitizedMap[originalValue] = sanitizedValue
+		}
+
+		if field.CanSet() {
+			field.SetString(sanitizedValue.(string))
+		} else {
+			fmt.Println("Cannot set sanitized value for field:", field.Type(), "with value:", originalValue)
```
harshadixit12, line 135: Do we want this to be communicated to user?

sanitize/sanitize.go
```diff
@@ -0,0 +1,143 @@
...
+
+		if fieldValueSet.IsValid() && fieldValueSet.CanInterface() && !fieldValueSet.IsZero() {
+			s.sanitizeField(fieldValueSet)
+		}
+	}
+
+	return s.content, nil
+}
+
+func (s *Sanitizer) sanitizeField(field reflect.Value) {
```
harshadixit12, line 71: IMO this might benefit from unit tests - since there is branching and this might undergo more iteration

sanitize/exempts.go
Prashansa-K: In the sanitisation code, I will add the special handling. Separate functions for each entityType.
Those entities still need to be exempted to ensure the sanitisation doesn't overwrite the special handling done.

sanitize/sanitize.go
Prashansa-K: Yes! So, that they can omit the field if necessary before sharing the dump file with Kong or anyone else for the matter.

sanitize/utils.go
Prashansa-K: Gateway expects `/` or `~/` in the beginning of paths. Similarly for content-types, it expects a `\` in between.
Thus, it needs special handling. Otherwise, the validation and sync would fail

sanitize/sanitize.go
Prashansa-K: > Why do we want to check for CanInterface() here?

Later in the code, if we wish to use a value via Interface(), we need to ensure it beforehand to avoid panics later. Also, with this check, code wouldn't attempt to sanitize unexported fields, if any, present in any struct. This would avoid runtime panics.

> Also, wouldn't https://pkg.go.dev/reflect#Value.IsValid already cover checking for zero value?

These are different checks - it's possible for a field to be valid (initialized properly) but still contain its zero value. The code wants to skip both invalid fields (which would cause panics if accessed) and zero-valued fields (which may not need sanitization).

sanitize/sanitize.go
Prashansa-K: I will add some.

harshadixit12 (approved):

PR #1718, opened 2025-08-08
feat: added sanitization for kong Configuration 

Prashansa-K:
This feature recursively checks for
configuration inside a plugin, partial, etc
and sanitizes it.
Added unit tests along the sanitizer function

### DIFF
--- a/sanitize/sanitize.go
+++ b/sanitize/sanitize.go
@@ -187,10 +187,6 @@ func (s *Sanitizer) sanitizeConfig(config reflect.Value) interface{} {
 	switch config.Kind() {
 	case reflect.Invalid:
 		return nil
-	case reflect.Ptr:
-		if config.IsNil() {
-			return nil
-		}
 	case reflect.String:
 		configValue := config.String()
 		if configValue == "" {
@@ -230,7 +226,9 @@ func (s *Sanitizer) sanitizeConfig(config reflect.Value) interface{} {
 			}
 		}
 	default:
-		return config.Interface()
+		if config.CanInterface() {
+			return config.Interface()
+		}
 	}
 
 	return sanitizedConfig


sanitize/sanitize.go
```diff
@@ -174,3 +179,57 @@ func (s *Sanitizer) hashValue(value string) string {
...
 }
+
+func (s *Sanitizer) sanitizeConfig(config reflect.Value) interface{} {
+	sanitizedConfig := make(map[string]interface{})
+
+	//nolint:exhaustive
+	switch config.Kind() {
+	case reflect.Invalid:
+		return nil
+	case reflect.String:
```
harshadixit12, line 190: can the repetition in sanitizeField and here be avoided?

sanitize/sanitize.go
```diff
@@ -98,10 +98,15 @@ func (s *Sanitizer) sanitizeField(field reflect.Value) {
 			}
 
 			// needs special handling for configs as they are not pointers to structs
-			// commented for now to satisfy linter
-			// if fieldValue.Type() == reflect.TypeOf(kong.Configuration{}) {
-			// 	// handle configs - tba
-			// }
+			if fieldValue.Type() == reflect.TypeOf(kong.Configuration{}) {
```
harshadixit12, line 101: What other structs might we need special handling for?

Prashansa-K: We will have to mock contents for keys, certs, ca_certs. So, those would require special handling as well.

sanitize/sanitize.go
Prashansa-K: The entire block has certain repetition. The reason I had to tackle kong.Configuration{} separately is because it is not added as a pointer. 
Eg; If you look at Plugin struct, it is defined as follows:
type Plugin struct {
	ID            *string         `json:"id,omitempty" yaml:"id,omitempty"`
	Name          *string         `json:"name,omitempty" yaml:"name,omitempty"`
	....
	ConsumerGroup *ConsumerGroup  `json:"consumer_group,omitempty" yaml:"consumer_group,omitempty"`
	Config        Configuration   `json:"config,omitempty" yaml:"config,omitempty"`
	....
}

Every other field is a pointer to a struct and hence directly referenceable. So, we can change the fields directly in-place. With Config, it is passed as a normal struct. So, we have to replace it. That's what I do in the main sanitizeField() func.
If I combine these two functions, it can be a bit confusing, specially for handling config maps. I would have to unnecessary manage some returns in the recursion just for kong.Configuration. Other structs don't need it, That's why, I kept it separate. It increases some duplication but reduces overall complexity and confusion.

harshadixit12 (approved):

PR #1719, opened 2025-08-12
feat: pulling in entity schemas to check for exemptions to generate. 

Prashansa-K:
- This includes fields that maybe marked as `one_of` or `enum`
- Added unit test for main utility function
- (pending) integration test


tests/integration/dump_test.go
```diff
@@ -603,3 +608,65 @@ func Test_Dump_KeysAndKeySets_Konnect(t *testing.T) {
 		})
 	}
 }
+
+// test scope:
+//
+// - >=3.10.0
+func Test_Dump_Deterministic_Sanitizer(t *testing.T) {
+	runWhen(t, "kong", ">=3.10.0")
```
Prashansa-K: I will expand this in subsequent PRs.
Limited this for now as partials are being tested

tests/integration/testdata/dump/008-sanitizer/expected.yaml
```diff
@@ -0,0 +1,84 @@
+_format_version: "3.0"
```
harshadixit12, line 1: Entities like certificates, etc is still pending to be added for sanitization?

tests/integration/testdata/dump/008-sanitizer/expected.yaml
```diff
@@ -0,0 +1,84 @@
...
+consumer_groups:
+- id: 0a1451d7-4ba5-4a72-bd83-3ab775192393
+  name: bd9a838f7686d3a0e36d47f65a51602404dea10dfa827508b3eb43275d2309e0
```
harshadixit12, line 4: Could the length of hashed string cause http 400 from API when we re-sync due to limitation on length? I'm guessing it is not a priority, since any re-sync will be done by us and we can work around it.

tests/integration/dump_test.go
```diff
@@ -603,3 +608,65 @@ func Test_Dump_KeysAndKeySets_Konnect(t *testing.T) {
...
+			require.NoError(t, err)
+
+			sanitizer := sanitize.NewSanitizer(&sanitize.SanitizerOptions{
+				Ctx:     ctx,
+				Client:  client,
+				Content: fileContent.DeepCopy(),
+				Salt:    "deterministic-test-salt-12345",
+			})
+
+			sanitizedOutput, err := sanitizer.Sanitize()
```
harshadixit12, line 649: Why not `dump` with `--sanitize` flag - is it because the salt would be different and we'd end up getting different output?

Prashansa-K: Yes. In subsequent PRs, I will add a hidden flag to pass a salt. I didn't want to add in this one to not lengthen it.

tests/integration/testdata/dump/008-sanitizer/expected.yaml
Prashansa-K: Not a priority for now.

tests/integration/testdata/dump/008-sanitizer/expected.yaml
Prashansa-K: Yes. It will come up in subsequent PRs and I will add dedicated tests for them.

sanitize/utils.go
```diff
@@ -12,5 +29,90 @@ func shouldSkipSanitization(fieldName string, exemptionMap map[string]struct{})
 		return true
 	}
 
+	// checking for schema-generated exemptions per entity-type
```
harshadixit12, line 38: This function seems to accept exemptionMap as argument, but accesses entityLevelExemptedFieldsFromSchema directly - since entity name is now an argument, we can work without passing exemptionMap, no?

Prashansa-K: Yeah, we don't need it anymore as we are passing entityName now. We can directly fetch the map using that.
I will change this.

sanitize/exempts.go
```diff
@@ -29,3 +30,6 @@ var entityLevelExemptedFields = map[string]map[string]struct{}{
 var configLevelExemptedFields = map[string]struct{}{
 	"ID": {},
 }
+
+// dynamically generated maps of exempted fields from schemas
+var entityLevelExemptedFieldsFromSchema = map[string]map[string]bool{}
```
harshadixit12, line 35: I see there are nil checks on this field - but is it explicitly getting marked as nil anywhere?
it initialises to a map, so

sanitize/utils.go
```diff
@@ -12,5 +26,87 @@ func shouldSkipSanitization(fieldName string, exemptionMap map[string]struct{})
...
+			if _, exempt := exemptedFields[schemaField]; exempt {
+				return true
+			}
+		}
+	}
+
 	return false
 }
+
+func buildExemptedFieldsFromSchema(entityName string, entitySchema kong.Schema) {
```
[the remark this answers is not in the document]
harshadixit12: As of now, this function isn't called concurrently though, correct?
(can keep the mutex, as avoids any issues in future)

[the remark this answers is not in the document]
Prashansa-K: I think default parallelism is 10 which can run 10 go-routines at once. So, I would keep it to avoid issues.

sanitize/exempts.go
Prashansa-K: Not marking it nil right now anywhere but it is a safety check and doesn't have any perf issues anyway. In future, if we mark it as nil, the check can help to avoid panics.

sanitize/sanitize.go
```diff
@@ -84,22 +98,30 @@ func (s *Sanitizer) sanitizeField(field reflect.Value) {
 		t := field.Type()
 		entityName := t.Name()
 
-		entitySkipFields, hasEntitySkips := entityLevelExemptedFields[entityName]
+		// specificEntityName is used to identify plugin or partial types
+		// it is useful for building exempted fields
+		specificEntityName, entitySchema, err := s.fetchEntitySchema(entityName, field)
+		if err != nil {
+			warningMessage := fmt.Sprintf("Error fetching schema for entity: %s %v\n"+
```
harshadixit12, line 105: in case of error, do we also want to inform user they they may want to manually redact any sensitive info in this entity?

sanitize/schemas.go
```diff
@@ -0,0 +1,128 @@
...
+	req, err := s.client.NewRequest(http.MethodGet, endpoint, nil, nil)
+	if err != nil {
+		return schema, err
+	}
+	resp, err := s.client.Do(s.ctx, req, &schema)
+	if resp == nil {
+		return schema, fmt.Errorf("invalid HTTP response: %w", err)
+	}
+	if err != nil {
+		return schema, fmt.Errorf("failed to fetch schema: %w", err)
```
harshadixit12, line 124: why not return nil for schema when there is error?

Prashansa-K: It would be nil by default right? It is uninitialised.

sanitize/sanitize.go
Prashansa-K: The warning I have added informs them that some sanitization features may not work as expected. I guess that should be enough and they can modify the generated info if anything is sensitive or invalid.

sanitize/schemas.go
harshadixit12: ah yes ✅

PR #1726, opened 2025-08-16
feat: special handling added for certs and keys

Prashansa-K:
- Creating mock certs and keys for sanitization.
- This is done so that the generated result is valid and syncable with the gateway.

### DIFF
--- a/sanitize/custom_entity_specific.go
+++ b/sanitize/custom_entity_specific.go
@@ -39,7 +39,7 @@ func (s *Sanitizer) handleEntity(entityName string, fieldValue reflect.Value) er
 	case Key:
 		return s.handleKey(fieldValue)
 	default:
-		return fmt.Errorf("no specific handler for entity: %s\n", entityName)
+		return fmt.Errorf("no specific handler for entity: %s", entityName)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


sanitize/custom_entity_specific.go
```diff
@@ -0,0 +1,236 @@
...
+	// Self-signing the certificate
+	derBytes, err := x509.CreateCertificate(rand.Reader, &template, &template, &priv.PublicKey, priv)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	// PEM encoding the certificate and key
+	certPEMBytes := string(pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: derBytes}))
+	keyPEMBytes := string(pem.EncodeToMemory(&pem.Block{
+		Type:  "RSA PRIVATE KEY",
```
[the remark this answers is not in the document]
Prashansa-K: We are using PKCS#1

sanitize/exempts.go
```diff
@@ -35,5 +35,11 @@ var configLevelExemptedFields = map[string]struct{}{
 	"dictionary_name": {}, // present in rla, upstream_oauth
 }
 
+var entitiesToHandleDifferently = map[string]struct{}{
+	"CACertificate": {},
+	"FCertificate":  {},
```
harshadixit12, line 40: Should this be `Certificate` instead of `FCertificate`🤔

Prashansa-K: No. GDR defines certs as a slice FCertificate. Refer: https://github.com/Kong/go-database-reconciler/blob/main/pkg/file/types.go#L1144

PR #1728, opened 2025-08-18
feat: added vault sanitisation

Prashansa-K:


sanitize/sanitize.go
```diff
@@ -79,6 +79,8 @@ func (s *Sanitizer) Sanitize() (*file.Content, error) {
 		}
 	}
 
+	s.sanitizeInfo(s.content.Info)
```
Prashansa-K, line 82: Merged info sanitisation PR to this.

sanitize/sanitize.go
```diff
@@ -264,3 +276,23 @@ func (s *Sanitizer) sanitizeConfig(entityName string, config reflect.Value) inte
 
 	return sanitizedConfig
 }
+
+func (s *Sanitizer) sanitizeInfo(info *file.Info) {
```
Prashansa-K, line 280: From info sanitisation PR

tests/integration/testdata/dump/008-sanitizer/select-tags.expected.yaml
```diff
@@ -0,0 +1,12 @@
+_format_version: "3.0"
```
Prashansa-K, line 1: From info sanitisation PR.

tests/integration/dump_test.go
```diff
@@ -705,6 +707,13 @@ func Test_Dump_Sanitize(t *testing.T) {
 			expectedFile: "testdata/dump/008-sanitizer/consumergroup-plugins36.expected.yaml",
 			runWhen:      func(t *testing.T) { runWhen(t, "enterprise", ">=3.6.0") },
 		},
+		{
+			name:         "dump sanitize with select-tags",
```
Prashansa-K, line 711: From info sanitisation PR.

tests/integration/dump_test.go
```diff
@@ -761,6 +776,33 @@ func Test_Dump_Sanitize_Special_Entities(t *testing.T) {
 			stateFile: "testdata/dump/008-sanitizer/ca-cert.yaml",
 			runWhen:   func(t *testing.T) { runWhen(t, "kong", ">=2.8.0") },
 		},
+		{
+			name:           "dump sanitize env vault and vault references",
+			stateFile:      "testdata/dump/008-sanitizer/env-vault.yaml",
```
harshadixit12, line 781: This and the next testcase seem to use same state file - can be merged?

tests/integration/dump_test.go
```diff
@@ -761,6 +776,33 @@ func Test_Dump_Sanitize_Special_Entities(t *testing.T) {
...
+			skipValidation: true, // env vault validation endpoint not available in 3.4.0
+		},
+		{
+			name:      "dump sanitize env vault and vault references",
+			stateFile: "testdata/dump/008-sanitizer/env-vault.yaml",
+			runWhen:   func(t *testing.T) { runWhen(t, "enterprise", ">=3.5.0") },
+		},
+		{
+			name:           "dump sanitize vault config",
+			stateFile:      "testdata/dump/008-sanitizer/vault-configs.yaml",
```
harshadixit12, line 792: Same for this and next

sanitize/sanitize.go
```diff
@@ -243,6 +251,10 @@ func (s *Sanitizer) sanitizeConfig(entityName string, config reflect.Value) inte
 			switch v := val.Interface().(type) {
 			case string:
 				sanitizedVal := s.sanitizeValue(v)
+				if requiresURISanitization(k) {
+					sanitizedVal = hashURI(v, s.salt)
```
harshadixit12, line 255: Can you give some examples of where we hash uris?

harshadixit12: This is for fields inside vault.config?

tests/integration/dump_test.go
Prashansa-K: This one skips validation command as 3.4 gateway didn't have any vault validation endpoint.

tests/integration/dump_test.go
Prashansa-K: Same as above. The differentiator is skipValidation param.

sanitize/utils.go
```diff
@@ -122,3 +125,32 @@ func findRelevantFieldNamesWithKey(exemptedFieldsFromSchema map[string]bool,
 	}
 	walkSchema(schemaFields, false)
 }
+
+func requiresURISanitization(fieldName string) bool {
+	fieldName = strings.ToLower(fieldName)
+	suffixes := []string{"uri", "url", "endpoint"}
```
harshadixit12, line 133: Hashicorp vault has a `host` - would we want to sanitize that as well?
Would it be better to identify URIs based on the value rather than field name?

sanitize/sanitize.go
Prashansa-K: This would be useful for both plugin and vault config.
In case a uri is required, gateway expects a valid uri to be passed, otherwise sync fails.

Example for vault:
```
- config:
    client_id: null
   ...
    vault_uri: http://example-vault.com
  name: azure
```

sanitize/utils.go
Prashansa-K: > Would it be better to identify URIs based on the value rather than field name?

I had tried that earlier but golang url parser is not reliable. A non-uri value is also sometimes parsed as uri in case it has a `/` or a `:` and then sync fails.

> Hashicorp vault has a host - would we want to sanitize that as well?

Not required. It accepts normal string values too.

Prashansa-K: Ref example for url parsing issues: https://konghq.atlassian.net/browse/APIOPS-122

harshadixit12: Let's add a comment here that this is not an exhaustive list?

PR #1731, opened 2025-08-19
feat: added sanitization for route expressions

Prashansa-K:


tests/integration/dump_test.go
```diff
@@ -803,6 +803,11 @@ func Test_Dump_Sanitize_Special_Entities(t *testing.T) {
 			stateFile: "testdata/dump/008-sanitizer/vault-configs311.yaml",
 			runWhen:   func(t *testing.T) { runWhen(t, "enterprise", ">=3.11.0") },
 		},
+		// {
```
Prashansa-K: We don't spin up gateways with expression routers. Thus, commented this for now. Checking if we can create expression routers.

sanitize/expressions_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	tests := []struct {
+		name       string
+		expression string
+		expected   string
+	}{
+		{
+			name:       "Sanitize simple path in expression",
+			expression: `http.path == "/api/users"`,
+			expected:   `http.path == "ab1c696d4b37997a1c29a09610f47a7032949190db6c78d06a8d09b050ab91ea"`,
```
harshadixit12, line 23: We dont want to retain the slash in the path?

sanitize/expressions_test.go
```diff
@@ -0,0 +1,65 @@
...
+			name:       "Sanitize complex expression",
+			expression: `(http.path == "/api/users" || http.path == "/api/products") && http.method == "POST"`,
+			expected: ` ( http.path == "ab1c696d4b37997a1c29a09610f47a7032949190db6c78d06a8d09b050ab91ea"  || ` +
+				` http.path == "af59ca84d7d64ca44795aa3c0217c5ebe63f2ee593a67b3d77fad31bf3b84372" )   && ` +
+				` http.method == "POST"`,
+		},
+		{
+			name:       "Sanitize expression with regex",
+			expression: `http.path ~ r#"^/users/\d+$"#`,
+			expected:   `http.path ~ r#"dab30f7e826083df3eee6538650e002873dc7eae7cd8d617fc4dd2de2b97000b"#`,
```
harshadixit12, line 55: We lose context that user was trying to check for the path **beginning** with a certain snippet (here /users) - intended?

harshadixit12: I guess not something deck should be concerned with - as long as the output is a valid regex

Prashansa-K: yupp.

sanitize/expressions_test.go
Prashansa-K: no required here as it is an expression. A string is fine and accepted by the gateway.

sanitize/expressions_test.go
```diff
@@ -0,0 +1,65 @@
...
+		salt:         "test-salt-123",
+		sanitizedMap: make(map[string]interface{}),
+	}
+
+	tests := []struct {
+		name       string
+		expression string
+		expected   string
+	}{
+		{
```
harshadixit12, line 20: Can we cover each supported case mentioned in https://developer.konghq.com/gateway/routing/expressions/?
Also with, without leading / trailing spaces

sanitize/expressions.go
```diff
@@ -0,0 +1,229 @@
...
+	// 1. Fields: These are the identifiers like http.path, net.protocol, etc
+	// 2. Operators: These are the comparison operators like ==, !=, >, etc.
+	// 3. Values/Constant Values: These are the constant values like strings, numbers, IP addresses, etc.
+	// These components create predicates that can be combined with logical operators (&&, ||, etc).
+	// We need to sanitize an expression while preserving its structure.
+	// We will only sanitize values, not fields or operators.
+
+	// First, handling parentheses and logical operators to maintain expression structure
+	parenRegex := regexp.MustCompile(`(\(|\))`)
+	logicalOpRegex := regexp.MustCompile(`(&&|\|\|)`)
```
harshadixit12, line 61: Does this cover the NOT operator - `!`?

sanitize/expressions_test.go
Prashansa-K: Trailing/leading/middle spaces don't matter in expressions as checked from UI. So, I have left that and covered all cases

sanitize/expressions.go
Prashansa-K: ! is not a separator

PR #1737, opened 2025-08-25
fix: default lookup behaviour for consumer-group and consumers

Prashansa-K:


Prashansa-K:
Not required anymore.
The behaviour was fixed as a part of #1749

PR #1746, opened 2025-08-28
chore: updated timeout for goreleaser workflow

Prashansa-K:
go-releaser is taking 10 minutes and timing out before uploading assets. Thus, increasing the timeout period to avoid operation cancellation as it prevents us from releasing the cli.
Approval to increase timeout period: https://kongstrong.slack.com/archives/C027PKSNBC5/p1756369785658399?thread_ts=1756369711.863809&cid=C027PKSNBC5

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   goreleaser:
-    timeout-minutes: ${{ fromJSON(vars.GHA_DEFAULT_TIMEOUT) }}
+    timeout-minutes: 20
     runs-on: ubuntu-latest
     permissions:
       contents: write


.github/workflows/release.yaml
```diff
@@ -7,7 +7,7 @@ on:
 
 jobs:
   goreleaser:
-    timeout-minutes: ${{ fromJSON(vars.GHA_DEFAULT_TIMEOUT) }}
```
harshadixit12, line 10: Is this var common across org or repo?
Approving to unblock release - IMO better to set it as repo level variable

harshadixit12 (approved):

.github/workflows/release.yaml
Prashansa-K: This is an org level var

PR #1748, opened 2025-09-01
Upgrade vacuum

mheap:


mheap:
Data race issue raised at https://github.com/daveshanley/vacuum/issues/702

main.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 func registerSignalHandler() context.Context {
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancel(context.Background()) // #nosec G118
```
harshadixit12: golangci-lint v2.11 flags that cancel here is unused - because it is used within the go routine, and that means no guarantee it was called. However, we can't call it with `defer` because this function registers signal handler, and waits for interrupt - so ignoring this rule.

PR #1749, opened 2025-09-02
fix: consumer lookups while using default lookup tags for consumer-goups

Prashansa-K:
Syncing and diffing with default_lookup_tags
for consumer-groups was errorneous. While
dumping config, we would dump all consumers
in case of default-lookup tags. Thus, in case of
serial syncing of consumers, previous consumers
would end up getting deleted.
This fix ensures that no unexpected deletions occur
while using consumer-group lookups. 

Issue is described in detail in
the doc and FTI below: 
- [Document](https://konghq.atlassian.net/wiki/spaces/~7120204d7e8997ba384e09b1eee597d986a1dc/pages/4618027031/Default+LookUp+Tags+for+Deck)
- [FTI](https://konghq.atlassian.net/browse/FTI-6754)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/fatih/color v1.18.0
 	github.com/google/go-cmp v0.7.0
 	github.com/kong/go-apiops v0.1.49
-	github.com/kong/go-database-reconciler v1.27.1-0.20250916081703-5aa27ee92256
+	github.com/kong/go-database-reconciler v1.27.1
 	github.com/kong/go-kong v0.68.0
 	github.com/mitchellh/go-homedir v1.1.0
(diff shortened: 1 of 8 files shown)


go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/fatih/color v1.18.0
 	github.com/google/go-cmp v0.7.0
 	github.com/kong/go-apiops v0.1.49
-	github.com/kong/go-database-reconciler v1.26.0
+	github.com/kong/go-database-reconciler v1.26.1-0.20250902104042-b53c60cc6db1
```
Prashansa-K: To be updated post this pr merge: https://github.com/Kong/go-database-reconciler/pull/336

harshadixit12 (approved):

PR #1756, opened 2025-09-11
Replace periods underscore kong2tf

### LINKED ISSUE #1752: kong2tf when
There's an error in `kong2tf`. When using a consumer with username that includes . like `demo.some_value`. Terraform will throw and error when we use the generated resources.

```
│ Error: Invalid resource name
│
│   on cp.tf line 20, in resource "konnect_gateway_acl" "demo.some_value":
│   20: resource "konnect_gateway_acl" "demo.some_value" {
│
│ A name must start with a letter or underscore and may contain only letters, digits, underscores, and dashes.
```

mheap:
Resolves #1752


kong2tf/generate_resource.go
```diff
@@ -467,3 +467,23 @@ func contains(slice []string, item string) bool {
 	}
 	return false
 }
+
+// slugify converts a string into a slug using underscores
+func slugify(input string) string {
```
harshadixit12, line 483: This would still allow characters like `@, #` to go through, but terraform [does not allow in resource names](https://developer.hashicorp.com/terraform/language/syntax/configuration#identifiers).
Kong allows these in the entity names though.

What if we do something like this using `unicode` package - 
```suggestion
func keepRuneUnicode(r rune) rune {
	if unicode.IsLetter(r) || unicode.IsDigit(r) || r == '_' || r == '-' {
		return r
	}
	return '_'
}

func slugify(input string) string {
	slug := strings.Map(keepRuneUnicode, input)

	// Remove any consecutive underscores
	for strings.Contains(slug, "__") {
		slug = strings.ReplaceAll(slug, "__", "_")
	}

	// Trim leading and trailing underscores
	slug = strings.Trim(slug, "_")

	return slug
}
```

This would replace all disallowed characters with `_`

mheap: This works for me. Can you commit + add any additional test cases you think are important?

harshadixit12: Yes, I'll do 😄

harshadixit12: Yes, I'll do 😄

harshadixit12: I've added tests and this change in [40b4b33](https://github.com/Kong/deck/pull/1756/commits/40b4b33de19cd9363c440427d8fe2ed8747e818e), will merge if looks good.

mheap: LGTM

Issue #1747, opened 2025-08-28
Import hashed credentials for basic auth


mheap:
Support importing hashed credentials for basic auth. There's an existing branch from @GGabriele that we need to polish and release. This is a Konnect only feature

PR #1765, opened 2025-09-25
feat: added skip-hash for basic auth for Konnect

Prashansa-K:
This feature introduces a flag to be used with sync
command to skip-hashing of basic-auth credential
passwords when using with Konnect.
An info field is also added for the same.

Linked PRs:
https://github.com/Kong/go-kong/pull/576
https://github.com/Kong/go-database-reconciler/pull/342

For #1747


cmd/gateway_sync.go
```diff
@@ -106,6 +106,9 @@ to get Kong's state in sync with the input state.`,
 			"thus gaining some performance with large configs.\n"+
 			"Usage of this flag without apt select-tags and default-lookup-tags can be problematic.\n"+
 			"This flag is not valid with Konnect.")
+	syncCmd.Flags().BoolVar(&dumpConfig.SkipHashForBasicAuth, "skip-hash-for-basic-auth",
```
harshadixit12, line 109: Do we not want this flag for `gateway apply`?

Prashansa-K: will add

harshadixit12: Also, in cases where the basic auth has been synced before without the flag, using this flag leads to no diff and so no action - that is expected, right?

Prashansa-K: Yeah. That is expected.
The GET /basic-auth endpoint doesn't have any skip-hash param. So, there's no way of knowing if the returned password is a hash or not. Earlier too we didn't compare for passwords while syncing/diffing. So, I am retaining the same behaviour.

cmd/common.go
```diff
@@ -513,6 +519,18 @@ func determinePolicyOverride(targetContent file.Content, config dump.Config) boo
 	return false
 }
 
+func determineSkipHashForBasicAuth(targetContent file.Content, config dump.Config) bool {
+	if config.SkipHashForBasicAuth {
+		return true
+	}
+
+	if targetContent.Info != nil && targetContent.Info.SkipHashForBasicAuth {
```
harshadixit12, line 527: Why is this block necessary? 🤔

Prashansa-K: A user can choose to pass a command-line flag or use the info block to trigger the skip-hash functionality. So, this block looks at the info block to see if there's a skip-hash there.

harshadixit12: Not blocking, for knowledge purposes - we don't write the bool to the `Info` field on dump though, correct? Is adding field to `Info` block a common approach users take?

Prashansa-K: > we don't write the bool to the Info field on dump though, correct?

For this one, we can't. This is because the GET /basic-auth endpoint doesn't tell us if the password is a hash or not.
So, this particular thing is not valid for dump.

> Is adding field to Info block a common approach users take?

Yes. The metadata remains a part of the config. So, it is "safer", as folks can forget to pass the CLI flag.

harshadixit12 (approved):

PR #1770, opened 2025-09-30
ci: pinned commit shas to github workflows

Prashansa-K:
- Pinning all modules used in GH workflows with commit SHAs.
- Added a script that can help to fetch commit SHAs via GH API.
- Added a new workflow that ensures that all actions are pinned.

This is a security enhancement. We are pinning all
third-party actions that are used in our GH workflows.
This is to ensure that an incident like this doesn't recur:
- https://semgrep.dev/blog/2025/popular-github-action-tj-actionschanged-files-is-compromised/
- https://kongstrong.slack.com/archives/C08HW2ANB5G

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -177,6 +177,8 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 		cmd = "diff"
 	}
 
+	dumpConfig.SkipHashForBasicAuth = determineSkipHashForBasicAuth(*targetContent, dumpConfig)
+
 	var kongClient *kong.Client
 	mode := getMode(targetContent)
 	if mode == modeKonnect {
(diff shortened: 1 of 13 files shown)


harshadixit12:
@Prashansa-K can you add the reason for these changes as well?

scripts/get-action-commit-shas.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
```
Prashansa-K, line 1: Generated script.

.github/workflows/security.yaml
```diff
@@ -0,0 +1,19 @@
+on: push
+
+name: Security
+
+jobs:
+  ensure-pinned-actions:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@f4a75cfd619ee5ce8d5b864b0d183aff3c69b55a # v2.13.1
```
harshadixit12, line 10: not a blocker, calling out for visibility - but we seem to be using 2.11.1 in other workflows.

Issue #1753, opened 2025-09-10
Deck not treating _workspace tag in YAML in the same manner as --workspace flag


mheap:
When using "deck gateway validate" with a user whose permissions are scoped only to the workspace in question this only seems to function when passing in the --workspace flag on the command, when this is omitted deck seems to try and hit some endpoints which the user does not have permissions to access and fails, despite the workspace being specified within the YAML file itself.
 
```yaml
_format_version: "3.0"
_workspace: workspace2
services:
- connect_timeout: 60000
  enabled: true
  host: httpbin-ct.home.arpa
  name: workspace2_httpbin_service
  path: /anything
  plugins:
  - config:
      custom_fields_by_lua: {}
      path: /dev/stdout
      reopen: false
    enabled: true
    name: file-log
    protocols:
    - grpc
    - grpcs
    - http
    - https
  port: 80
  protocol: http
  read_timeout: 60000
  retries: 5
  routes:
  - https_redirect_status_code: 426
    name: workspace2_httpbin_route
    path_handling: v0
    paths:
    - /httpbin
    preserve_host: false
    protocols:
    - http
    - https
    regex_priority: 0
    request_buffering: true
    response_buffering: true
    strip_path: true
  write_timeout: 60000
```

but when running the command without the --workspace flag set this fails:

```bash
$ deck gateway validate workspace2.yaml --headers Kong-Admin-Token:workspace2_admin --verbose 2
GET /schemas/services HTTP/1.1
Host: localhost:8001
User-Agent: Go-http-client/1.1
Accept-Encoding: gzip

HTTP/1.1 403 Forbidden
Content-Length: 81
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: http://10.10.87.190:8002
Connection: keep-alive
Content-Type: application/json; charset=utf-8
Date: Tue, 02 Sep 2025 14:46:07 GMT
Server: kong/3.11.0.2-enterprise-edition
Vary: Origin
X-Kong-Admin-Latency: 31
X-Kong-Admin-Request-Id: 45b2a705c5f2516576598112e5b49f49

{"message":"workspace2_admin, you do not have permissions to read this resource"}
Error: building state: creating defaulter: get defaults for services: retrieve schema for services from Kong: HTTP status 403 (message: "workspace2_admin, you do not have permissions to read this resource")
```

with the --workspace flag this command succeeds as expected:

```bash
[...]

Issue #1700, opened 2025-07-22
deck validate ignore the _workspace field of state file


mheap:
Let's say we have workspace default and aaa. We also have a state file called kong.yaml, please see attached kong59954.yaml.
Inside the kong.yaml, _workspace is written as bbb, it is not exist in my kong gateway.

Next I test with below command

Command 1:

```
$ deck gateway validate kong.yaml
succeeded, even if _workspace is using a not existing workspace in kong.yaml
```

Command 2:

```
$ deck gateway validate kong.yaml -w aaa
Warning: Workspace 'aaa' specified via --workspace flag is different from workspace 'bbb' found in state file(s).
succeeded, even if _workspace is using a not existing workspace in kong.yaml
```

Command 3

```
$ deck gateway validate kong.yaml -w bbb
Error: workspace doesn't exist: bbb
This is expected, because bbb workspace do not existing.
```

So the command1 and command2 shows "deck validate" will ignore the _workspace field, which is not expected by customer.

The documentation on the flag when calling deck gateway validate -h has this description for the workspace flag:

```
validate configuration of a specific workspace (Kong Enterprise only).
This takes precedence over _workspace fields in state files.
```

This indicates that the _workspace field in the state file should be used if the flag isn’t provided but this doesn’t seem to be the case as Command 1 test shows, the field is instead ignored.

PR #1783, opened 2025-10-14
fix: use _workspace when given for gateway validate

harshadixit12:
Fixes:
1. https://github.com/Kong/deck/issues/1753
2. https://github.com/Kong/deck/issues/1700

Root cause for both: `gateway validate` did not respect `_workspace` set in the state file. The validation ran against `default` workspace, and therefore was incorrect as it did not load the referenced entities and schemas correctly. 
This led to unauthorised access of default workspace when user had permissions scoped to a different workspace as seen in https://github.com/Kong/deck/issues/1753.

With this fix, the online validation runs against the workspace specified in state file (unless overridden using `-w or --workspace`.

### DIFF
--- a/tests/integration/dump_test.go
+++ b/tests/integration/dump_test.go
@@ -803,9 +803,10 @@ func Test_Dump_Sanitize_Special_Entities(t *testing.T) {
 			skipValidation: true, // env vault validation endpoint not available in 3.4.0
 		},
 		{
-			name:      "dump sanitize env vault and vault references",
-			stateFile: "testdata/dump/008-sanitizer/env-vault.yaml",
-			runWhen:   func(t *testing.T) { runWhen(t, "enterprise", ">=3.5.0") },
+			name:           "dump sanitize env vault and vault references",
+			stateFile:      "testdata/dump/008-sanitizer/env-vault.yaml",
+			runWhen:        func(t *testing.T) { runWhen(t, "enterprise", ">=3.5.0") },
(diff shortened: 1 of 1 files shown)


tests/integration/sync_test.go
```diff
@@ -3246,7 +3246,7 @@ func Test_Sync_Upstream_Target_Konnect(t *testing.T) {
 			kongFile: "testdata/sync/004-create-upstream-and-target/kong3x.yaml",
 			expectedState: utils.KongRawState{
 				Upstreams: upstream,
-				Targets:   target,
+				Targets:   targetPost312,
```
harshadixit12, line 3249: This is deployed to dev in konnect - https://github.com/Kong/platform-api/pull/1868#discussion_r2425248307, so needs the change.

tests/integration/validate_test.go
```diff
@@ -246,6 +251,19 @@ func Test_Validate_Gateway_EE(t *testing.T) {
...
 			additionalArgs: []string{"--workspace=default"},
 		},
+		{
+			name:          "validate with non existent _workspace in state file",
+			stateFile:     "testdata/validate/kong-ee-non-existent-workspace.yaml",
+			errorExpected: true,
+			errorString:   "workspace doesn't exist: nonexistent",
+		},
+		{
+			name:           "validate with non-default _workspace and default_lookup_tags",
```
harshadixit12, line 261: This test would fail without the changes in `gateway_validate.go`

PR #1795, opened 2025-10-28
feat: added --skip-defaults flag in dump command

Prashansa-K:
Adding konnect integration tests for most used usecases.
As the customer asking for the feature is on konnect, I prioritised this.
Some changes are also pushed to the [GDR PR](https://github.com/Kong/go-database-reconciler/pull/352) to accomodate konnect schemas.

Todos in priority order:
- Adding gateway tests in next PR.
- Would handle vaults for konnect separately by creating separate GDR PR and tests

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/fatih/color v1.18.0
 	github.com/google/go-cmp v0.7.0
 	github.com/kong/go-apiops v0.2.0
-	github.com/kong/go-database-reconciler v1.28.2-0.20251029122407-53311dad7581
+	github.com/kong/go-database-reconciler v1.29.0
 	github.com/kong/go-kong v0.69.0
 	github.com/mitchellh/go-homedir v1.1.0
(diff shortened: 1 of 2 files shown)


tests/integration/dump_test.go
```diff
@@ -886,3 +886,59 @@ func Test_Dump_BasicAuth_SkipHash(t *testing.T) {
...
+			dumpArgs := []string{"-o", "-", "--skip-defaults"}
+			output, err := dump(dumpArgs...)
+			require.NoError(t, err)
+
+			expected, err := readFile(tc.expectedFile)
+			require.NoError(t, err)
+			assert.Equal(t, expected, output)
+
+			// ensure that dump can sync back without errors
+			require.NoError(t, sync(ctx, tc.stateFile))
```
harshadixit12: What do you think about adding one more step to this test - to dump again after sync?
This should tell us if any fields are getting changed by syncing the dumped-file with skipped defaults
PR LGTM otherwise.

harshadixit12 (approved):

PR #1799, opened 2025-10-30
tests: gateway tests for skip-defaults

Prashansa-K:
Added tests for gateway enterprise LTS versions.
Not adding for every version as the core logic would remain same across all, but the difference in schemas would lead to too many unnecessary files.

GDR PR (handling jwt_secrets sync): https://github.com/Kong/go-database-reconciler/pull/353


tests/integration/testdata/dump/009-skip-defaults/enterprise/3.4/service-scoped copy.yaml
```diff
@@ -0,0 +1,63 @@
+_format_version: "3.0"
```
harshadixit12: This file is unused?

Prashansa-K: deleted

PR #1802, opened 2025-11-03
ci: running expression router by default

Prashansa-K:


Prashansa-K:
Schemas for routes are different for both router types. Thus, these can't be merged for the sake of diffs and dumps.

Issue #1808, opened 2025-11-05
Allow skip-defaults for shorthand_fields


Prashansa-K:
At the moment, shorthand_fields don't have defaults defined in the schema. Thus, they are dumped with --skip-defaults flag.
Wherever possible, we can match the defaults with the new fields and skip them.

Points to remember:
- Old gateway versions use "translate_backwards" field with an object to get the new fields.
- New gateway versions (post 3.9) use "replaced_with" with an array to get the new fields.

PR #1810, opened 2025-11-06
tests: test fix and add for shorthand_fields handling in gateway dump

Prashansa-K:
Corresponding GDR PR: https://github.com/Kong/go-database-reconciler/pull/357
For #1808

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/fatih/color v1.18.0
 	github.com/google/go-cmp v0.7.0
 	github.com/kong/go-apiops v0.2.0
-	github.com/kong/go-database-reconciler v1.29.2-0.20251106074241-cb16789917c4
+	github.com/kong/go-database-reconciler v1.29.2
 	github.com/kong/go-kong v0.69.0
 	github.com/mitchellh/go-homedir v1.1.0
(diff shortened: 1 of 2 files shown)


tests/integration/testdata/dump/009-skip-defaults/enterprise/3.10+/plugin-partial-2.expected.yaml
```diff
@@ -13,18 +13,6 @@ plugins:
     - abc-client
     client_secret:
     - abc-secret
-    cluster_cache_redis:
-      timeout: 2000
     issuer: http://example-issuer.com
-    redis:
-      timeout: 2000
-    session_redis_cluster_max_redirections: 5
```
Prashansa-K, line 21: These were shorthand_fields. Hence, with the current change, they would no longer show up.

harshadixit12: Do we have an instance in tests where shorthand field was configured with a non-default value?

Prashansa-K: I will add one

PR #1813, opened 2025-11-10
fix: consumerGroupConsumer pulling modified to account for presence of consumers in targetContent

Prashansa-K:
We were checking for group memberships while pulling consumer-groups via default_lookup_tags.
Checking for these memberships is important when consumers are a part of the targetContent. That way, we
can ensure that any removals or updates in groups for consumers can go through.
However, if the targetContent under process doesn't have any consumers, we can skip the check and add the
necessary consumerGroupConsumers. This also ensures that we aren't deleting any consumerGroupConsumers
by accident, in case a group match fails.

FTI: https://konghq.atlassian.net/browse/FTI-6966?focusedCommentId=182193&sourceType=mention

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -504,11 +504,6 @@ func addUniqueConsumersInTargetContent(targetContent *file.Content, consumers []
 			if groupExists {
 				consumerGroupObject.Consumers = append(consumerGroupObject.Consumers, consumer)
 			}
-			// else {
-			// 	groups = append(groups, consumerGroup)
-			// 	targetContent.Consumers[index].Groups = groups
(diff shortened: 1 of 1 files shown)


cmd/common.go
```diff
@@ -493,6 +504,11 @@ func addUniqueConsumersInTargetContent(targetContent *file.Content, consumers []
 			if groupExists {
 				consumerGroupObject.Consumers = append(consumerGroupObject.Consumers, consumer)
 			}
+			// else {
```
harshadixit12: This branch is not required?

Prashansa-K: Stray code 😮‍💨  . I will remove it. Leave a comment when the initial review is completed.

PR #1816, opened 2025-11-12
chore: bumped golang to v1.25.3

Prashansa-K:
- Bumping golang to v1.25.3 to account for [FTI](https://konghq.atlassian.net/browse/FTI-7068?focusedCommentId=182771&sourceType=mention)
- This bump requires us to bump golangci-lint to a supported version.
- golangci-lint v8 only supports v2 lint-config. Config is migrated via `golangci-lint migrate` command based on the [guide](https://golangci-lint.run/docs/product/migration-guide/).
- Linted code.

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -457,11 +457,11 @@ func addUniqueConsumersInTargetContent(targetContent *file.Content, consumers []
 ) {
 	containsConsumerInTargetContent := func(consumer *kong.Consumer) (bool, int) {
 		for i, c := range targetContent.Consumers {
-			if c.Consumer.ID != nil && consumer.ID != nil && *c.Consumer.ID == *consumer.ID {
+			if c.ID != nil && consumer.ID != nil && *c.ID == *consumer.ID {
 				return true, i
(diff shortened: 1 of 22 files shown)


.github/workflows/test.yaml
```diff
@@ -46,7 +46,7 @@ jobs:
       - uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5
         with:
           go-version-file: go.mod
-      - uses: golangci/golangci-lint-action@ec5d18412c0aeab7936cb16880d708ba2a64e1ae # v6.2.0
+      - uses: golangci/golangci-lint-action@4afd733a84b1f43292c63897423277bb7f4313a9 # v8.0.0
```
harshadixit12, line 49: Let's update this in the `validate-action-shas.sh` script as well?

Prashansa-K: Where is that script?  This action seems to be referred here only.

harshadixit12: Ah my bad, it's not there in repo. It was a custom script on my local. PR lgtm.

harshadixit12 (approved):

PR #1845, opened 2026-01-19
feat: fetch all targets for all upstreams at once

GGabriele:
Avoid needing to loop through all upstreams to fetch all targets.

Requires:
- https://github.com/Kong/go-kong/pull/599
- https://github.com/Kong/go-database-reconciler/pull/365


Prashansa-K:
This PR is not required anymore. The gdr upgrade was taken up as a part of a separate PR which would have incorporated the changes proposed here.
Closing this.

Issue #1844, opened 2026-01-16
`--skip-consumers` creates consumers instead of plugins when using `default_lookup_tags`


alphayogi22:
### Summary
When using `--skip-consumers` flag combined with `default_lookup_tags.consumers` to reference existing consumers for consumer-scoped plugin creation, deck is **creating consumers** instead of **creating plugins**.
This issue appears related to #532 which reported similar behavior.

### Expected Behavior
When running `deck gateway diff` with:
- `--skip-consumers` flag
- `default_lookup_tags.consumers` configured to reference existing consumers by tag
- A plugin scoped to both a consumer and service

The output should show:
```
creating plugin rate-limiting-advanced for service my-service and consumer my-consumer-username
Summary:
  Created: 1
```

### Actual Behavior
With `--skip-consumers`, deck outputs:
```
creating consumer my-consumer-username
creating consumer another-consumer
Summary:
  Created: 2
```
**The plugin is NOT created. Instead, consumers are being created** even though `--skip-consumers` should prevent consumer operations.

#### 1. Ensure a consumer exists in Kong with a specific tag
Consumer in Konnect:
- **Username:** `my-consumer-username`
- **Tags:** `["MyConsumerTag"]`

#### 2. Create a deck state file
**deck-test-minimal.json:**
```json
{
  "_format_version": "3.0",
  "services": [{
    "name": "my-service",
    "host": "example.com",
    "port": 443,
    "protocol": "https",
    "path": "/api",
    "tags": ["my-service"]
  }],
  "plugins": [{
    "name": "rate-limiting-advanced",
    "consumer": "my-consumer-username",
    "service": "my-service",
    "config": {
      "limit": [30000],
      "window_size": [60],
      "strategy": "local",
      "window_type": "fixed",
      "namespace": "my-namespace"
    },
    "tags": ["my-service"]
  }],
  "_konnect": {
    "control_plane_name": "my-control-plane"
  },
  "_info": {
    "select_tags": ["my-service"],
    "default_lookup_tags": {
      "consumers": ["MyConsumerTag"]
    }
  }
}
```

#### 3. Run deck diff WITH `--skip-consumers`
```bash
deck gateway diff --konnect-addr https://eu.api.konghq.com \
  --konnect-token "$TOKEN" \
  --skip-ca-certificates \
  --skip-consumers \
  deck-test-minimal.json
```
**Output (INCORRECT):**
It is creating all the consumers again with the same tags (Why????)
```
creating consumer my-consumer-username
[...]

PR #1875, opened 2026-02-04
fix: added validation for --skip-consumers with default lookup consum…

shivaygupta-dotcom:
https://github.com/Kong/deck/issues/1844
https://konghq.atlassian.net/browse/FTI-7230

We have added a validation on deck, where --skip-consumers flag with default-look-up tags(consumer and consumer-groups) shouldn't be allowed together.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-ARG GO_VERSION=1.25.5
+ARG GO_VERSION=1.25.6
 FROM golang:${GO_VERSION} AS build
 WORKDIR /deck
 COPY go.mod ./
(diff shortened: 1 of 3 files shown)


cmd/common.go
```diff
@@ -155,6 +155,13 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 	if err != nil {
 		return err
 	}
+
+	if dumpConfig.SkipConsumers && targetContent.Info != nil && targetContent.Info.LookUpSelectorTags != nil &&
```
Prashansa-K: We can move the validation to a specific function for better readability.
Plus, there are functions already present in this file that check for lookup tags' presence. Those could be reused.

Issue #1839, opened 2025-12-22
Cannot use --skip-defaults when basic authentication is configured


imurata:
When attempting to use --skip-defaults while using the basic auth plugin, the following error is output:
```sh
$ deck gateway dump --skip-defaults
Error: getting Kong state: reading configuration from Kong: error removing defaults from basic auths: error fetching schema for entity basicauth_credential of type basicauth_credential: HTTP status 404 (message: "No entity named 'basicauth_credential'")
```
If no options are specified, a warning will be displayed but the dump will proceed normally.
```
$ deck gateway dump
Warning: basic-auth credentials detected, IDs will be exported
_format_version: "3.0"
consumers:
- basicauth_credentials:
  - id: 9e9590da-c63b-46c0-9ca3-742ff1e10fb6
    password: 02e83da18cc7ed2945b79e469c357e18a8f0e10f
    username: yamada
  custom_id: yamada
  id: 000e90f3-1f12-4758-ab41-335152d07c0f
  keyauth_credentials:
  - id: b2e76a34-ee67-4462-ab14-9a1cc54ac367
    key: yamada-key
  username: yamada
services:
- connect_timeout: 60000
  enabled: true
  host: httpbin.org
  id: 4f4296f9-204f-47f1-8226-b7d00f53eb29
  name: httpbin-service
  path: /
  port: 443
  protocol: https
  read_timeout: 60000
  retries: 5
  routes:
  - https_redirect_status_code: 426
    id: a48c0736-297b-4524-92a8-447ec937afa6
    name: httpbin-route
    path_handling: v0
    paths:
    - /httpbin
    plugins:
    - config:
        anonymous: null
        hide_credentials: false
        realm: service
      enabled: true
      id: f84eb5e8-589d-40ff-9f5e-e1fb81488d6f
      name: basic-auth
      protocols:
      - grpc
      - grpcs
      - http
      - https
      - ws
      - wss
    - config:
        anonymous: null
        hide_credentials: false
        identity_realms:
        - id: null
          region: null
          scope: cp
        key_in_body: false
        key_in_header: true
        key_in_query: true
        key_names:
        - apikey
        realm: null
        run_on_preflight: true
      enabled: true
      id: 32e110cb-a99c-4deb-9dc2-169dca5d972b
      name: key-auth
      protocols:
      - grpc
      - grpcs
      - http
      - https
      - ws
      - wss
    preserve_host: false
    priority: 0
    protocols:
    - http
    - https
    regex_priority: 0
[...]

PR #1889, opened 2026-02-12
fix: basic-auths dump with skip-defaults

Prashansa-K:
When `--skip-defaults` was set and upstream had basic-auth credentials,
dump was failing with error:
`Error: getting Kong state: reading configuration from Kong: error removing defaults from basic auths: error fetching schema for entity basicauth_credential of type basicauth_credential: HTTP status 404 (message: "No entity named 'basicauth_credential'")`

This is because the credential entity name was wrong in GDR.
This is corrected in this PR: https://github.com/Kong/go-database-reconciler/pull/374

Tests are added for kong gateway and konnect versions for all supported credential types.

For https://github.com/Kong/deck/issues/1839
FTI: https://konghq.atlassian.net/browse/FTI-7180

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/fatih/color v1.18.0
 	github.com/google/go-cmp v0.7.0
 	github.com/kong/go-apiops v0.2.2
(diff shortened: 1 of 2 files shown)


Prashansa-K:
Bumped to gdr to latest version. @harshadixit12

PR #1890, opened 2026-02-12
Add Gateway 3.14 conversion tooling

mheap:


convert/testdata/11/output-expected.yaml
```diff
@@ -0,0 +1,145 @@
...
+  plugins:
+  - config:
+      hide_credentials: false
+      ssl_verify: false
+    name: basic-auth
+  - config:
+      account_email: test@example.com
+      ssl_verify: false
+    name: acme
+  protocols:
```
Prashansa-K, line 82: ~This top level route had no protocols defined. As per the ruleset, this should have got `https` only, right?~

I interpreted it wrong. If no protocols are set, we are explicitly setting them to [http, https] (old defaults). I believe this is done so that nothing breaks in the config.

Issue #1859, opened 2026-02-03
deck: ability to automatically list and create new workspaces


mheap:
Related to the workspace support in Konnect

PR #1892, opened 2026-02-20
feat: Added changes to support workspace with konnect flow

shivaygupta-dotcom:
https://konghq.atlassian.net/browse/KOKO-3141
https://github.com/Kong/deck/issues/1859

Supporting PRs:
https://github.com/Kong/go-kong/pull/602
https://github.com/Kong/go-database-reconciler/pull/376

**Summary**:

This PR introduces workspace support for Konnect as part of the referenced Jira task. The changes extend existing commands to support workspaces when operating in Konnect mode. Added changes in sync, diff, reset, apply, dump and validate command. 
The overall flow is similar to the existing Kong workspace implementation, but Konnect uses different workspace API endpoints, which required additional handling.
In Konnect/Kong, all entities are attached to the default workspace by default but Konnect endpoint is bit different then what we have in Kong . Therefore, when the user explicitly specifies the default workspace, the system should follow the existing normal Konnect flow, without triggering workspace-specific logic.

Testing:
Verified that Konnect workspace behavior aligns with Kong workspaces, ensuring consistent behavior across both flows.
I have added all the integration tests below for the review reference.

**Integration tests For Konnect Workspaces:**

**_Dump Command_**

1. dump empty workspace
9. dump workspace with single entity
10. dump without _workspace field in state file
11. dump with default workspace name
12. workspace isolation - entities synced to workspace should not appear in CP-level dump
13. dump all workspaces (workspace1, workspace2, default)

**_Reset Command_**

1. reset workspace with single entity
8. reset workspace with multiple entities
9. reset without workspace flag (CP-level)
10. reset with default workspace name
11. reset workspace isolation - reset workspace1, verify workspace2 unaffected
12. reset all workspaces

**_Sync Command_**

1. sync entities into konnect workspace
2. sync empty workspace - no entities
3. sync without _workspace field in state file
4. sync with different workspace name - workspace mismatch
5. sync then reset to empty workspace

**_Diff Command_**

1. create konnect workspace with entity creation
2. create konnect workspace with existing entity deletion and new entity creation
3. update single entity in konnect workspace
4. empty state file should show deletion of all entities
5. workspace mismatch - different workspace name should create new workspace
6. idempotency - no diff after sync

**_Apply Command_**

1. apply entity into konnect workspace
2. apply and update entity in konnect workspace
3. apply is additive - does not delete existing entities
4. apply without _workspace field in state file

**_Validate Command_**

1. validate with _workspace in state file
2. validate with --workspace flag
3. validate without workspace (CP-level)
4. validate with default workspace in state file
5. validate with --workspace=default flag
6. validate empty workspace
14. validate with non-existent --workspace flag
15. validate workspace isolation - sync entity to workspace, validate both workspace and CP-level files

### DIFF
--- a/cmd/file_openapi2mcp.go
+++ b/cmd/file_openapi2mcp.go
@@ -1,0 +1,141 @@
+package cmd
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/kong/go-apiops/deckformat"
+	"github.com/kong/go-apiops/filebasics"
+	"github.com/kong/go-apiops/logbasics"
+	"github.com/kong/go-apiops/openapi2mcp"
+	"github.com/spf13/cobra"
+)
+
+var (
+	cmdO2MinputFilename       string
+	cmdO2MoutputFilename      string
+	cmdO2MdocName             string
+	cmdO2MoutputFormat        string
+	cmdO2MentityTags          []string
+	cmdO2MskipID              bool
+	cmdO2Mmode                string
+	cmdO2MpathPrefix          string
+	cmdO2MincludeDirectRoute  bool
+	cmdO2MignoreSecurityError bool
+)
+
+// Executes the CLI command "openapi2mcp"
+func executeOpenapi2MCP(cmd *cobra.Command, _ []string) error {
+	verbosity, _ := cmd.Flags().GetInt("verbose")
+	logbasics.Initialize(log.LstdFlags, verbosity)
+	_ = sendAnalytics("file-openapi2mcp", "", modeLocal)
+
+	if len(cmdO2MentityTags) == 0 {
+		cmdO2MentityTags = nil
+	}
+
+	cmdO2MoutputFormat = strings.ToUpper(cmdO2MoutputFormat)
+
+	options := openapi2mcp.O2MOptions{
+		Tags:                 cmdO2MentityTags,
+		DocName:              cmdO2MdocName,
+		SkipID:               cmdO2MskipID,
+		Mode:                 cmdO2Mmode,
+		PathPrefix:           cmdO2MpathPrefix,
+		IncludeDirectRoute:   cmdO2MincludeDirectRoute,
+		IgnoreSecurityErrors: cmdO2MignoreSecurityError,
+	}
+
+	trackInfo := deckformat.HistoryNewEntry("openapi2mcp")
+	trackInfo["input"] = cmdO2MinputFilename
+	trackInfo["output"] = cmdO2MoutputFilename
+	trackInfo["uuid-base"] = cmdO2MdocName
+
+	// do the work: read/convert/write
+	content, err := filebasics.ReadFile(cmdO2MinputFilename)
+	if err != nil {
+		return err
+	}
+	result, err := openapi2mcp.Convert(content, options)
+	if err != nil {
+		return fmt.Errorf("failed converting OpenAPI spec '%s' to MCP; %w", cmdO2MinputFilename, err)
+	}
+	deckformat.HistoryAppend(result, trackInfo)
+	return filebasics.WriteSerializedFile(cmdO2MoutputFilename, result, filebasics.OutputFormat(cmdO2MoutputFormat))
+}
+
+//
+//
+// Define the CLI data for the openapi2mcp command
+//
+//
+
+func newOpenapi2MCPCmd() *cobra.Command {
+	openapi2mcpCmd := &cobra.Command{
+		Use:   "openapi2mcp",
+		Short: "Convert OpenAPI files to Kong's decK format with MCP (Model Context Protocol) configuration",
+		Long: `Convert OpenAPI files to Kong's decK format with ai-mcp-proxy plugin configuration.
+
+This command generates a Kong service with an MCP route that includes the ai-mcp-proxy
+plugin configured with tools derived from the OpenAPI specification operations.
+
+Each OpenAPI operation is mapped to an MCP tool definition:
+  - operationId -> tool name (kebab-case normalized)
+  - summary/description -> tool description
+  - parameters -> tool parameters array
+  - requestBody -> tool request_body
+
+Security/ACL generation:
+  When an oauth2 security scheme includes the x-kong-mcp-acl extension, ACL entries
+  are automatically generated for each tool based on the operation's security scopes.
+  The plugin config will include acl_attribute_type, access_token_claim_field, and
+  per-tool acl.allow arrays. Use x-kong-mcp-default-acl at the document level to
+  set a default ACL for the plugin. Use --ignore-security-errors to skip unsupported
+  security configurations instead of failing.
+
+Supported x-kong extensions:
+  - x-kong-name: Custom entity naming
+  - x-kong-tags: Tags for all entities
+  - x-kong-service-defaults: Service entity defaults
+  - x-kong-route-defaults: Route entity defaults
+  - x-kong-upstream-defaults: Upstream entity defaults
+  - x-kong-plugin-*: Additional plugins
+
+MCP-specific extensions:
+  - x-kong-mcp-tool-name: Override generated tool name
+  - x-kong-mcp-tool-description: Override tool description
+  - x-kong-mcp-exclude: Exclude operation from tool generation (boolean)
+  - x-kong-mcp-proxy: Override ai-mcp-proxy plugin config at document level
+
+Security extensions:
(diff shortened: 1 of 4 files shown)


tests/integration/validate_test.go
```diff
@@ -63,19 +63,18 @@ func Test_Validate_Konnect(t *testing.T) {
...
-			name:           "validate with no konnect config in file, passed via cli flag konnect control plane",
-			stateFile:      "testdata/validate/konnect_invalid.yaml",
-			additionalArgs: []string{"--konnect-control-plane-name=default"},
-			errorExpected:  false,
+			name:          "validate with non existent _workspace in state file",
+			stateFile:     "testdata/validate/konnect-non-existent-workspace.yaml",
+			errorExpected: true,
+			errorString:   "workspace doesn't exist: nonexistent",
 		},
 		{
```
shivaygupta-dotcom, line 68: removed runtime group test because, `--konnect-runtime-group-name` is deprecated and CI was throwing error.

validate/validate.go
```diff
@@ -97,6 +97,18 @@ func (v *Validator) validateEntity(entityType string, entity interface{}) (bool,
 	nameOrID := getEntityNameOrID(entity)
 	errWrap := "validate entity '%s (%s)': %s"
 	endpoint := fmt.Sprintf("/schemas/%s/validate", entityType)
+
+	if v.client.IsKonnectMode() {
```
shivaygupta-dotcom, line 102: self review: just have check for req inside if, rest all can be common.

cmd/common_konnect.go
```diff
@@ -66,9 +66,18 @@ func GetKongClientForKonnectMode(
 	}
 
 	if konnectConfig.Token != "" {
-		konnectConfig.Headers = append(
-			konnectConfig.Headers, "Authorization:Bearer "+konnectConfig.Token,
-		)
+		found := false
```
Prashansa-K, line 70: Why is this change here?
What does it do?

shivaygupta-dotcom: In newly added workspace flow, there were multiple calls to `GetKongClientForKonnectMode` to obtain the client. As a result, the token was being appended repeatedly, which caused request failures. To fix this, I added a check to prevent the token from being appended multiple times.

Prashansa-K: Could you add the same thing as a comment above the if-else flow? That will make things clear for future contributors too.

cmd/common.go
```diff
@@ -74,6 +76,24 @@ func workspaceExists(ctx context.Context, config reconcilerUtils.KongClientConfi
 		return true, nil
 	}
 
+	if isKonnectMode {
+		if workspaceName == "default" {
```
Prashansa-K, line 80: This block seems unnecessary.
Condition on L69 already handles this and would exit even before we come here.

cmd/common.go
```diff
@@ -416,9 +434,17 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 			cprint.CreatePrintln("Creating workspace", wsConfig.Workspace)
 		}
 		if !dry {
-			_, err = rootClient.Workspaces.Create(ctx, &kong.Workspace{Name: &wsConfig.Workspace})
-			if err != nil {
-				return err
+			if mode == modeKonnect {
+				kongClient.SetKonnectFlag(true)
```
Prashansa-K: We have already set this on L248.
Why again here?

cmd/common.go
```diff
@@ -416,9 +434,17 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
...
+				_, err := kongClient.Workspaces.Create(ctx, &kong.Workspace{Name: &konnectConfig.WorkspaceName})
+				if err != nil {
```
Prashansa-K: Can be made common and returned later instead of repeating the check twice in both if-else block.

cmd/common_konnect.go
```diff
@@ -116,22 +126,63 @@ func resetKonnectV2(ctx context.Context) error {
...
 	if err != nil {
-		return err
+		return fmt.Errorf("getting initial konnect client: %w", err)
 	}
-	_, err = performDiff(ctx, currentState, targetState, false, 10, 0, client, true, resetJSONOutput, ApplyTypeFull)
-	if err != nil {
-		return err
+
+	var workspaces []string
+	if resetAllWorkspaces {
```
Prashansa-K: Do we need this duplicate logic? Can we re-utilise what we have in gateway_reset.go executeReset for this as well?

We can take out what is common and re-use in both flows - konnect as well as on-prem gateway.

cmd/common_konnect.go
Prashansa-K, line 149: Same as reset comment.
Let's see if we can re-utilise the normal dump flow and reduce some duplication here.

cmd/gateway_validate.go
```diff
@@ -335,40 +335,45 @@ func validateWithKong(
...
-		}
-	}
-
 	var (
 		kongClient *kong.Client
 		err        error
 	)
 	if mode == modeKonnect {
+		dumpConfig.KonnectControlPlane = konnectControlPlane
+		if validateWorkspace != "" {
```
Prashansa-K: We can reduce the duplication here and the else block too.

validate/validate.go
Prashansa-K: What did we find here?

tests/integration/validate_test.go
Prashansa-K: It is deprecated but it may still be used by clients.
What was the error?

tests/integration/validate_test.go
```diff
@@ -63,19 +63,18 @@ func Test_Validate_Konnect(t *testing.T) {
...
-			name:           "validate with no konnect config in file, passed via cli flag konnect runtime group",
+			name:           "validate with no konnect config in file, passed via cli flag konnect control plane",
 			stateFile:      "testdata/validate/konnect_invalid.yaml",
-			additionalArgs: []string{"--konnect-runtime-group-name=default"},
```
Prashansa-K, line 78: Why are these tests changed?

tests/integration/validate_test.go
```diff
@@ -91,7 +90,7 @@ func Test_Validate_Konnect(t *testing.T) {
 		{
 			name:           "validate with correct online list, passed via --online-entities-list cli flag",
 			stateFile:      "testdata/validate/kong3x.yaml",
-			additionalArgs: []string{"--online-entities-list=Services,Routes,Plugins"},
+			additionalArgs: []string{"--online-entities-list=Services,Routes,Plugins", "--konnect-control-plane-name=default"},
```
Prashansa-K: Why do we need to set additional CP name here?

tests/integration/validate_test.go
```diff
@@ -249,12 +248,6 @@ func Test_Validate_Gateway_EE(t *testing.T) {
 			stateFile:      "testdata/validate/kong-ee.yaml",
 			additionalArgs: []string{"--workspace=default"},
 		},
-		{
-			name:          "validate with non existent _workspace in state file",
-			stateFile:     "testdata/validate/kong-ee-non-existent-workspace.yaml",
-			errorExpected: true,
-			errorString:   "workspace doesn't exist: nonexistent",
```
Prashansa-K, line 256: This is a EE test. Why is this removed?

cmd/common.go
shivaygupta-dotcom: There might be condition, where user can specifically provide workspace as default, added comment below this line in code.

cmd/common_konnect.go
shivaygupta-dotcom: checked for this, code is bit too tightly coupled in gateway flow, so will check separately on this. I will update here, if it is possible to refactor.

tests/integration/validate_test.go
shivaygupta-dotcom: reverted.

tests/integration/validate_test.go
shivaygupta-dotcom: Sorry, by mistake.

tests/integration/validate_test.go
shivaygupta-dotcom: reverted.

tests/integration/validate_test.go
shivaygupta-dotcom: some tests were failing earlier, so added. reverted after discussion.

cmd/common_konnect.go
shivaygupta-dotcom: discussed offline with @Prashansa-K, we will skip dump refactoring due to reason mentioned above.

Issue #1842, opened 2026-01-13
Spike: Investigate sending defaults on create


mheap:
If we don't send defaults in a POST/PUT, deck shows a false diff when syncing. This causes issues in Konnect when there is a new default field added as the diff will show without user changes.

This is a spike to determine if we can remove defaults from the GET response before running a diff rather than injecting defaults into the create/update payload.

https://kongstrong.slack.com/archives/C07AQH7SAF8/p1767807300673819

Prashansa-K:
I ran a small check with services, plugins and partial on gateway 3.12.
If we remove the defaults from the payload we get from the backend and process that against the current file, we don't get any false diffs.
For plugins and partials, we had done a bunch of improvements in the differ as well as creation process to handle deprecations and auto fields. I will have to check this more thoroughly across different gateway versions and with Konnect for a final say. I will also have to test this against some other plugin types and credential entities.
The first experiment ran fine.
I also need to see what performance impact this change would have.

PR #1904, opened 2026-03-02
feat: no default injection during syncs

Prashansa-K:
This change prevents defaults from getting injected into state during sync/apply/diffs.
At the moment, this feature is Konnect only and needs the variable:
`DECK_SKIP_DEFAULTS_FILL=true` set in the environment to be enabled.

For testing, I have added dual testing mode for konnect - with default injection (current flow) and
without default injection (proposed flow).

Ref: https://github.com/Kong/go-database-reconciler/pull/391
For https://github.com/Kong/deck/issues/1842

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -17,6 +17,7 @@ import (
 	"github.com/kong/go-database-reconciler/pkg/diff"
 	"github.com/kong/go-database-reconciler/pkg/dump"
 	"github.com/kong/go-database-reconciler/pkg/file"
+	"github.com/kong/go-database-reconciler/pkg/schema"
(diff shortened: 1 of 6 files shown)


cmd/common.go
```diff
@@ -279,6 +286,9 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 		}
 	}
 
+	schemaRegistry = schema.NewRegistry(kongClient, mode == modeKonnect)
```
Prashansa-K, line 289: Using a common schema registry across Syncer, dump, rawState builder and plugin partial types to allow for more cache hits.

harshadixit12 (approved):

PR #1915, opened 2026-03-05
fix: data race issue due to mutating http.DefaultTransport

Prashansa-K:
Ref: https://github.com/Kong/go-database-reconciler/pull/392
Konnect tests often require re-runs due to data races' based failures. 
These races occur more often if we add more tests for konnect.
These data races are unpredictable due to the nature of golang's internal modules. 
The change in GDR removes the main cause of the data race.

This PR aims to make development velocity faster by avoiding such data races in 
Konnect or Kong gateway.

Added some corrections for konnect tests due to schema changes.

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -17,6 +17,7 @@ import (
 	"github.com/kong/go-database-reconciler/pkg/diff"
 	"github.com/kong/go-database-reconciler/pkg/dump"
 	"github.com/kong/go-database-reconciler/pkg/file"
+	"github.com/kong/go-database-reconciler/pkg/schema"
 	"github.com/kong/go-database-reconciler/pkg/state"
 	reconcilerUtils "github.com/kong/go-database-reconciler/pkg/utils"
 	"github.com/kong/go-kong/kong"
@@ -29,6 +30,8 @@ var (
 	assumeYes        bool
 	noMaskValues     bool
 	syncCmdAssumeYes bool
+
+	schemaRegistry *schema.Registry
 )
 
 type mode int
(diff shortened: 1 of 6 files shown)


cmd/gateway_ping.go
```diff
@@ -56,7 +57,10 @@ func pingKonnect(ctx context.Context) error {
 		return err
 	}
 	// get Konnect client
-	httpClient, err := utils.HTTPClientWithTLSConfig(rootConfig.TLSConfig)
+	httpClient, err := utils.HTTPClientWithOpts(utils.HTTPClientOptions{
```
Prashansa-K, line 60: Switched to the new method so that timeout field can be same as the one passed by the users in the config. If nothing is passed, default timeout = 10s is set.

cmd/konnect_dump.go
```diff
@@ -28,7 +29,12 @@ func newKonnectDumpCmd() *cobra.Command {
 	configure Konnect.` + konnectAlphaState,
 		Args: validateNoArgs,
 		RunE: func(cmd *cobra.Command, _ []string) error {
-			httpClient := utils.HTTPClient()
+			httpClient, err := utils.HTTPClientWithOpts(utils.HTTPClientOptions{
```
Prashansa-K, line 32: Bunch of these functions are not used anymore or are in alpha state and deprecated. Not dealing with removal as a part of this PR.

.github/workflows/integration-konnect.yaml
```diff
@@ -19,6 +19,7 @@ jobs:
       DECK_KONNECT_PASSWORD : ${{ secrets.DECK_KONNECT_PASSWORD }}
       DECK_KONNECT_ADDR : ${{ secrets.DECK_KONNECT_ADDR }}
       DECK_KONNECT_TOKEN : ${{ secrets.DECK_KONNECT_TOKEN }}
+      DECK_TIMEOUT: ${{ vars.DECK_TIMEOUT || '60' }}
```
Prashansa-K, line 22: 10s timeout for deck commands is often not enough for konnect-dev APIs. Thus, increased them to 60 here.

harshadixit12 (approved):

PR #1946, opened 2026-03-23
fix: validating passed address and workspace params before creating a http target

Prashansa-K:
Ref: https://konghq.atlassian.net/browse/APIOP-2

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -778,8 +778,11 @@ func fetchKongVersion(ctx context.Context, config reconcilerUtils.KongClientConf
 		if workspace == "" {
 			return "", err
 		}
+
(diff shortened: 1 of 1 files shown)


cmd/common.go
```diff
@@ -764,10 +778,26 @@ func fetchKongVersion(ctx context.Context, config reconcilerUtils.KongClientConf
 		if workspace == "" {
 			return "", err
 		}
+		// Validate address scheme before constructing the fallback URL.
+		if err := validateAddress(reconcilerUtils.CleanAddress(config.Address)); err != nil {
```
harshadixit12: One nit - if we store the returned value from `reconcilerUtils.CleanAddress(config.Address)`, we can avoid a subsequent call in line 796

LGTM other than this.

Issue #1856, opened 2026-02-03
openapi2kong generates duplicate plugin IDs when Content-Type header uses enum with multiple values


pexa-alao:
The openapi2kong command generates duplicate plugin IDs when a Content-Type header parameter is defined as an enum with multiple 
values in an OpenAPI 3.0 specification. This results in invalid Kong configuration that fails validation with the error: 
entity already exists.

Oas file:
```
openapi: 3.0.3
info:
  version: v1
  title: Example API
paths:
  /api/v1/example:
    post:
      operationId: exampleOperation
      x-kong-plugin-rate-limiting-advanced:
        enabled: true
        config:
          limit: [500]
      parameters:
        - name: Content-Type
          in: header
          required: true
          schema:
            type: string
            enum:
              - application/xml
              - text/xml
```

Generate Kong configuration:
`deck file openapi2kong --spec ./oas.yaml --output-file deck.yaml`

Validate the generated configuration:
`deck file validate deck.yaml`

[oas-minimal.yaml](https://github.com/user-attachments/files/25029208/oas-minimal.yaml)

[deck.yaml](https://github.com/user-attachments/files/25029211/deck.yaml)

Issue #1838, opened 2025-12-19
openapi2kong: Non-required enum header parameters are being treated as required


bhollingworth:
Using deck v1.52.0 or later, with the following command line:
`deck file openapi2kong --spec openapi.yaml`
and the following openapi.yaml file:
```
openapi: 3.0.4
info:
  title: example
  version: 1.2.0
servers:
  - url: https://example.com
paths:
  /example:
    get:
      parameters:
        - name: Accept-Language
          in: header
          required: false
          schema:
            type: string
            enum:
              - en-US
              - es-MX
      responses:
        '204':
          description: No content
```
Note that this OpenAPI specifies a **non-required** enum header parameter named `Accept-Language`.

The generated Kong configuration contains routes that match only if the incoming request contains an `Accept-Language` HTTP header with values of `en-US` or `es-MX`:
```
_format_version: "3.0"
services:
- host: example.com
  id: 730d612d-914b-5fe8-8ead-e6aa654318ef
  name: example
  path: /
  plugins: []
  port: 443
  protocol: https
  routes:
  - headers:
      Accept-Language:
      - en-US
    id: d2eea7d8-09dc-500f-bfa7-53fc469ba5db
    methods:
    - GET
    name: example_example_get_0
    paths:
    - ~/example$
    plugins: []
    regex_priority: 200
    strip_path: false
    tags: []
  - headers:
      Accept-Language:
      - es-MX
    id: a93257d4-0029-53bd-affd-7251a0c49250
    methods:
    - GET
    name: example_example_get_1
    paths:
    - ~/example$
    plugins: []
    regex_priority: 200
    strip_path: false
    tags: []
  tags: []
upstreams: []
```
If an incoming request does not contain an `Accept-Language` HTTP header, it receives a 404 `"no Route matched with those values"` error, even though the OpenAPI explicitly defines `required: false` for this header parameter.

PR #1954, opened 2026-03-25
feat: add a new command - deck file openapi2mcp

harshadixit12:
This PR adds three changes -
1. New command - `deck file openapi2mcp`
2. openapi2kong - Fix header based routing to only consider `required` headers, and add a fallback (through go-apiops bump)
3. openapikong - Fix duplicate plugin IDs generated for the routes when header based routing is used

Issues resolved:
https://github.com/Kong/deck/issues/1856
https://github.com/Kong/deck/issues/1838

### DIFF
--- a/cmd/file_openapi2mcp.go
+++ b/cmd/file_openapi2mcp.go
@@ -1,0 +1,141 @@
+package cmd
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/kong/go-apiops/deckformat"
+	"github.com/kong/go-apiops/filebasics"
+	"github.com/kong/go-apiops/logbasics"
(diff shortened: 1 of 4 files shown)


cmd/file_openapi2mcp.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+//
+//
+// Define the CLI data for the openapi2mcp command
+//
+//
+
+func newOpenapi2MCPCmd() *cobra.Command {
+	openapi2mcpCmd := &cobra.Command{
```
harshadixit12, line 75: Content taken from https://github.com/Kong/go-apiops/blob/main/cmd/openapi2mcp.go#L136

PR #1956, opened 2026-03-26
fix: Added test for upstream target weight update and moved update method from PATCH to PUT in go-kong

shivaygupta-dotcom:
Issue: https://github.com/Kong/deck/issues/1894

Summary:

1. decK sync was failing when modifying upstream target weights in Konnect because the update operation in target CRUD used the PATCH endpoint, which is not supported by Konnect. This has been fixed by switching the update operation to use PUT instead.
2. Added test for both kong and konnect to verify nothing should break after moving to PUT endpoint.

### DIFF
--- a/convert/plugin_names.go
+++ b/convert/plugin_names.go
@@ -10,6 +10,7 @@ const (
 	graphqlProxyCacheAdvancedPluginName   = "graphql-proxy-cache-advanced"
 	graphqlRateLimitingAdvancedPluginName = "graphql-rate-limiting-advanced"
 	proxyCacheAdvancedPluginName          = "proxy-cache-advanced"
+	redisPartialsPluginName               = "redis-partials"
 	aiProxyPluginName                     = "ai-proxy"
 	aiProxyAdvancedPluginName             = "ai-proxy-advanced"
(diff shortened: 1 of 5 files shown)


go.sum
```diff
@@ -211,8 +211,6 @@ github.com/kong/go-database-reconciler v1.33.0 h1:EAwap2hSCwHUJfVxbF+uw6UruqC47h
 github.com/kong/go-database-reconciler v1.33.0/go.mod h1:CTFu9hfOh2L4qPniCUCqLYhKiuVA0UsQe0Iv4jZc58o=
 github.com/kong/go-database-reconciler v1.35.0 h1:36Lmik7eoNMl+B/g6beWqBVgGl8b2vLXwNYNUVOtdRM=
 github.com/kong/go-database-reconciler v1.35.0/go.mod h1:AeQNEbEQ/Wy3OwJiWIo3PwWCWfav25F/FsTESqbelac=
-github.com/kong/go-kong v0.73.0 h1:Ct4EHJsJqoaYod39+gGy/JnU1664jzDbxTS2dk6Y628=
```
shivaygupta-dotcom, line 214: removed unused sum entries post running `go mod tidy`

PR #1974, opened 2026-04-06
fix: file conversion from 2.8 to 3.4 for post-function plugin

### LINKED ISSUE #1933: 2.8 -> 3.4 conversion is overwriting function definitions from "access" fields during conversions.
Consider a postfunction plugin that defines both `functions` and `access` fields.
"functions" was the old way. "access" is the new way.
While converting, deck changes `function` -> `access` and thus overwrites over the original access function. 
This is unintentional and leads to invalid conversions where customers miss out on the custom functions they wrote. Thus, before converting deck should check for the pre-existing access field and then perform the conversions.

Refer schemas and docs to understand how 2.8 dealt with both access and functions present together.

Refer: https://kongstrong.slack.com/archives/C04349E4KRC/p1773153822687389

harshadixit12:
## Summary
Referring to https://developer.konghq.com/gateway/upgrade/lts-upgrade-28-34/, if the 2.8 config has both `functions` (deprecated) and `access` keys in the plugin config, we will modify logic to replace the legacy key with new key.

New approach:
- Both exist, old field is empty: Choose new
- Both exist, both have values:: Error
- Both exist, new field is empty: Choose old
- Only old exists: Choose old

## How 2.8 used to work with `functions` and `access`
Based on https://legacy-gateway-2-8--kongdocs.netlify.app/hub/kong-inc/post-function/how-to/phases/ - the fields `functions` and `access` refer to the different phases in which we can configure the plugin to execute.
In kong 3.0, `functions` was removed, and `access` phase was to be used instead.

**Even in 2.8 - `functions` and `access` are treated as mutually exclusive** - sync fails if both are present in config and are non-empty.

Fixes https://github.com/Kong/deck/issues/1933

### DIFF
--- a/tests/integration/testdata/file-convert/002-kong-gateway-28x-to-34x-migration/28x-plugins-legacy-new-replace.yaml
+++ b/tests/integration/testdata/file-convert/002-kong-gateway-28x-to-34x-migration/28x-plugins-legacy-new-replace.yaml
@@ -6,20 +6,6 @@ services:
   host: mockbin.org
   name: mockbin
   routes:
-  - name: r1
-    https_redirect_status_code: 301
-    paths:
-    - /r1
-    plugins:
-    - config:
-        # Both legacy and new fields are non empty
-        access:
-        - kong.log.err("bar")       kong.response.exit(418)
-        # legacy field
-        functions:
-        - kong.log.err("foo")       kong.response.exit(418)
-      enabled: true
-      name: post-function
   - name: r2
     https_redirect_status_code: 301
     paths:
--- a/tests/integration/testdata/file-convert/002-kong-gateway-28x-to-34x-migration/28x-plugins-legacy-new-replace.yaml
+++ b/tests/integration/testdata/file-convert/002-kong-gateway-28x-to-34x-migration/28x-plugins-legacy-new-replace.yaml
@@ -1,0 +1,59 @@
+_format_version: "1.1"
+
+services:
+- connect_timeout: 60000
+  enabled: true
+  host: mockbin.org
+  name: mockbin
+  routes:
+  - name: r1
+    https_redirect_status_code: 301
+    paths:
+    - /r1
+    plugins:
+    - config:
+        # Both legacy and new fields are non empty
+        access:
+        - kong.log.err("bar")       kong.response.exit(418)
+        # legacy field
+        functions:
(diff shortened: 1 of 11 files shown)


tests/integration/testdata/file-convert/002-kong-gateway-28x-to-34x-migration/28x-plugins-legacy-new-replace.yaml
```diff
@@ -0,0 +1,59 @@
...
+  host: mockbin.org
+  name: mockbin
+  routes:
+  - name: r1
+    https_redirect_status_code: 301
+    paths:
+    - /r1
+    plugins:
+    - config:
+        # Both legacy and new fields are non empty
```
Prashansa-K: This config should have resulted in an error, right?
We have set `errorExpected: false` and the tests pass.

harshadixit12: Yes - I wanted to check with you about this   - during conversion of plugin config as of now in other functions, the errors are printed to console, but execution is not halted - https://github.com/Kong/deck/blob/main/convert/plugin_updates.go#L243
Shall I modify the function signature and return error instead?

This is what the output looks like

```
> go run main.go file convert --from 2.8 --to 3.4 --input-file x.yaml -o y.yaml

From the 'x.yaml' config file,
the _format_version field has been migrated from '1.1' to '3.0'.

Error: both legacy configuration field "functions" and new field "access" have values in plugin post-function, please remove the legacy field
Automatically converted legacy configuration field "functions" to the new field "access" in plugin post-function
```

Prashansa-K: We should halt the conversion and show the error. Otherwise, the error output would get mixed with other information printed. It would be hard to spot the error as the file size increases and there's too much info on the screen and a non-zero exit code won't give any feedback to our users to check for the printed warnings..

> Shall I modify the function signature and return error instead?

Yes.

harshadixit12: Makes sense, will update.

Prashansa-K (approved):

PR #1979, opened 2026-04-07
Make kong2tf plugin aware

mheap:
Automatically generate a `konnect_gateway_custom_plugin` instance if the provided plugin is not in the list of bundled plugins.

You can update the bundled plugins by running `make generate-plugin-list` while running a Gateway locally.


kong2tf/generate_resource.go
```diff
@@ -94,9 +94,34 @@ func generateResourceWithCustomizations(
...
 	}
 
+	isCustomPlugin := false
+	var s string
 	if entityType == "gateway_plugin" {
-		entityType = fmt.Sprintf("%s_%s", entityType, name)
-		delete(entity, "name")
+		// If `name` is not in availablePlugins, then it's a custom plugin and
+		// we should use the plugin name as the resource name
+		if !contains(availablePlugins, entity["name"].(string)) {
```
harshadixit12, line 102: When a new plugin is introduced in a new kong gateway version - wouldn't the previously released versions of deck treat it as custom plugin

mheap: Yes, that's correct. That's the tradeoff we're making (it would still work, just without schema checks in TF)

kong2tf/generate_resource.go
```diff
@@ -94,9 +94,34 @@ func generateResourceWithCustomizations(
...
+			isCustomPlugin = true
+			if customizations == nil {
+				customizations = map[string]string{}
+			}
+			customizations["config"] = "jsonencode"
```
harshadixit12, line 107: When we do jsonencode - underlying value of `config` is a string in https://github.com/Kong/terraform-provider-konnect/blob/main/src/custom_plugin_model.go#L40 - config.String() returns string wrapped in additonal characters, which breaks unmarshal. (ex: output is `"\"{\\\"hello\\\":\\\"world\\\"}\""` instead of `"{\"hello\":\"world\"}"`)
We'll need to fix this in custom_plugin_model.go (i'll do it)

I don't think we need to block this PR though.

harshadixit12 (approved):

PR #1994, opened 2026-04-10
feat: added no-merge flag to sync, diff, apply command

Prashansa-K:
--no-merge flag enables deck to process the
files passed one by one, rather than merging
them all together and send as a whole.
This is useful in case entities are split
across multiple files with different
select-tags. Merging would tag all entities
with the mentioned select-tags. A no-merge
option bypasses this issue. It also keeps our
request urls sent to admin-api under maxLength

Basing this over 3.13 and 3.14 branch tests branch and tests/konnect to avoid conflicts later.

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -494,6 +494,151 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 	return nil
 }
 
+// uniqueKeysByType maps each entity type name to the field names that must be
+// unique within that type. reflect.Value.FieldByName resolves promoted fields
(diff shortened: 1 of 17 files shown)


Prashansa-K:
Holding off on merge:
This is a basic usecase here.
The following issues still exist:
- If a directory is passed `deck gateway sync dirname --no-merge`, files aren't executed without merge.
- diff and validate don't check for "duplicates" yet.

Prashansa-K:
Will add more tests to this to cover directory traversal and duplicates checking.
Need to update validate and apply workflows to account for batching as well.

Issue #2037, opened 2026-04-28
`deck file openapi2kong` creates multiple routes based on path parameter enum - need an opt-out


harshadixit12:
deck file openapi2kong generates separate Kong routes for each header parameter combination when an OAS spec defines headers with both required: true and enum values. There is currently no CLI flag or OAS annotation to suppress this behaviour. 
Customers who cannot modify their OAS specs — because those specs feed external API catalogs shared with third-party consumers — have no way to restore the original single-route output.

PR #2039, opened 2026-04-29
fix: Added a cli flag for openapi2kong to skip header routes

shivaygupta-dotcom:
GH: https://github.com/Kong/deck/issues/2037
FTI: https://konghq.atlassian.net/browse/FTI-7492

Supporting PR:
https://github.com/Kong/go-apiops/pull/292

### DIFF
--- a/cmd/file_openapi2kong.go
+++ b/cmd/file_openapi2kong.go
@@ -23,7 +23,7 @@ var (
 	cmdO2Ksecurity            bool
 	cmdO2KignoreSecurityError bool
 	cmdO2KignoreCircularRefs  bool
-	cmdO2KskipHeaderRoutes    bool
+	cmdO2KskipRouteByHeader   bool
 )
 
 // Executes the CLI command "openapi2kong"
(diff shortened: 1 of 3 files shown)


cmd/file_openapi2kong.go
```diff
@@ -23,6 +23,7 @@ var (
 	cmdO2Ksecurity            bool
 	cmdO2KignoreSecurityError bool
 	cmdO2KignoreCircularRefs  bool
+	cmdO2KskipHeaderRoutes    bool
```
harshadixit12: nitpick - we can call this also `cmdO2KskipRouteByHeader`?

cmd/file_openapi2kong.go
```diff
@@ -115,6 +117,9 @@ The output will be targeted at Kong version 3.x.
 		"ignore errors for unsupported security schemes")
 	openapi2kongCmd.Flags().BoolVar(&cmdO2KignoreCircularRefs, "ignore-circular-refs", false,
 		"ignore circular $ref errors in the OpenAPI spec (dangerous, use with caution)")
+	openapi2kongCmd.Flags().BoolVar(&cmdO2KskipHeaderRoutes, "skip-route-by-header", false,
+		"skip generation of separate routes for each header parameter enum value.\n"+
```
harshadixit12: can we update this to 
```
skip generation of separate route for each required header path parameter enum value.
```?

harshadixit12: @shivaygupta-dotcom apologies for the miss - let's remove `path` - LGTM otherwise

cmd/file_openapi2kong.go
harshadixit12: This is addressed?

shivaygupta-dotcom: yes addressed previously, but somehow got reverted. just pushed again.

harshadixit12 (approved):

PR #2049, opened 2026-05-06
Add -W and -E flags to allow users to control error severity

mheap:
Uses emitter from https://github.com/Kong/go-database-reconciler/pull/466

When we added the OIDC validation we received feedback that some customers want to treat this as a warning and not an error. We do not want to make this a warning only, and so I've added the ability for users to choose which validation are an error and which are a warning using the `-W` and `-E` flags

**NOTE: This is using a branch from go-database-reconciler. Once the GDR PR is merged, I'll update this PR to point at a release**


tests/integration/diagnostic_policy_live_test.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+func Test_SyncDiagnosticSeverityOverrides_RLAWarningAsError(t *testing.T) {
+	runWhen(t, "enterprise", ">=3.4.0")
+	setup(t)
+
+	err := sync(
+		context.Background(),
+		"testdata/render/005-diagnostics/rla-consumer-groups.yaml",
+		"-W", "rla-consumer-groups-deprecated",
```
Prashansa-K: I don't think this particular code should be allowed as a warning. Do customers require this deprecated way of defining configs?
We have had a lot of consumer-group reconciliation bugs and I don't want a deprecated flow that used to error out since at least 2 years to be allowed now. I don't know what other can of worms would it open up if folks start using it.

tests/integration/diagnostic_policy_test.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func Test_RenderDiagnosticSeverityOverrides_InvalidCode(t *testing.T) {
+	_, err := render(
+		"testdata/render/005-diagnostics/route-regex.yaml",
+		"-E", "not-a-real-code",
+	)
+	require.Error(t, err)
+	require.ErrorContains(t, err, "Valid diagnostic codes")
+	require.ErrorContains(t, err, "route-regex-path-format")
```
Prashansa-K, line 59: @mheap How would folks know about these valid codes?
I don't see them in the help output.

tests/integration/diagnostic_policy_live_test.go
mheap: There's one customer that still uses this heavily. It used to be a 2.8 gate, and we moved it to be 3.4.

By adding this option we provide a "you can make this a warning, but you're on your own" option. The alternative is that we bump the gate to 3.10 (and 3.14 etc as they upgrade) and have to officially support it

tests/integration/diagnostic_policy_test.go
mheap: The intention is that we only tell people who hit issues. We should continue to be opinionated (e.g. OIDC salt is an error). Support will be enabled on which codes are available to respond.

The "hidden" way to get a list is to pass an invalid value:

```
❯ go run main.go file render -W foobar
Error: invalid value for -W/--errors-as-warnings: unknown diagnostic code: foobar. Valid diagnostic codes: [oidc-missing-required-config rla-consumer-groups-deprecated route-regex-path-format]
exit status 1
```

Prashansa-K (approved):

PR #2057, opened 2026-05-12
Make DECK_FORMAT work

### LINKED ISSUE #2055: DECK_FORMAT=json environmental variable not being respected when running commands that use the --format flag
### Issue Context:
We use json files as our standard config file across many services and CI/CD builds and we want to make the deck cli tool default all config file based operations to default to json instead of yaml instead of requiring every relevant command to have `--format json` attached to the end of it.
The documentation says that all CLI flags can be set via environmental variables https://developer.konghq.com/deck/configuration/#environment-variables
But when setting the equivalent format local environmental variable `DECK_FORMAT=JSON` the deck tool does not appear to use it when running commands that take in the --format flag.

### Setup Info:
We see this using both temporary environmental variables and permanently saved variables, our current system is Windows and we see this behavior on Deck v1.60

### Reproduction:
Set Environmental Variables (via powershell):
```
$env:DECK_KONNECT_CONTROL_PLANE_NAME = "dev-us-east-1"
$env:DECK_FORMAT = "json"
```
Or permanently:
```
[System.Environment]::SetEnvironmentVariable("DECK_KONNECT_CONTROL_PLANE_NAME", "dev-us-east-1", "User")
[System.Environment]::SetEnvironmentVariable("DECK_FORMAT", "json", "User")
```

Deck command that doesn't respect this environmental variable:
`deck gateway dump --consumer-group-policy-overrides --with-id -o devgateway.json`

Equivalent Command that produces the expected output:
`deck gateway dump --consumer-group-policy-overrides --with-id -o devgateway.json --format json`

mheap:
Enables using the `DECK_FORMAT` environment variable for `--format`

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -28,6 +28,18 @@ import (
 
 const exitCodeDiffDetection = 2
 
+// getFormatFlagValue returns the effective format value for a command.
+// CLI flag takes priority, then DECK_FORMAT env var, then the flag default.
+func getFormatFlagValue(cmd *cobra.Command, flagValue string) string {
+	if cmd.Flags().Changed("format") {
+		return flagValue
+	}
+	if envVal := os.Getenv("DECK_FORMAT"); envVal != "" {
+		return envVal
+	}
+	return flagValue
+}
+
 var (
 	assumeYes        bool
 	noMaskValues     bool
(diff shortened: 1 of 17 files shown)


shivaygupta-dotcom (approved):

harshadixit12:
Skipping the konnect tests since this PR only covers formatting.

PR #2087, opened 2026-06-01
tests: removed unrelated tests from konnect test suite

Prashansa-K:
Konnect CI takes >20 minutes. To avoid random timeouts, and reduce time-to-merge, some tests that make no sense to be tested against Konnect are skipped here. This would give a little buffer to avoid timeouts.
A proper fix for this increased CI time would be to split testing across multiple orgs. While we do not implement that, we wish to keep the CI time under 25 minutes. Hence, the buffer.


Prashansa-K:
Bumping this up for review: @shivaygupta-dotcom

shivaygupta-dotcom (approved):

PR #2099, opened 2026-06-04
fix: Added check for lint cmd to handle empty or only comment files.

shivaygupta-dotcom:
GH: https://github.com/Kong/deck/issues/2091

Added validation in `deck file lint` to handle empty state files and files containing only comments.


lint/lint.go
```diff
@@ -81,6 +82,10 @@ func WithContent(
 	if err != nil {
 		return nil, err
 	}
+	var parsedContent interface{}
+	if err := yaml.Unmarshal(stateFileBytes, &parsedContent); err == nil && parsedContent == nil {
+		return nil, fmt.Errorf("state file is empty or contains only comments")
```
Prashansa-K: Can you check the linked FTI once and ask the reporter their expectations?
They expect a return code of `0` as per the description. 
An error would give return code as `1`.

An older deck version didn't error out in case of an empty/commented file. Maybe we can return with a warning message, rather than an error and give a return code `0`.

Prashansa-K: Ensure that the warning message doesn't break json output.

shivaygupta-dotcom: Just checked description, I will handle it with warning instead of error.

lint/lint.go
```diff
@@ -81,6 +83,17 @@ func WithContent(
 	if err != nil {
 		return nil, err
 	}
+	var parsedContent interface{}
+	if err := yaml.Unmarshal(stateFileBytes, &parsedContent); err == nil && parsedContent == nil {
+		// Return empty results with a warning instead of an error
+		// to maintain backward compatibility with older deck versions
+		_, _ = fmt.Fprintln(os.Stderr, "Warning: state file is empty or contains only comments, skipping linting")
```
Prashansa-K: You can utilise the cprint package from utils for the same.
```
warningMessage := fmt.Sprintf("...")
cprint.UpdatePrintlnStdErr(warningMessage)
```

Prashansa-K:
Changes look fine. Will approve once CI passes.

Prashansa-K (approved):

PR #2106, opened 2026-06-08
tests: cloned plugins and konnect tests

Prashansa-K:
Cloned Plugins are now available on Konnect Dev.


tests/integration/testdata/apply/013-cloned-plugin-definitions/update-plugin-config.yaml
```diff
@@ -3,7 +3,7 @@ plugins:
 - name: new-file-log
   config:
     path: /tmp/file-updated.log
-    reopen: false
+    reopen: true
```
Prashansa-K, line 6: Changed this as `reopen: false` is a default value.
Didn't want to create two separate flows for skip-defaults-fill konnect in order to handle just this field.
We are already updating tags and other field to test this plugin.

.github/workflows/test.yaml
```diff
@@ -49,6 +49,7 @@ jobs:
       - uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20 # v9.2.0
         with:
           version: v2.11.4
+          install-mode: goinstall
```
Prashansa-K, line 53: Switched to this mode to get rid of transient failures from GH.

shivaygupta-dotcom (dismissed):

shivaygupta-dotcom (approved):

PR #2114, opened 2026-06-17
fix(cmd): negative or zero parallelism

manishsvk:
### Summary

This change adds a `checkParallelism` helper so an invalid value is rejected up front with a usage error, before any state file is read or any request is sent to Kong. Valid values (>= 1) are unaffected.

### Full changelog

* [Added `checkParallelism` helper in `cmd/common.go`]
* [Fix for `--parallelism < 1` (covers gateway `sync`, `diff`, `apply` and `validate` commands)]
* [Updated `go-database-reconciler` version]
* [Added unit tests for `checkParallelism` (zero, negative, positive)]

### Issues resolved

Fix #375

### Testing

- [x] Unit tests
- [x] Manual testing on local gateway instance


harshadixit12:
LGTM - will approve once CI is fixed

harshadixit12 (dismissed):

Prashansa-K:
This should have a go.mod update with a new GDR version that has the changes added here:
https://github.com/Kong/go-database-reconciler/pull/483

harshadixit12:
> This should have a go.mod update with a new GDR version that has the changes added here: [Kong/go-database-reconciler#483](https://github.com/Kong/go-database-reconciler/pull/483)

We skipped it because it was not strictly required - it can go with next gdr release IMO

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/kong/deck
 
-go 1.26.3
+go 1.26.4
```
harshadixit12, line 3: @manishsvk since the golang version is changing, let's update the docker file as well to build with this version?
The digest for new version can be copied from https://hub.docker.com/layers/library/golang/1.26.4/images/sha256-84e4dd2658fb85e1debf9fbe653aa386ed5bf108429ea641ab0a7db31a8ed5cc

Prashansa-K: We can do this here for onboarding work. 
For future references, such updates are not necessary. Dockerfile picks up the go version from go.mod. The version in dockerfile is a "fallback". [Ref](https://github.com/Kong/deck/blob/main/.github/workflows/release.yaml#L91)

cmd/common.go
```diff
@@ -178,6 +178,9 @@ func RemoveConsumerPlugins(targetContentPlugins []file.FPlugin) []file.FPlugin {
 func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 	delay int, workspace string, enableJSONOutput bool, applyType ApplyType,
 ) error {
+	if err := checkParallelism(parallelism); err != nil {
```
Prashansa-K: This check should be in PreRunE hook for the diff, sync, apply commands.

harshadixit12: IMO having it here avoids duplication, any reason why we want in PreRunE hook instead?

Prashansa-K: It does avoid duplication but PreRunE is the hook for such things to fail early. We have other checks in syncMain because they rely on targetState building. 
It's mostly about Separation of Concerns. A validity check on a flag or argument goes in PreRunE.

manishsvk: done

go.mod
manishsvk: done

harshadixit12 (dismissed):

harshadixit12 (approved):

PR #2115, opened 2026-06-18
tests: fixes for upcoming 3.15 gateway release

Prashansa-K:
Schema changes led to some failures in pre-existing tests for 3.15
Adding those changes here.

Also, ignored config checking for conditional plugins as our aim is to check
for "condition" field. Each new release will lead to more config variations and we
don't wish to bloat the tests.

Validated the fixes for 3.15 here: https://github.com/Kong/deck/actions/runs/27747469499/job/82089250439

### DIFF
--- a/tests/integration/sync_test.go
+++ b/tests/integration/sync_test.go
@@ -304,6 +304,44 @@ var (
 		},
 	}
 
+	plugin315 = []*kong.Plugin{
+		{
+			Name: kong.String("basic-auth"),
+			Protocols: []*string{
+				kong.String("grpc"),
+				kong.String("grpcs"),
+				kong.String("http"),
+				kong.String("https"),
+			},
+			Enabled: kong.Bool(true),
+			Config: kong.Configuration{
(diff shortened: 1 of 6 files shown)


tests/integration/sync_test.go
```diff
@@ -4869,6 +5113,8 @@ u8S6yNlu2Q==
 // deployed certificate, and then a GET is performed to test the
 // proxy functionality, which should return a 200.
 func Test_Sync_Vault(t *testing.T) {
+	// Skipping till we make this test more resilient
+	t.Skip()
```
harshadixit12: Rather than skipping the entire test, can we comment out the part where we make http request to a route that proxies to httpbin?
We can still test for kong state reliably.

Prashansa-K:
The new commit has added vault test back which is limited to EE. Hence, skipped the konnect tests. They ran successfully the last time.

Prashansa-K:
https://github.com/Kong/deck/actions/runs/27809703711/job/82297036933
Tested the new commit against 3.15 rc. 🟢

harshadixit12 (approved):

Issue #1730, opened 2025-08-19
Support specifying kong-admin-token using an environment variable


jijiechen:
The `--kong-addr` can be specified using an environment variable `DECK_KONG_ADDR`, this is very good and it saves a lot of time when executing `deck` command multiple times.

If you can also support using another environment variable, for example `DECK_KONG_ADMIN_TOKEN`, to specify the admin token, that is even better. 

It is not only more convenient, it's more secure. Because when people share their steps, they copy commands: by only specifying the admin token once, they hide the token from subsequent commands.

<img width="1433" height="699" alt="Image" src="https://github.com/user-attachments/assets/a1322fc4-aeae-4cac-a5ab-028a46dbd7fb" />

VarunAthreya:
@jijiechen we have released a new flag for kong-admin-token in v1.64.0.

PR #2123, opened 2026-06-23
feat(cmd): add kong-admin-token flag

VarunAthreya:
### Summary
Adds a new `--kong-admin-token` persistent flag that lets users authenticate against Kong's Admin API. The token is injected as a `Kong-Admin-Token` header on requests to the gateway.

GH: #1730 

### Changes
- **New `--kong-admin-token` flag** registered on the root command. It can also be set via the `DECK_KONG_ADMIN_TOKEN` environment variable (bound through Viper), consistent with deck's other config options.
- **Refactored `extendHeaders`** to accept a variadic list of optional headers via a new `header{name, value}` struct. Headers with an empty value are skipped, so the `Kong-Admin-Token` header is only added when a token is actually provided.
- **Wired the token into `initConfig`**, appending the `Kong-Admin-Token` header alongside the existing `User-Agent` header.

### Notes
- Empty tokens are a no-op — no header is sent if the flag/env var is unset, preserving existing behavior.
- The variadic refactor keeps the existing `User-Agent` injection intact while making the helper reusable for future headers.


cmd/root.go
```diff
@@ -451,9 +462,19 @@ func initKonnectConfig() error {
...
-func extendHeaders(headers []string) []string {
-	userAgentHeader := fmt.Sprintf("User-Agent:decK/%s", VERSION)
-	headers = append(headers, userAgentHeader)
+type header struct {
+	name  string
+	value string
+}
+
+func extendHeaders(headers []string, extra ...header) []string {
+	headers = append(headers, fmt.Sprintf("User-Agent:decK/%s", VERSION))
```
harshadixit12: What happens if we pass both `--kong-admin-token xyz`, and `--headers "kong-admin-token: xyz"` in same command? Do we prefer one over the other?
have we verified it works with a config file ([ref](https://support.konghq.com/support/s/article/Using-the-deck-configuration-file))?

Also, `konnect-token` seems to support input using a file with flag ` --konnect-token-file ` - is that something we want for kong admin token as well?

VarunAthreya: Adding code to give preference to `--kong-admin-token`, since the user has explicitly provided the token value.

cmd/root.go
```diff
@@ -451,10 +462,39 @@ func initKonnectConfig() error {
...
+
+	for _, h := range extra {
+		if h.value != "" {
+			result = append(result, fmt.Sprintf("%s:%s", h.name, h.value))
+		}
+	}
+	return result
+}
+
+// reports whether the given header name collides with an explicit
```
harshadixit12: let's complete this sentence?

cmd/root.go
```diff
@@ -451,10 +462,39 @@ func initKonnectConfig() error {
...
+	value string
+}
+
+func extendHeaders(headers []string, extra ...header) []string {
+	result := make([]string, 0, len(headers)+len(extra)+1)
+	result = append(result, fmt.Sprintf("User-Agent:decK/%s", VERSION))
+
+	for _, h := range headers {
+		name, _, _ := strings.Cut(h, ":")
+		if !overriddenBy(name, extra) {
```
harshadixit12, line 443: Let's add one comment to explain why we are doing this

.github/workflows/integration-enterprise.yaml
```diff
@@ -72,3 +72,61 @@ jobs:
...
         continue-on-error: ${{ matrix.kong_image == 'kong/kong-gateway-dev:latest' }}
+
+  integration-rbac:
+    # Skip if the PR has the 'skip-ee' label
+    if: ${{ github.event_name == 'push' || !contains(github.event.pull_request.labels.*.name, 'skip-ee') }}
+    timeout-minutes: ${{ fromJSON(vars.DECK_WORKFLOW_GW_TIMEOUT || '10') }}
+    strategy:
+      matrix:
+        kong_image:
+        - 'kong/kong-gateway:2.8'
```
harshadixit12, line 83: @VarunAthreya lets stick to LTS versions of kong gateway, and latest.

harshadixit12: 2.8, 3.4, 3.10 and 3.14

harshadixit12 (dismissed):

tests/integration/rbac_admin_token_test.go
```diff
@@ -0,0 +1,79 @@
...
+		// the request reaches Kong unauthenticated.
+		t.Setenv("DECK_KONG_ADMIN_TOKEN", "")
+
+		err := validate(ONLINE, stateFile)
+		require.Error(t, err,
+			"online validate should fail against an RBAC-enabled Kong without an admin token")
+		// Assert it fails *specifically* because of authentication (HTTP 401)
+		// rather than some unrelated error (bad file, gateway down, etc.).
+		// go-kong formats API errors as `HTTP status 401 (message: ...)`.
+		assert.Contains(t, err.Error(), "401",
```
harshadixit12: @VarunAthreya nitpick - since anyway you've to sync with main through merge / rebase, can you add one more commit to use `require` rather than assert?

VarunAthreya: done

harshadixit12 (approved):

PR #2137, opened 2026-07-08
Feat: ai gateway file convert

harshadixit12:
## Summary
In this PR, we are adding support for a new command `deck file ai2kong` which accepts a AI gateway declarative config, and converts it to kong 3.x format declarative config to be used with on prem AI Gateway.

## Implements
https://github.com/Kong/deck/issues/2148


cmd/file_ai2kong.go
```diff
@@ -0,0 +1,113 @@
...
+		OutputMode: "deck",
+	})
+	if err != nil {
+		return fmt.Errorf("conversion failed: %w", err)
+	}
+
+	// Print warnings to stderr
+	if len(warnings) > 0 {
+		for _, warning := range warnings {
+			fmt.Fprintf(os.Stderr, "Warning: %v\n", warning)
```
Prashansa-K: This would break json output, right?
Are we concerned about that right now?

cmd/file_ai2kong.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+func execute(_ *cobra.Command, _ []string) error {
+	// Read source file
+	sourceContent, err := os.ReadFile(convertSourceFile)
+	if err != nil {
+		return fmt.Errorf("failed to read source file: %w", err)
+	}
+
+	converted, warnings, err := convert.Convert(sourceContent, convert.Options{
```
Prashansa-K: nitpick: Can we import this as ai2kong and call `ai2kong.Convert()`?
It will be more verbose and avoid the confusing repetition.

cmd/file_ai2kong.go
```diff
@@ -0,0 +1,113 @@
...
+
+	cmd.Flags().StringVarP(&convertSourceFile, "source", "s", "", "AI Gateway source file (required)")
+	cmd.Flags().StringVarP(&convertOutputFile, "output-file", "o", "",
+		"Output Kong decK YAML file (optional, defaults to stdout)")
+
+	return cmd
+}
+
+func validateAi2KongFlags(_ *cobra.Command, _ []string) error {
+	if convertSourceFile == "" {
```
Prashansa-K, line 45: We can take only one file at a time for conversion, right?
If yes, can we add a validation for that too.

cmd/file_ai2kong.go
```diff
@@ -0,0 +1,113 @@
...
+	err = yaml.Unmarshal(converted, &doc)
+	if err != nil {
+		return fmt.Errorf("failed to parse converted YAML: %w", err)
+	}
+
+	// Ensure the document is a map and add select_tags to _info
+	if docMap, ok := doc.(map[string]interface{}); ok {
+		if infoMap, ok := docMap["_info"].(map[string]interface{}); ok {
+			// _info exists, update select_tags
+			infoMap["select_tags"] = []string{"managed-by:deck-ai"}
```
Prashansa-K: This can over-write the existing select-tags.

cmd/file_ai2kong.go
```diff
@@ -0,0 +1,113 @@
...
+		} else {
+			// _info doesn't exist, create it with select_tags
+			docMap["_info"] = map[string]interface{}{
+				"select_tags": []string{"managed-by:deck-ai"},
```
Prashansa-K: 1. `managed-by:deck-ai` should be a constant
2. Elsewhere, the convention I have seen is `managed_by`. Should we follow that or is this one a request from product?

cmd/file_ai2kong.go
```diff
@@ -0,0 +1,113 @@
...
+var (
+	convertSourceFile string
+	convertOutputFile string
+)
+
+func newAi2KongCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "ai2kong",
+		Short:   "Generate Kong configuration from AI Gateway configuration",
+		Long:    `This command takes an AI Gateway 2.0 entity model and converts it to a standard decK state file`,
```
Prashansa-K: Since we are adding a tag ourselves for our internal working, I think we should explicitly call that out in the description. It will help endusers to know that the tag addition is expected and they shouldn't remove it.

cmd/file_ai2kong.go
Prashansa-K: I see that we are using YAML everywhere for now. We can revisit this later. Let's add a TODO comment on top so that we don't miss this when we add JSON support.

cmd/file_ai2kong.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// Print warnings to stderr
+	if len(warnings) > 0 {
+		for _, warning := range warnings {
+			fmt.Fprintf(os.Stderr, "Warning: %v\n", warning)
+		}
+	}
+
+	// Parse the converted YAML to add select_tags to _info
+	var doc interface{}
```
Prashansa-K: nit: I feel L64 to L81 can be abstracted out in a different function.

cmd/file_ai2kong.go
```diff
@@ -0,0 +1,113 @@
...
+	var outputWriter io.Writer
+	if convertOutputFile != "" {
+		outFile, err := os.Create(convertOutputFile)
+		if err != nil {
+			return fmt.Errorf("failed to create output file: %w", err)
+		}
+		defer outFile.Close()
+		outputWriter = outFile
+	} else {
+		outputWriter = os.Stdout
```
Prashansa-K: This can be the default IMO rather than adding a nesting here with `else`.

cmd/file_ai2kong.go
```diff
@@ -0,0 +1,113 @@
...
+	_, err = outputWriter.Write(output)
+	if err != nil {
+		return fmt.Errorf("failed to write output: %w", err)
+	}
+
+	return nil
+}
+
+// marshalToYAML encodes v as YAML using a two-space indent.
+func marshalToYAML(v interface{}) ([]byte, error) {
```
Prashansa-K: Any reason for a separate func for this?

tests/integration/file_ai2kong_test.go
```diff
@@ -0,0 +1,68 @@
...
+			content, err := os.ReadFile(tc.expectedOutputFile)
+			require.NoError(t, err)
+
+			err = yaml.Unmarshal(content, &expectedOutput)
+			require.NoError(t, err)
+
+			err = yaml.Unmarshal([]byte(output), &currentOutput)
+			require.NoError(t, err)
+
+			assert.Equal(t, expectedOutput, currentOutput)
```
Prashansa-K, line 65: Should we consider syncing the output to an ai gateway for validity?
We need to support the decomposed flow as well.

cmd/file_ai2kong.go
harshadixit12: Ideally - yes, we should - but i'll get to this once all PRs have been merged.

cmd/file_ai2kong.go
harshadixit12: This is converting an ai gateway file -> deck, the ai gateway file won't have select tags at global level

cmd/file_ai2kong.go
harshadixit12: Will be made constant - this was a suggestion, not a request. I'll update it to `managed_by`.

tests/integration/file_ai2kong_test.go
harshadixit12: I was planning to do that in the tests for sync and dump in the chained PR

cmd/file_ai2kong.go
harshadixit12: For flag type of `string` - cobra follows "last value" wins - so additional validation is not needed. (same in openapi2kong)

cmd/file_ai2kong.go
```diff
@@ -0,0 +1,124 @@
...
+
+	ai2kong "github.com/Kong/ai-deck-converter/convert"
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/yaml"
+)
+
+var (
+	convertSourceFile string
+	convertOutputFile string
+	managedByTag      = "managed_by:deck-ai"
```
Prashansa-K: const

harshadixit12: renamed this in the chained PR and made it a const - https://github.com/Kong/deck/pull/2143/changes/25fdbde97881e35f8ba2e8c34868ce4d7b3c07cd

Prashansa-K (dismissed):

harshadixit12:
@Prashansa-K need review on 3 new commits - 
https://github.com/Kong/deck/pull/2137/commits/f7f58413f7f9a8adc7d5dfee455e294eba9cff90
https://github.com/Kong/deck/pull/2137/commits/a68ee040d02aa49131fe1a108222edd75d6915db
https://github.com/Kong/deck/pull/2137/commits/9759b05227d1fdc1ae565d807978623c56443f9f

Prashansa-K (approved):

PR #2143, opened 2026-07-10
feat: support ai dump and sync

harshadixit12:
## Summary 
In this PR, we are supporting two commands - `deck ai sync` and `deck ai dump` - which will accept, and return declarative config containing the AI Gateway entities - and internally convert to Kong 3.x format and sync with on-prem AI Gateway.

Depends on -
https://github.com/Kong/go-database-reconciler/pull/496
https://github.com/Kong/go-kong/pull/624


harshadixit12:
@Prashansa-K - The UX isnt perfect - I'm doing a self review and will add 1-2 more commits. Opened for review so that I could get another set of eyes on the integration tests for ai sync and dump - using the inputs and outputs from file_ai2kong itself.

cmd/ai.go
```diff
@@ -0,0 +1,112 @@
...
+		Use:   "ai [sub-command]...",
+		Short: "Subcommand to host decK AI Gateway operations",
+		Long:  `Subcommand to host decK AI Gateway operations.`,
+	}
+
+	return aiSubCmd
+}
+
+// errAIManagedEntitiesOnKonnect is returned when a user attempts to sync/apply
+// AI Gateway entities (tagged managed-by: deck-ai) to Konnect using decK. Those
```
harshadixit12: noted updating tags in comments, will be done in next commit.

cmd/ai.go
```diff
@@ -0,0 +1,112 @@
...
+// entities must be managed with kongctl instead.
+func errAIManagedEntitiesOnKonnect() error {
```
harshadixit12, line 44: This is to be applied for deck ai dump as well - tbd in next commit.

cmd/ai_dump.go
```diff
@@ -0,0 +1,175 @@
...
+
+	if yes, err := utils.ConfirmFileOverwrite(aiDumpCmdOutputFile, aiDumpCmdStateFormat, assumeYes); err != nil {
+		return err
+	} else if !yes {
+		return nil
+	}
+
+	// Set the selector tags to only get AI-managed entities
+	originalTags := dumpConfig.SelectorTags
+	dumpConfig.SelectorTags = []string{managedByAIDeckTag}
```
harshadixit12, line 44: this looks suspiciously like overwriting tags - will verify.

cmd/ai_dump.go
```diff
@@ -0,0 +1,175 @@
...
+
+	// Parse the AI Gateway YAML to add select_tags to _info
+	var doc interface{}
+	err = yaml.Unmarshal(aiGatewayYAML, &doc)
+	if err != nil {
+		return fmt.Errorf("failed to parse AI Gateway YAML: %w", err)
+	}
+
+	// Ensure the document is a map and add select_tags to _info
+	if docMap, ok := doc.(map[string]interface{}); ok {
```
harshadixit12: this is not required - _info.select_tags does not exist in the world of AI gateway entities, will remove.

cmd/ai_dump.go
```diff
@@ -0,0 +1,175 @@
...
+	isAIGateway, err := kong.IsKongAIGateway()
+	if err != nil {
+		return fmt.Errorf("checking if Kong is an AI Gateway: %w", err)
+	}
+
+	writeConfig := file.WriteConfig{
+		SelectTags:                       dumpConfig.SelectorTags,
+		Workspace:                        aiDumpWorkspace,
+		Filename:                         "",
+		WithID:                           false,
```
harshadixit12, line 65: revert layer of ai-deck-converter does not support retaining IDs

cmd/ai_sync.go
```diff
@@ -0,0 +1,137 @@
...
+		return
+	}
+	content.Info.SelectorTags = append(content.Info.SelectorTags, managedByAIDeckTag)
+}
+
+func newAiSyncCmd() *cobra.Command {
+	aiSyncCmd := &cobra.Command{
+		Use:   "sync [flags] [ai-gateway-state-file]",
+		Short: "Sync AI Gateway configuration to Kong",
+		Long: `The ai sync command converts an AI Gateway configuration file to Kong
```
harshadixit12: This can be simplified - we need not expose the internal mechanism of converting to kong 3.x in the description.

cmd/ai_sync.go
```diff
@@ -0,0 +1,137 @@
...
+// or from stdin when filename is "-".
+func readAiSyncSource(filename string) ([]byte, error) {
+	if filename == "-" {
+		content, err := io.ReadAll(os.Stdin)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read from stdin: %w", err)
+		}
+		return content, nil
+	}
+	content, err := os.ReadFile(filename)
```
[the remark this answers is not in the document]
Prashansa-K: @AikidoSec ignore: CLI functionality requires reading of files and working on them. This is a valid usecase and not a malicious behaviour.

cmd/ai_dump.go
```diff
@@ -0,0 +1,156 @@
...
+		KongVersion:                      kongVersion,
+		IsKongAIGateway:                  isAIGateway,
+		IsConsumerGroupPolicyOverrideSet: dumpConfig.IsConsumerGroupPolicyOverrideSet,
+		SanitizeContent:                  false,
+		IncludePluginDefinitions:         dumpConfig.IncludePluginDefinitions,
```
Prashansa-K: This should also be explicitly set as false, right?
Same for `IsConsumerGroupPolicyOverrideSet`

tests/integration/test_utils.go
```diff
@@ -365,6 +375,43 @@ func sync(ctx context.Context, kongFile string, opts ...string) error {
 	return deckCmd.ExecuteContext(ctx)
 }
 
+// aiSync converts an AI Gateway source file to Kong configuration and syncs it
+// directly to Kong (the equivalent of `deck ai sync <source-file>`). It runs
+// non-interactively so it can be used in tests.
+func aiSync(ctx context.Context, sourceFile string, opts ...string) error {
```
Prashansa-K, line 394: Are aiming for multi file sync in this PR?

tests/integration/ai_dump_test.go
```diff
@@ -0,0 +1,145 @@
...
+// entity: after seeding Kong with ai_models, the dumped file must contain the
+// same entities. ai_model is only available on AI Gateway instances.
+func Test_Dump_AIModels(t *testing.T) {
+	runWhenAIGateway(t, ">=2.0.0")
+	setup(t)
+
+	ctx := context.Background()
+
+	// ID and CreatedAt/UpdatedAt are server-assigned, so they are ignored.
+	ignoreFields := []cmp.Option{
```
Prashansa-K: Rename to cmpOpts as this has more than just ignoreFields.

tests/integration/ai_sync_test.go
```diff
@@ -0,0 +1,199 @@
...
+	return &content
+}
+
+// assertAIStateEqual asserts that two dumped AI-managed states are equivalent.
+//
+// Plugins are sorted by their content before comparing: `deck gateway dump`
+// orders plugins by name and foreign-key ID (see FPlugin.sortKey in
+// go-database-reconciler), so entities that share a name and route - such as
+// the two ai-proxy-advanced plugins bound to one route - are emitted in a
+// server-ID-dependent order that differs between independent syncs. A KeyAuth
```
Prashansa-K: Can this create unnecessary diffs for our endusers?

tests/integration/ai_sync_test.go
```diff
@@ -0,0 +1,199 @@
...
+			// re-sync with no error
+			err = sync(ctx, tc.kongFile)
+			require.NoError(t, err)
+		})
+	}
+}
+
+// managedByAIDeckTag is the selector tag `deck ai sync` stamps on every entity it
+// manages, mirroring the scope of `deck ai dump`.
+const managedByAIDeckTag = "managed_by:deck-ai"
```
Prashansa-K: L80 to L124 should be moved to utils file.
If test_utils feels too crowded, a separate one just for ai_gw is fine too.

cmd/common.go
Prashansa-K, line 57: Adding a new mode here like modeAIGateway and setting it when using ai commands will help us with telemetry and maybe some internal work later. Worth looking into.

cmd/common.go
```diff
@@ -231,6 +249,12 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 	isKonnect := false
 	workspaceName := getWorkspaceName(workspace, targetContent, enableJSONOutput)
 
+	// AI Gateway entities (tagged managed_by:deck-ai) must be managed with
+	// kongctl on Konnect, not decK. Fail fast before any Konnect calls.
+	if mode == modeKonnect && contentHasmanagedByAIDeckTag(targetContent) {
```
Prashansa-K: This can be moved inside the following if condition and checked before anything else.

tests/integration/test_utils.go
harshadixit12: no - will send a separate PR

cmd/ai_dump.go
harshadixit12: yes, will do

tests/integration/ai_sync_test.go
harshadixit12: verifying

cmd/common.go
harshadixit12: good idea, will look into it

tests/integration/ai_sync_test.go
harshadixit12: Shouldn't - we already use sorting before comparing state for plugins in the integration tests ([src](https://github.com/Kong/deck/blob/main/tests/integration/test_utils.go#L175)) - it was missing `model` as part of its sort key
This needed a new function with own sorting logic because it is not comparing against a static kong state - let me check if this duplication can be resolved.

cmd/common.go
harshadixit12: Done - have added a new mode, and sent relevant mode in sync, dump and ai2kong

tests/integration/ai_sync_test.go
harshadixit12: Gave it some more thought - existing way of testing kong state in tests does this - creates a dump, and then compares with hardcoded kong state - but here we are comparing two dumps which have happened - and not hardcoded state.
Because these dumps are from different syncs - the IDs will be different, hence the custom comparison by parsing into `file.Content`.

cmd/ai.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+	case reflect.Struct:
+		t := v.Type()
+		for i := 0; i < v.NumField(); i++ {
+			field := t.Field(i)
+			if !field.IsExported() {
+				continue
+			}
+			fieldValue := v.Field(i)
+			if field.Name == "Tags" {
```
Prashansa-K: nit: "Tags" should be a const

cmd/ai_dump.go
Prashansa-K: Was this verified?

cmd/ai_dump.go
```diff
@@ -0,0 +1,156 @@
...
+	// kongctl on Konnect, not decK. Since ai dump only ever reads AI-managed
+	// entities, fail fast before any Konnect calls.
+	if inKonnectMode(nil) {
+		return errAIManagedEntitiesOnKonnect()
+	}
+
+	// Set the selector tags to only get AI-managed entities
+	originalTags := dumpConfig.SelectorTags
+	dumpConfig.SelectorTags = []string{managedByAIDeckTag}
+	defer func() {
```
Prashansa-K: 🤔 I didn't get this

cmd/ai_dump.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	// Convert Kong state to YAML in memory
+	kongYAML, err := kongsStateToYAML(ks, writeConfig)
+	if err != nil {
+		return fmt.Errorf("converting Kong state to YAML: %w", err)
+	}
+
+	// Revert Kong YAML back to AI Gateway format
+	aiGatewayYAML, warnings, err := revert.Revert(kongYAML, revert.Options{})
```
Prashansa-K: nit: rename revert.Revert to ai2kong.Revert()

cmd/ai_dump.go
```diff
@@ -0,0 +1,156 @@
...
+	var output io.Writer
+	if aiDumpCmdOutputFile != "" && aiDumpCmdOutputFile != "-" {
+		outFile, err := os.Create(aiDumpCmdOutputFile)
+		if err != nil {
+			return fmt.Errorf("failed to create output file: %w", err)
+		}
+		defer outFile.Close()
+		output = outFile
+	} else {
+		output = os.Stdout
```
Prashansa-K: Default should be os.Stdout, right?
I prefer as less nesting as possible. It helps with code readability.

cmd/ai_dump.go
harshadixit12: yes - its effectively a no-op. let me just remove.

cmd/ai_sync.go
```diff
@@ -0,0 +1,136 @@
...
+	if aiSyncJSONOutput {
+		initJSONOutput()
+	}
+
+	sourceContent, err := readAiSyncSource(aiSyncSourceFile)
+	if err != nil {
+		return err
+	}
+
+	convertedYAML, warnings, err := convert.Convert(sourceContent, convert.Options{})
```
Prashansa-K: convert.Convert -> ai2kong.Convert()

cmd/ai_dump.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	// Revert Kong YAML back to AI Gateway format
+	aiGatewayYAML, warnings, err := revert.Revert(kongYAML, revert.Options{})
+	if err != nil {
+		return fmt.Errorf("reverting Kong configuration to AI Gateway format: %w", err)
+	}
+
+	// Print warnings to stderr
+	if len(warnings) > 0 {
```
Prashansa-K: This should also use something like `reportAIConversionWarnings` function, right? That will handle json as well.
Ignore if this is dealt with in the final PR.

cmd/ai_sync.go
```diff
@@ -0,0 +1,136 @@
...
+
+	aiSyncCmd.Flags().StringVarP(&aiSyncWorkspace, "workspace", "w", "",
+		"Sync configuration to a specific workspace "+
+			"(Kong Enterprise only).\n"+
+			"This takes precedence over _workspace fields in state files.")
+	aiSyncCmd.Flags().IntVar(&aiSyncParallelism, "parallelism",
+		10, "Maximum number of concurrent operations.")
+	aiSyncCmd.Flags().IntVar(&aiSyncDBUpdateDelay, "db-update-propagation-delay",
+		0, "artificial delay (in seconds) that is injected between insert operations \n"+
+			"for related entities (usually for Cassandra deployments).\n"+
```
Prashansa-K: Does AI gateway support Cassandra deployments?

cmd/ai_dump.go
```diff
@@ -0,0 +1,156 @@
...
+		Short: "Dump AI Gateway configuration from Kong (managed by deck-ai)",
+		Long: `The ai dump command reads AI Gateway entities (tagged with 'managed_by:deck-ai')
+from Kong and writes them to a local file in AI Gateway format.
+
+The command exports only entities that have the 'managed_by:deck-ai' tag.`,
+		Args: validateNoArgs,
+		RunE: executeAiDump,
+	}
+
+	aiDumpCmd.Flags().StringVarP(&aiDumpWorkspace, "workspace", "w",
```
Prashansa-K, line 172: Do we have a requirement for dumpAllWorkspaces too?

cmd/ai_test.go
```diff
@@ -0,0 +1,99 @@
...
+					{Service: kong.Service{
+						Name: kong.String("s1"),
+						Tags: kong.StringSlice(otherTag, managedByAIDeckTag),
+					}},
+				},
+			},
+			want: true,
+		},
+		{
+			name: "tag on a nested route under a service",
```
Prashansa-K, line 110: 1. Can we check for more levels of nesting too?
2. Can we check for a nested entity >1 tag with and without ai_tag?

cmd/common.go
```diff
@@ -232,6 +265,12 @@ func syncMain(ctx context.Context, filenames []string, dry bool, parallelism,
 	workspaceName := getWorkspaceName(workspace, targetContent, enableJSONOutput)
 
 	if mode == modeKonnect {
+		// AI Gateway entities (tagged managed_by:deck-ai) must be managed with
+		// kongctl on Konnect, not decK. Fail fast before any Konnect calls.
+		if contentHasmanagedByAIDeckTag(targetContent) {
```
Prashansa-K, line 270: For every sync, diff, apply that we do for Konnect, we will walk the entire configuration reflection tree. :mild_panic: 

If this is truly necessary to do, can we optimize `contentHasmanagedByAIDeckTag`?

cmd/ai.go
```diff
@@ -0,0 +1,112 @@
...
+			return false
+		}
+		return valueHasTag(v.Elem())
+	case reflect.Slice, reflect.Array:
+		for i := 0; i < v.Len(); i++ {
+			if valueHasTag(v.Index(i)) {
+				return true
+			}
+		}
+	case reflect.Map:
```
Prashansa-K: Check for a configMap for any plugin will go through each of the fields, right? That can be expensive IMO.

Let's see if we can add exemptions on certain Types like `Configuration` where we know that tags won't live + the size is dynamic.

Also, do we need to check all entities? Like keys, vaults, certs, custom entities? Can't we exempt them by default?

cmd/ai.go
```diff
@@ -0,0 +1,112 @@
...
+		return true
+	}
+	return valueHasTag(reflect.ValueOf(content))
+}
+
+// valueHasTag recursively walks v looking for any exported field named "Tags"
+// (a []*string, as used by every Kong entity) that contains managedByAIDeckTag.
+// Non-Tags fields are traversed structurally but their string values are never
+// matched, so plugin config values or names cannot trigger a false positive.
+func valueHasTag(v reflect.Value) bool {
```
Prashansa-K: If it's possible to limit this to a bunch of fields which AI gateway supports, it would be better and faster for us. This function lies in a critical path.

IIUC, AI gateway supports services, routes, plugins, ai_models, right? Can we limit the checks to just those? It will be cheaper and won't require extra allocations.

cmd/ai.go
```diff
@@ -0,0 +1,112 @@
...
+			if valueHasTag(fieldValue) {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// tagsContain reports whether a []*string Tags field contains managedByAIDeckTag.
+func tagsContain(tags reflect.Value) bool {
```
Prashansa-K: Can we rename this better to indicate what exactly we are searching for?

cmd/common.go
harshadixit12: I think there should be a faster way to do this, let me check.

cmd/ai_sync.go
harshadixit12: Not explicitly asked - but was kept with the intent of being as close to `gateway sync` as possible. On second thought maybe unnecessary - I'll remove.

cmd/ai_dump.go
harshadixit12: No explicit requirement

cmd/ai.go
harshadixit12: we don't need to check the entities - limited to those supported by AI gateway.

Prashansa-K (approved):

PR #2149, opened 2026-07-13
test: added gw 3.15 in test matrix

Prashansa-K:
- Added 3.15 test coverage.
- Some tests were flaky as plugin definitions took time to load in the Kong instance. If a plugin with those definitions was created in the while, the tests would fail. The changes in this PR ensure that plugin definitions and plugins are created in separate syncs and an artificial delay is added in between. This is done to reduce flakiness so that we don't have to re-run pipelines unnecessarily.


shivaygupta-dotcom:
@Prashansa-K Is it possible for a plugin instance to be created and then referenced almost immediately in a real-world scenario? If so, could this become a breaking case?
Or am I missing some context or assumption here?

Prashansa-K:
> @Prashansa-K Is it possible for a plugin instance to be created and then referenced almost immediately in a real-world scenario? If so, could this become a breaking case? Or am I missing some context or assumption here?

This isn't about plugins. This is about plugin definitions that are loaded in the gateway, which may take some time. Plugins require schemas to be created which comes from the definitions.

Not a breaking case. It's a test change.

shivaygupta-dotcom (approved):

PR #2153, opened 2026-07-14
Feat/support deck ai sync

harshadixit12:


go.work
```diff
@@ -0,0 +1,8 @@
+go 1.26.4
```
Prashansa-K, line 1: Let's remove this and go.work.sum

Prashansa-K: Feel free to add them to git ignore

harshadixit12: yep yep - will do

harshadixit12:
Closing this since the parent PR is quite small, and makes more sense to test together.

PR #2164, opened 2026-07-21
feat(docker): switch to distroless

zekth:
To avoid having security issues imported from many deps that alpine pulls, switching the docker image to distroless.


Prashansa-K:
@harshadixit12 I have added a few changes to the original PR here. Please review. 
Here's a gist of my changes:
- Updated the runtime base image from gcr.io/distroless/base to gcr.io/distroless/static-debian12:nonroot. `distroless/base` had pushed the final image size to 104 MB (36% larger than our current alpine based one). Both deck (CGO_ENABLED=0) and the bundled jq are statically linked, so the glibc base isn't needed. Moving to `distroless/static` brings the image down to 70.7 MB while keeping non-root and jq intact.
- I also set the HOME env variable as deck resolves $HOME on startup, and distroless has no shell/getent fallback. Adding a HOME explicitly here ensures that the binary doesn't panic in case a runtime leaves $HOME unset. This is less likely to happen but I erred on the side of caution.

Verified locally with podman: deck runs, jq works and the container runs as nonroot.

zekth:
> had pushed the final image size to 104 MB (36% larger than our current alpine based one)

thats not great :(

harshadixit12 (approved):

PR #2169, opened 2026-07-28
fix: use the workspace client for AI Gateway check

harshadixit12:
We used `kong.IsAIGateway()` function to determine whether the instance deck is interacting with an AI-gateway, however, it uses the default client and does not respect flags like --kong-addr and `--kong-admin-token`.

Instead, we are now using a wrapper over `client.Server()` function, and passing the same client we use within dump and sync.

How I have tested this - 
1. Integration test on an RBAC enabled gateway in CI ✅ 
2. Test `gateway sync/dump`, `ai sync/dump` on local where kong is deployed on a non-default port (9001) ✅ 
3. `deck ai sync/dump` with on AI Gateway instance with RBAC enabled ✅ 
4. Verified the other common flows we touch for ai-gateway support in https://github.com/Kong/deck/pull/2137 ✅

### DIFF
--- a/tests/integration/ai_dump_test.go
+++ b/tests/integration/ai_dump_test.go
@@ -12,7 +12,6 @@ import (
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/kong/go-database-reconciler/pkg/file"
 	"github.com/kong/go-kong/kong"
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
@@ -123,11 +122,6 @@ func Test_AIDump(t *testing.T) {
 			aiConfig, err := aiDump("-o", "-")
 			require.NoError(t, err)
 			require.NotEmpty(t, aiConfig)
-
-			// Dumping the same state again must be deterministic.
-			aiConfigAgain, err := aiDump("-o", "-")
-			require.NoError(t, err)
-			assert.Equal(t, aiConfig, aiConfigAgain)
 
 			// Round-trip: syncing the dumped AI Gateway config into a fresh Kong
 			// must reproduce the same AI-managed state.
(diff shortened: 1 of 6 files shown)


shivaygupta-dotcom (dismissed):

shivaygupta-dotcom:
@harshadixit12 Just trying to understand this a bit better. Is there any reason why this wasn’t failing earlier in CI or being caught by our existing tests? I may be missing some context here, so I’d really appreciate any insights.

tests/integration/ai_dump_test.go
```diff
@@ -124,11 +123,6 @@ func Test_AIDump(t *testing.T) {
 			require.NoError(t, err)
 			require.NotEmpty(t, aiConfig)
 
-			// Dumping the same state again must be deterministic.
-			aiConfigAgain, err := aiDump("-o", "-")
-			require.NoError(t, err)
-			assert.Equal(t, aiConfig, aiConfigAgain)
```
harshadixit12, line 130: @shivaygupta-dotcom this is why the test was flaky - we were doing string comparison of dump output - but the value of `KeyAuth.TTL` changes between two dumps - and the string comparison does not ignore it.
In line 136 we are anyway comparing the state (which ignores `TTL`) - so removing this block as it is redundant.

harshadixit12:
> @harshadixit12 Just trying to understand this a bit better. Is there any reason why this wasn’t failing earlier in CI or being caught by our existing tests? I may be missing some context here, so I’d really appreciate any insights.

In CI - we always use the default kong address, and the tests for RBAC enabled instance did not cover `deck gateway dump`

shivaygupta-dotcom (approved):

tests/integration/rbac_admin_token_test.go
```diff
@@ -75,4 +75,18 @@ func Test_RBAC_AdminToken(t *testing.T) {
 			"online validate should succeed: --kong-admin-token must override the "+
 				"colliding Kong-Admin-Token supplied via --headers")
 	})
+
+	t.Run("dump succeeds when kong-admin-token is passed", func(t *testing.T) {
```
harshadixit12, line 79: Tracking further additions to RBAC tests here - https://github.com/Kong/deck/issues/2170

Prashansa-K (approved):

PR #2184, opened 2026-08-10
feat(tests): add tests for handling empty array plugin configurations

VarunAthreya:
Add integration tests to verify decK correctly handles plugin configurations with explicitly empty arrays, preventing perpetual diffs that never converge. Tests cover various scenarios top-level and nested empty arrays, enterprise plugins, and string arrays. Also validates that genuine changes from empty to populated arrays are still detected, and that omitted fields match empty arrays.

An unintended side effect was discovered and has been documented here: #2188

### DIFF
--- a/tests/integration/diff_test.go
+++ b/tests/integration/diff_test.go
@@ -1072,17 +1072,17 @@ func Test_Diff_EmptyArrayPluginConfig_NoPerpetualDiff(t *testing.T) {
 
 			if isKonnect {
 				runDualTestWithSkipDefaults(t, tc.name, func(t *testing.T) {
-					testDiffEmptyArrayPluginConfigNoPerpetualDiffImpl(t, ctx, tc.stateFile)
+					testDiffEmptyArrayPluginConfigNoPerpetualDiffImpl(ctx, t, tc.stateFile)
 				})
 			} else {
 				// for enterprise run once
-				testDiffEmptyArrayPluginConfigNoPerpetualDiffImpl(t, ctx, tc.stateFile)
+				testDiffEmptyArrayPluginConfigNoPerpetualDiffImpl(ctx, t, tc.stateFile)
 			}
 		})
 	}
 }
 
-func testDiffEmptyArrayPluginConfigNoPerpetualDiffImpl(t *testing.T, ctx context.Context, stateFile string) {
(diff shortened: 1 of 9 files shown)


tests/integration/diff_test.go
```diff
@@ -1027,6 +1027,111 @@ func Test_Diff_PluginConfigReorderArraySetValues(t *testing.T) {
...
+	tests := []struct {
+		name      string
+		stateFile string
+		runWhen   func(t *testing.T)
+	}{
+		{
+			// the fixtures set plugin instance_name, which Kong only supports from 3.2.0.
+			name:      "top-level empty array field",
+			stateFile: "testdata/diff/010-empty-array-plugin-config/kong.yaml",
+			runWhen:   func(t *testing.T) { runWhenKongOrKonnect(t, ">=3.2.0") },
```
Prashansa-K, line 1044: Can you test this in dual mode for Konnect please? 
We can have two test functions if needed - for only Kong and then only Konnect with dual modes. Refer other test files to see how we test against Konnect - with skip defaults and without.

Prashansa-K: Same for other tests

VarunAthreya: done

tests/integration/diff_test.go
```diff
@@ -1119,6 +1148,15 @@ func Test_Diff_EmptyArrayPluginConfig_RealChangeStillDetected(t *testing.T) {
 
 // Test_Diff_EmptyArrayPluginConfig_OmittedFieldMatchesEmptyArray pins down that omitting an array field entirely
 // is indistinguishable from setting it to [], since decK fills both sides with nil before comparing.
+//
+// NOTE: Deliberately NOT wrapped in runDualTestWithSkipDefaults, unlike the two tests above. That helper runs the same
```
VarunAthreya, line 1152: @Prashansa-K reason for not adding dual test here

should we go ahead with this solution and create a issue to work on this edge case in future or address this now?

Prashansa-K (approved):

tests/integration/diff_test.go
VarunAthreya: created a issue regarding this:
https://github.com/Kong/deck/issues/2188

PR #2194, opened 2026-08-16
Fix/dump with degraphql referring different workspace

### LINKED ISSUE #2180: `gateway dump --all-workspaces` fails for globally listed DeGraphQL routes with cross-workspace parent Services
## `gateway dump --all-workspaces` fails for globally listed DeGraphQL routes with cross-workspace parent Services

  ### Summary

  `deck gateway dump --all-workspaces` fails when Kong returns `degraphql_routes`
  from a global collection while decK is building a different workspace from the
  one containing the route’s parent Service.

  The DeGraphQL route and its parent Service are valid. The failure occurs because
  decK tries to resolve the parent Service only in the workspace currently being
  exported.

  ### Environment

  - Kong Gateway Enterprise: `3.4.3.28`
  - decK: `1.64.0`
  - Database-backed Gateway with multiple workspaces

  ### Actual behaviour

  An all-workspaces dump fails with:

  ```text
  Error: getting Kong state for workspace 'workspace-b':
  building state: building degraphql route from custom entity:
  service must be of type object with a valid string id or name
````

  The failing DeGraphQL entities are returned from the global
  /degraphql_routes collection. Their parent Service resolves successfully in
  workspace-a, but not in workspace-b, which is the workspace decK is
  currently processing.

  This is not an orphaned reference: resolving the parent Service across all
  workspaces finds it exactly once, in its owning workspace.

  ### Expected behaviour

  decK should support all-workspaces export when DeGraphQL routes are returned
  globally.

  Possible correct behaviours include:

  1. Associate each returned degraphql_routes entity with the workspace that
     contains its parent Service, then write it to that workspace’s export; or

  2. Exclude a DeGraphQL entity while exporting a workspace that does not contain
     its parent Service.

  decK should not fail the entire all-workspaces export for a valid
  cross-workspace reference.

  ### Minimal reproduction

  1. Create two workspaces: workspace-a and workspace-b.
  2. In workspace-a, create:
      - a Service, for example example-service;
      - a degraphql plugin on that Service;
      - one DeGraphQL route referencing that Service.

  3. Confirm that the DeGraphQL route is visible from the global
     /degraphql_routes Admin API collection.

  4. Run:

     deck gateway dump \
       --kong-addr "$KONG_ADMIN_URL" \
       --headers "Kong-Admin-Token:$KONG_ADMIN_TOKEN" \
       --all-workspaces \
       --yes

  5. Observe that decK fails when it processes workspace-b, because it receives
     the DeGraphQL entity from workspace-a but cannot resolve
     example-service in workspace-b.

  ### Additional context

  This differs from the tag-filtering problem in
  https://github.com/Kong/deck/issues/1609. No --select-tag filtering is used
  here. Tag-based exclusion is also not a workaround because DeGraphQL custom
  entities do not support tags:
  https://github.com/Kong/deck/issues/1564.

  Could you confirm whether this is a known Gateway Admin API scoping behaviour
  and whether decK should handle it during workspace and all-workspaces dumps?

### PR DESCRIPTION
**Pull request by harshadixit12**
### Summary
While building state for degraphql routes and cost decoration - we were ensuring that the service they are attached to is present in state. However, this leads to a problem - since these two custom entities are global, they are present in dump of every workspace - and any workspace not containing the service they are linked to ends up erroring out during dump - effectively blocking the users.

Fix: Relax the constraint to avoid erroring out if service is not in state, and instead print a warning instead - same as we do for plugin partials when not in state.

### Tests added
- Dump a different workspace when custom entities are configured

Fixes https://github.com/Kong/deck/issues/2180
GDR PR: https://github.com/Kong/go-database-reconciler/pull/511

### DIFF
--- a/tests/integration/dump_test.go
+++ b/tests/integration/dump_test.go
@@ -1351,6 +1351,43 @@ func Test_Dump_GraphqlRateLimitingCostDecorations_Konnect(t *testing.T) {
 	}
 }
 
+// test scope:
+//   - enterprise, workspaces
+func Test_Dump_DegraphqlRoute_CrossWorkspaceService(t *testing.T) {
+	runWhen(t, "enterprise", "=3.4.3.25 || =3.10.0.10 || =3.11.0.9 || =3.12.0.5 || =3.13.0.3"+
+		" || >=3.14.0.2")
+	setup(t)
+	t.Cleanup(func() {
+		reset(t, "--all-workspaces")
+	})
+
+	ctx := context.Background()
+
+	// workspace test-b: create an unrelated service. This runs before test-a is
+	// populated below: since degraphql_routes/cost decorations are global
+	// entities (see below), a `sync` against test-b done *after* they exist
+	// would fetch them as part of its "current state" and, not finding them in
+	// this file, delete them as part of its normal reconciliation.
+	require.NoError(t, sync(ctx,
+		"testdata/dump/015-degraphql-cross-workspace-service/workspace-b-service.yaml"))
+
+	// workspace test-a: create a service, plus a degraphql route and a graphql
+	// ratelimiting cost decoration that reference it.
+	require.NoError(t, sync(ctx,
+		"testdata/dump/015-degraphql-cross-workspace-service/workspace-a-initial.yaml"))
+
+	// degraphql_routes and graphql_ratelimiting_cost_decorations are global
+	// entities in Kong: they aren't scoped to the workspace they were created
+	// in, so dumping test-b also picks up test-a's route and decoration. Their
+	// "service" can't be resolved against test-b's own state, since that
+	// service lives in test-a. That must produce a warning, not an error.
+	output, err := dump("-o", "-", "--workspace", "test-b")
+	require.NoError(t, err)
+
+	assert.Contains(t, output, "1a111111-1111-4111-8111-111111111111",
+		"the cross-workspace service id should still be present in the dump")
+}
+
 func Test_Dump_KonnectWorkspace(t *testing.T) {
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: tests/integration/dump_test.go [hunk 1]
```diff
@@ -1351,6 +1351,43 @@ func Test_Dump_GraphqlRateLimitingCostDecorations_Konnect(t *testing.T) {
 	}
 }
 
+// test scope:
+//   - enterprise, workspaces
+func Test_Dump_DegraphqlRoute_CrossWorkspaceService(t *testing.T) {
+	runWhen(t, "enterprise", "=3.4.3.25 || =3.10.0.10 || =3.11.0.9 || =3.12.0.5 || =3.13.0.3"+
```
harshadixit12, line 1357: Not necessary for konnect - because the custom entities are strictly workspace scoped.

Ideally, we wouldn't need to use `||` in the gateway versions here - let me update that as well.

shivaygupta-dotcom: @harshadixit12 should we also add a test for cost decorator as well ?

harshadixit12: it is already there no?
https://github.com/Kong/deck/pull/2194/changes#diff-c0c2cc3bbaae9158019ae124ecd9f743a46f5d14a8c5f3162dc81af28e32fe89

shivaygupta-dotcom: ahh, I got confused with the name of test. Can we update test name accordingly pls ?

harshadixit12: will do

shivaygupta-dotcom (approved):

Issue #2203, opened 2026-08-26
deck gateway dump fails when the Kong Admin API omits the optional Server response header


pritho:
# Description

## Summary

`deck gateway dump` exits before exporting any entities when the Kong Admin API root response does not include a `Server` header, even though the API is healthy, authenticated, and returns a valid Kong version in its JSON body.

This prevents declarative backup/export in environments where Kong is configured not to emit server-identification headers or where an ingress/reverse proxy removes them. The HTTP `Server` response field is optional, so its absence should not make a normal Kong Gateway dump fail.

## Environment

- decK: `v1.65.2` (`d796fcd`)
- Also present in decK `v1.65.1` (`7fcd888`)
- Kong Gateway Enterprise: `3.4.3.28`
- Deployment: hybrid mode; request sent to the Control Plane Admin API
- Authentication: RBAC using `Kong-Admin-Token` (value redacted)
- Kong configuration: response-identification headers disabled (`headers=off`)
- Client OS: macOS

The same behavior was confirmed through both the external Admin API ingress and a direct connection to the Control Plane Admin service. Both returned a healthy Admin API response without a `Server` header.

## Steps to reproduce

1. Expose a Kong Admin API whose `GET /` response has no `Server` header. This can be done by disabling/removing server-identification headers in Kong or by stripping the header in a reverse proxy.

2. Confirm that the Admin API is healthy and returns its version while omitting the header:

   ```bash
   curl --silent --show-error \
     --dump-header /tmp/kong-admin-headers.txt \
     --output /tmp/kong-admin-root.json \
     --header "Kong-Admin-Token: ${KONG_ADMIN_TOKEN}" \
     "${KONG_ADMIN_URL}/"

   jq -r '.version' /tmp/kong-admin-root.json
   grep -i '^server:' /tmp/kong-admin-headers.txt
   ```

   The first command succeeds, `jq` prints the Kong version, and `grep` produces no output.

3. Confirm that decK can authenticate and reach the same Admin API:

   ```bash
   deck gateway ping --analytics=false \
     --kong-addr "${KONG_ADMIN_URL}" \
     --headers "Kong-Admin-Token:${KONG_ADMIN_TOKEN}"
   ```

   This succeeds.

4. Attempt an export:

   ```bash
   deck gateway dump --analytics=false \
     --kong-addr "${KONG_ADMIN_URL}" \
     --headers "Kong-Admin-Token:${KONG_ADMIN_TOKEN}" \
     --all-workspaces \
[...]

PR #2205, opened 2026-08-26
fix: avoid erroring out when server header does not exist and fall back

harshadixit12:
## Problem
We were depending on `Server` [header to identify AI Gateway instance](https://kongstrong.slack.com/archives/C0BEBVDL5TL/p1783929124305359?thread_ts=1783687758.252459&cid=C0BEBVDL5TL) - which was then used to format the output of `gateway dump` and to send Analytics. However, it is possible to turn headers off in Kong Gateway, and this leads to `gateway dump` being blocked.

https://github.com/Kong/deck/issues/2203

## Fix
Two steps
1. Stop erroring out when `Server` header is not found / when checking if it is AI Gateway
2. Use the Kong Gateway root response to check if `ai-model-selector` plugin is available

## Testing
Manual:
- [x] Use with Kong 2.x API Gateway with header enabled
- [x] Use with Kong 2.x API Gateway without header enabled
- [x] Use with Kong 3.x API Gateway with header enabled
- [x] Use with Kong 3.x API Gateway without header enabled
- [x] Use with Kong AI Gateway with header enabled
- [x] Use with Kong AI Gateway without header enabled

CI:
- [ ] Include `KONG_HEADERS=off` in RBAC instance tests

This is a stop-gap - and the actual fix is to have a [value in Kong root](https://github.com/Kong/kong-ee/pull/21602) indicating the `Gateway Family`


cmd/common.go
```diff
@@ -781,14 +781,39 @@ func fetchCurrentState(ctx context.Context, client *kong.Client, dumpConfig dump
 
 // This function uses client.Server to check for the Server header in the response.
 // If the server header contains "ai-gateway", it returns true.
+// If client.Server fails, it falls back to checking if "ai-model-selector" plugin exists in kongRoot.
+// For it to be considered an AI Gateway, the major Kong version must be 2.
 // Note that this takes client as input, and does not use the default client.
 // Therefore, the client configuration including headers, non-default addresses are respected.
-func isAIGatewayInstance(ctx context.Context, client *kong.Client) (bool, error) {
+func isAIGatewayInstance(ctx context.Context, client *kong.Client, kongRoot map[string]interface{}) bool {
```
harshadixit12: Also, this function won't error out anymore - we will assume "not AI Gateway" if both the checks fail.

cmd/common.go
```diff
@@ -781,14 +781,39 @@ func fetchCurrentState(ctx context.Context, client *kong.Client, dumpConfig dump
...
+		return true
+	}
+
+	// Check Kong version - must be major version 2 for AI Gateway
+	version, ok := kongRoot["version"].(string)
+	if !ok {
+		return false
+	}
+	parsedVersion, _ := reconcilerUtils.ParseKongVersion(version)
+	if parsedVersion.Major != 2 {
```
harshadixit12, line 811: Since deck is not planned to support AI Gateway v3

cmd/common.go
```diff
@@ -781,14 +781,39 @@ func fetchCurrentState(ctx context.Context, client *kong.Client, dumpConfig dump
...
+	}
+
+	// Fallback: check if ai-model-selector plugin exists in kongRoot
+	plugins, ok := kongRoot["plugins"].(map[string]interface{})
+	if !ok {
+		return false
 	}
-	return strings.Contains(server, aiGatewayServer), nil
+
+	available, ok := plugins["available"].(map[string]interface{})
```
VarunAthreya: just clarifying: `available` is only present in aigw and not in gw as there we have `available_on_server`

.github/workflows/integration-enterprise.yaml
```diff
@@ -92,6 +92,7 @@ jobs:
       KONG_ENFORCE_RBAC: 'on'
       KONG_PASSWORD: 'kong_admin_token'
       KONG_ADMIN_TOKEN: 'kong_admin_token'
+      KONG_HEADERS: 'off'
```
VarunAthreya: i think we should also add this in `setup_kong_ee.sh`

cmd/common.go
harshadixit12: good observation - no, this is a miss. I'm fixing, and adding tests.

.github/workflows/integration-enterprise.yaml
harshadixit12: yes

cmd/common.go
```diff
@@ -779,16 +783,46 @@ func fetchCurrentState(ctx context.Context, client *kong.Client, dumpConfig dump
 	return currentState, nil
 }
 
+// ServerGetter is an interface for getting the server information from Kong.
+type ServerGetter interface {
```
harshadixit12, line 789: Added to make `isAIGatewayInstance()` testable

.github/workflows/integration-enterprise.yaml
harshadixit12: Adding this in integration tests is a little more complex than I estimated - we are depending on the server header to run / skip tests based on version (needs updates to dependencies). Let me take it up as an improvement - here's a branch with passing integration tests where headers have been turned off - https://github.com/Kong/deck/pull/2211

VarunAthreya (approved):

PR #2214, opened 2026-08-31
feat: support new AI Gateway 2.0 features

harshadixit12:
## Summary
In this PR, we are - 
1. Bumping ai-deck-converter to latest version
2. Adding tests for `sni`, `certificate` and `ca_certificate` entities
3. Updating tests to account for new changes in interface
4. Use 2.0.3 runtime in tests

Note: for certificate entity, there was a known bug during round trip - https://github.com/Kong/ai-deck-converter/issues/159 which was fixed too.


tests/integration/ai_dump_test.go
```diff
@@ -106,7 +107,17 @@ func Test_AIDump(t *testing.T) {
 		},
 		{
 			name:      "identity providers",
```
VarunAthreya: shouldn't we also update the name here?

VarunAthreya (dismissed):

tests/integration/ai_dump_test.go
harshadixit12: yes - will do.
I need to get Shivay's PR merged to fix konnect test, then rebase - will push this change after.

VarunAthreya (approved):

PR #2227, opened 2026-09-08
Chore: sign and notarize binaries

harshadixit12:
## Summary
In this PR, we are updating goreleaser config to sign and notarize the binary for macOs
We are also bumping the goreleaser version to v1.26.2 to make use of native macOs signing feature.

A successful dry run (and binary) is available here: https://github.com/Kong/deck/actions/runs/34182608893/job/101924553295

The necessary secrets have been added.

## How to test
1. Download the binaries
2. `codesign --verify --deep --strict --verbose=2 /path/to/binary` - should pass with "valid on disk" and "satisfies its Designated Requirement"
3. Repeat step 2 with installed decK - `/opt/homebrew/bin/deck` - this will not be signed
4. Repeat step 2 with `codesign -dv --verbose=4 /path/to/binary` - you should see Kong Inc as Authority

## How I have tested this
- [x] Use macos binary from dry run to sync / dump
- [x] Use linux binary from dry run with docker


.goreleaser.yml
```diff
@@ -20,6 +21,20 @@ builds:
   - -trimpath
   ldflags:
   - -s -w -X github.com/kong/deck/cmd.VERSION={{ .Tag }} -X github.com/kong/deck/cmd.COMMIT={{ .ShortCommit }}
+notarize:
```
harshadixit12, line 24: No separate verification step added - because both signing and notarization will fail the build if unsuccessful.

VarunAthreya (approved):

Issue #2199, opened 2026-08-21
Partials have drift when configured on Konnect


harshadixit12:
When a partial like `redis-ee` is configured on Konnect usinf DecK, the subsequent `diff/sync` commands show drift.

More context: https://kongstrong.slack.com/archives/C04349E4KRC/p1787132806996019

VarunAthreya:
decK fills schema defaults into the partial's config before diffing. Konnect doesn't store or return those fields the same way, so every `diff`/`sync` sees a difference and reports an update even though nothing actually changed. The config on Konnect is correct  the drift is cosmetic, but it does mean each sync fires an unnecessary update call.

Workaround for now: skip the defaults filling.

```sh
DECK_SKIP_DEFAULTS_FILL=true deck gateway diff -s kong.yaml
```

Export it once in your shell or CI job and it applies to `deck gateway sync` too. With that set, the `redis-ee` partial drops out of the diff.

PR #2234, opened 2026-09-11
fix: fill schema defaults on current partial before diffing

VarunAthreya:
Fill schema defaults on partials before diffing, preventing spurious update diffs.

Fixes Issue: #2199

Corresponding GDR PR: https://github.com/Kong/go-database-reconciler/pull/527


harshadixit12:
@VarunAthreya as a rule of thumb - let's add integration tests in decK for any bugs.
GDR does not test with konnect.

